PR #536, opened 2018-02-07
support board with 2 relays and stm8s103

maciekczwa:
Hi
This pull request adds support for generic board which is based on esp-01 module and stm8s103 microcontroller.

Microcontroller is connected to esp8266 by serial port and is listening for commands to turn on and off relays.

Here is the link to the module:
https://www.aliexpress.com/item/5V-ESP8266-ESP-01-2-Channel-WiFi-Relay-Module-2-Channel-Relay-Module-For-IOT-Smart/32843970608.html

The buttons on board are inaccessible - they are connected to the stm8s103, and are used to restore/reset settings when using ESP-01 with Ai-Thinker firmware, they just send some AT+RST, AT+RESTORE commands when being hold.

The firmware in stm8s103 takes few seconds to start, and doesn't allow sending commands one after another - 8c629ba7ab1e0a915f230a40d730769270379b7b - this fixes the state of relays after boot up

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -275,6 +275,7 @@ PROGMEM const char* const custom_reset_string[] = {
 #define RELAY_PROVIDER_DUAL     1
 #define RELAY_PROVIDER_LIGHT    2
 #define RELAY_PROVIDER_RFBRIDGE 3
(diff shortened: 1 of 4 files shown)


xoseperez:
Thank you!

PR #558, opened 2018-02-13
 Add Reset to factory settings in admin page of web UI

hyteoo:
Good useful option for the web UI


code/html/custom.js
```diff
@@ -459,6 +459,30 @@ function doRestore() {
     return false;
 }
 
+function doFactoryReset() {
+    var response;
+
+    ask = (typeof ask == "undefined") ? true : ask;
+
+    if (numChanged > 0) {
```
xoseperez: Unnecessary, you are about to delete all settings with the factory reset, there is no need to save changes before.

code/html/custom.js
```diff
@@ -459,6 +459,30 @@ function doRestore() {
...
+        response = window.confirm("Some changes have not been saved yet, do you want to save them first?");
+        if (response === true) {
+          return doUpdate();
+        }
+    }
+
+    if (ask) {
```
xoseperez: Always ask. Remove the ask check (and references to the variable).

hyteoo:
True, that saved settings stuff was indeed useless.

PieBru:
BTW Xose changes above are a good quick tutorial for me on how to add a button to the WEB UI, a nice and simple jump start to tinkering also on that side of the Espurna framework.
Thanks to all.
Piero

xoseperez:
I'm on holidays now. As soon as I'm back home I will test it and merge it. 

Thank you!

Issue #562, opened 2018-02-13
Dimmer Scheduler


wysiwyng:
Is support for scheduled dimming already planned? I want to use Espurna on a NodeMCU board with multiple PWM controlled LED strips, and be able to change their brightness via a schedule. This might also be doable with MQTT, but I prefer to use this otherwise great software in a standalone fashion.

I had a look at the code, and it seems that firmware-side implementation should be not too difficult. However I have only rudimentary knowledge of web UI programming and thus no clue how to start there.

Tardymo:
Do you have a special Rf receiver connected to your car's central locking? Stock remotes have rolling codes so adding fixed code could be a security concern as somebody could sniff that fixed code and unlock your car...

xoseperez:
@wysiwyng  I'm flagging this as an enhancement but I don't plan to add such functionality soon...

@Tardymo  I guess this comment goes to #561 .

Tardymo:
Yep, somehow misposted on my phone. Deleted the comment.

wysiwyng:
solved in #574.

PR #574, opened 2018-02-16
Pwm dimmer

wysiwyng:
This adds support for schedulable dimmers as proposed in #562.


xoseperez:
I'm on holidays now. As soon as I'm back home I will test it and merge it. 

Thank you!

PR #592, opened 2018-02-21
UART<->MQTT

AlbertWeterings:
Adds support to send data over MQTT to UART and from UART to MQTT

### DIFF
--- a/code/espurna/config/arduino.h
+++ b/code/espurna/config/arduino.h
@@ -90,7 +90,7 @@
 //#define TERMINAL_SUPPORT       0
 //#define THINGSPEAK_SUPPORT     0
 //#define WEB_SUPPORT            0
-#define UART_MQTT_SUPPORT      1
+//#define UART_MQTT_SUPPORT      1
 
 //--------------------------------------------------------------------------------
 // Sensors (values below are non-default values)
--- a/code/espurna/config/arduino.h
+++ b/code/espurna/config/arduino.h
@@ -60,6 +60,7 @@
 //#define ARILUX_AL_LC02
 //#define WEMOS_D1_TARPUNA_SHIELD
 //#define MAGICHOME_LED_CONTROLLER_23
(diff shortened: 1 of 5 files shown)


hyteoo:
+1

A good addition to this would be to enable/disable this from the MQTT page in the web UI

AlbertWeterings:
@hyteoo Can you explaine the purpose of being able to switch the feature on and off in the webinterface?  The only two things I can imagine is to be able to enable debugging and have the terminal commands enabled, maybe you have other thoughts?

hyteoo:
Well, the purpose of the web UI is, apart from visualizing data and turning on/off outputs, enabling and disabling functionalities without recompiling+building+reflashing every single time.
From an UX standpoint, adding this functionality also to the web UI would make sense from my point of view.

AlbertWeterings:
Visualizing the data is I think the easiest to build in, it can be two fields like sensor value that are being updated with passing by data 1 for each direction. Personally I would only use it for data received over MQTT as in the opposite way I visualize it when it is send over MQTT. Both directions are also visualized in the telnet debug screen by the way.

Having the feature switchable sounds also great to me I would also add setting the serial baud rate from the UI in that case. I will have a look at how features like MQTT are being switched on/off and if a similar way fits this feature. 

I'm not familiar with the Webserver GUI so if someone wants to help with that part would be great.

So make it more concrete:
1. On/off switch for the feature in the Admin menu.
2. A new menu option only visual if feature is on.
3. In the new menu visualization of data.
4. Ability to set baud rate.

hyteoo:
You can take a look over a pull I've made, where I've added a button in the web ui [here](https://github.com/xoseperez/espurna/pull/558)
Check the flow from html-js-c in the pull as an inspiration.
After that, read the docs on how to rebuild the html into c-code [here](https://github.com/xoseperez/espurna/wiki/WebInterface)

PR #597, opened 2018-02-22
repair RFBRIDGE raw send issue https://github.com/xoseperez/espurna/i…

AlbertWeterings:
…ssues/580


AlbertWeterings:
This solution did not solve the problem correctly.

PR #605, opened 2018-02-25
feat: add support GizWits Witty Cloud

theonedemon:
add new device.
Info:
1) https://vimalb.github.io/IoT-ESP8266-Starter/Witty/info.html
2) http://www.my-iota.com/Development%20boards/Witty%20Cloud%20(GizWits)%20-%20ESP8266%20Development%20Board/Witty%20Cloud%20(GizWits)%20-%20ESP8266%20Development%20Board.pdf
Schema:
3) http://www.instructables.com/id/Witty-Cloud-Module-Adapter-Board/


theonedemon:
AnalogSensor not use pinMode(A0) and no possibility change pin.
For GL55XXSensor I planned using library https://github.com/QuentinCG/Arduino-Light-Dependent-Resistor-Library but it did not work to connect it as a dependency.

At the moment I can add in AnalogSensor new constant linking to a pin A0 ( default GPIO0 )

xoseperez:
AnalogSensor does use A0, it's the only ADC in the ESP8266. As for the library, it just adds a rule of three conversion from counts to luxes, no need to add the whole library for that.

theonedemon:
remove GL55XXSensor

PR #611, opened 2018-02-27
Espurna with push button (PUBLISHED FOR ERROR, MOVED)

Mattleyu30:
hi xoseperez
first congratulations to Espurna, an excellent firmware for sonoff, for the moment I have not pututo test alri products but I would like to start with the project that I would undergo.
I state that as a management software use Domoticz
to a Wemos D1 pro I would like to connect 3 push buttons on pins D5-GPIO14 / D6-GPIO12 / D7GPIO13, with these buttons I would like to activate relays' of a sonoff, but I would not switch from domoticz because the response times from the moment of pressure de button at the moment of closing the relay we vary from 1 to 3 seconds.
Is there the possibility of communicating directly with another Espurna without going from domoticz?
or is there a way to speed up communication?


xoseperez:
I guess this was wrong :smile:

PR #612, opened 2018-02-27
Reset relays on MQTT disconnection

a-tom-s:
With option #define MQTT_DISCONNECT_RELAY_DEFAULT set to 1, all relays
will be reset in case of MQTT disconnection to avoid a relay connected device (f.e. any water valves)
kept switched on while MQTT client is not under broker (and home automation) control any more.


theonedemon:
i think is good idea but current realization bad.
Set default option to 0, and add this setting on web interface and change his via UI

xoseperez:
I agree with @theonedemon comments. The default value should be the current behavior. It would be good to be able to set this option via the web UI also. 

But my main concern is that there should be some tiem buffer to prevent the relays to flip off and on when a transient disconnection occurs.

xoseperez:
Merged #653

Issue #623, opened 2018-03-05
Travis reporting 'build passing" while build is failing


lobradov:
If you look at the latest build: 
https://travis-ci.org/xoseperez/espurna/builds/349209615

it actually failed, while travis reported it's ok.

PR #624, opened 2018-03-05
Build process improvements (supposed to fail)

lobradov:
* #623 - Make builds fail on error (should help Travis show actual results, rather than always saying 'passing' even if there's a problem.) CI for this PR is supposed to fail, as normal build is actually failing (see travis logs) 
* Reference platform/framework by name/version, rather than from git.


lobradov:
hah, didn't see that you have `command || exit`, preventing `bash -e` to exit on failure.

PR #639, opened 2018-03-07
Dynamic serial settings for UART<->MQTT

mcspr:
There's different rate for RF Bridge (19200) and Dual (19230)
And uart-mqtt thingy could actually be used to send raw commands to the bridge.

To-do:
- [x] same buffer size as terminal/telnet
- [x] do not disable debugging and terminal with `SoftwareSerial`
- [x] settings for `SoftwareSerial` rx and tx pins
- [x] setting for both `Serial` and `SoftwareSerial` baudrate
- [x] opposite of `_uartmqttSetup()` - flagging loop as disabled, stopping mqtt and main loop callbacks, `end()` for serial.
- [x] ws and terminal methods to re-run `_uartmqttSetup()`, which now either uses `Serial` or recreates `SoftwareSerial` object with new pins from settings.
- [x] decide which default baudrate to use - either global or module one
- [x] actual web interface portion, apply settings on websocket
- [x] ifndefs for settings
- [x] hexlify mode a-la rfbridge rfraw, when data is displayed as hex string and sent as bytes


AlbertWeterings:
I would say completely right. But... Every device you can connect can have its own baud rate therefore the less confusion will be if you can set the baud rate for UART<->MQTT on one place in the configuration. So maybe it is better to exclude HW that is already using HW serial from the UART<->MQTT feature. This HW can then use software serial.

Any more idea?

mcspr:
I was going with less confusion too. Defining a custom board you can just define serial speed once. And maybe sometimes enable or disable uart-mqtt support.

About software serial - should it still disable terminal support plus debugging then?

xoseperez:
I added some modifications to the UART-MQTT module to support software serial. I think there should  be different serial speed configuration for this feature and for the debug serial output or the RFBridge or...

In the future, it will make sense to add module configuration to the web UI, so you can choose the speed and port for HW or GPIOs for software serial.

With software serial it's not necessary to disable debugging, you are right.

mcspr:
Considering this, maybe this should be reworked to be as such?
I already see semi dynamic sw serial in some sensors:
[espurna/sensors/MHZ19Sensor.h](https://github.com/xoseperez/espurna/blob/dev/code/espurna/sensors/MHZ19Sensor.h)
[espurna/sensors/PMSX003Sensor.h](https://github.com/xoseperez/espurna/blob/dev/code/espurna/sensors/PMSX003Sensor.h)
[espurna/sensors/V9261FSensor.h](https://github.com/xoseperez/espurna/blob/dev/code/espurna/sensors/V9261FSensor.h)

So this could be adjusted to have:
- [ ] setting for both `Serial` and `SoftwareSerial` baudrate
- [ ] settings for `SoftwareSerial` rx and tx pins
- [ ] opposite of `_uartmqttSetup()` - flagging loop as disabled, stopping mqtt and main loop callbacks, `end()` for serial.
- [ ] ws and terminal methods to re-run `_uartmqttSetup()`, which now either uses `Serial` or recreates `SoftwareSerial` object with new pins from settings.

mcspr:
@xoseperez I think this is done. Although, couple of conserns.
- This cannot work with rfbridge enabled, as `Serial.read()` is done by the bridge code first. Sending works ofc. And I do not know if it is warranted to make shared serial buffer for multiple modules to use or just mask them via defines.
- Another one is also shared with rfbridge - just as rfraw, this uses `strtol` without error checking.
- And maybe overlapping code converting hex back and forth can be shared (`rfbToChar` and `rfbToArray`)

PR #642, opened 2018-03-08
Load average

lobradov:
Calculate poor-mans load average of CPU utilization and display it on status page and MQTT telemetrics. 

Credits for idea go to ESPEasy team.

Useful for understanding if there's some polling that's holding up the CPU (like sync sensors reading or similar).


xoseperez:
Nice! Thank you

PR #646, opened 2018-03-09
logic for determining travis builds is already in build.sh

lobradov:
Since there's already logic to determine travis builds in build.sh, no need to explicitly pass the list of builds in .travis.yaml


xoseperez:
Damn, I forgot :)

PR #647, opened 2018-03-09
Additional Travis build coverage 

lobradov:
Added 2 more travis builds (though, almost empty) to test for: 

- Additional relay types (Light, dual and My92xx) 
- DHT
- My9231 light type
- SSL 

`travis02` and `travis03` image leave a lot room (pins) for additional sensor options and 3 images should still be faster than 50 :)


mcspr:
Try this with platform 1.6.0 that pulls SDK 2.4.1
https://github.com/esp8266/Arduino/commit/45f177f9855bab3df2c17243ff3533aa354e6908

Issue #650, opened 2018-03-09
Log console on "Status" page


lobradov:
Just heads up on idea I'm working: 

I'm thinking about log console on "status" page to show messages that would normally be displayed on Telnet/UDP/Serial port.

Motivation is, I guess obvious: telnet is no-no, UDP a bit awkward to use, and not all devices are reachable to be debugged via serial console.

As I'm not into WS/HTML/JavaScript programming, I'm trying to find some examples from around the net and fit them into Espurna. Any help is appreciated.

What I think needs to be done: 
- [x] - Implement a proper console placeholder in status page - either `<textarea>` / `<pre>` block, or some fancy JS console widget
- [x] - Check how to tie frontend back to Espurna via WS - periodic flushing of a buffer from `_wsLoop()` or ad-hoc call to `wsSend()`. I'm still looking at how whole frontend/backend works in Espurna so if anyone has idea - please point me in the right direction.
- [x] - Implement a single wrapper function for sending same message to Serial/Telnet/WS, depending on what's configured/connected.
- [x] - Replacing all serial/telnet/debug message calls with above wrapper

I can share a branch I'm working on, if anyone is interested to take a look and help out, before I raise final PR towards `dev` branch.

Please comment if you have suggestions and or ideas.

lobradov:
I found out that `_debugSend()` already exists and is hidden behind `DEBUG_MSG_P` macro, so last item is taken care of ;)

lobradov:
To document what I did so others can manage with code: 

# Firmware

I found that `DEBUG_MSG_P` is used for a lot of debugging, which is really a macro to `debugSend()` or `debugSend_P()`, which in turn are wrappers around `_debugSend()` that looks at all debugging channels and sends the message to them. I added a call to modified `wsSend_P()` using a `weblog` JSON key.

I had to modify `_debugSend()` to accept parametrized string, rather than a fixed one. I left old function commented out - if PR gets accepted, feel free to remove it.

# Web frontend 

I used a `<textarea>` block with some minor styling around it. 

Log is updated using `weblog` key, and `custom.js :: processData()` is just adding whatever the value is to the above `<textarea>`. 

To properly account for `\n` and `\r`, I had to change `custom.js :: connect()`.

PR #658, opened 2018-03-10
Debug console on web

lobradov:
My stab at building a web console, in reference to #650 
I agree it's ugly - feel free to change the visuals, especially alignment. 

Looking at Status page, it would be good if it was 2 columned layout in Desktop, and single columned (like now) in mobile versions.

### DIFF
--- a/code/espurna/debug.ino
+++ b/code/espurna/debug.ino
@@ -60,6 +60,14 @@ void _debugSend(char * message) {
         _telnetWrite(message, strlen(message));
     #endif
 
+    #if DEBUG_WEB_SUPPORT
+        #if DEBUG_ADD_TIMESTAMP
+            wsSend_P(PSTR("{\"weblog\": \"%s %s\"}"), timestamp, message);
+        #else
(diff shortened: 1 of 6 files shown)


code/espurna/debug.ino
```diff
@@ -60,6 +60,14 @@ void _debugSend(char * message) {
         _telnetWrite(message, strlen(message));
     #endif
 
+    #if DEBUG_WEB_SUPPORT
+        #if DEBUG_ADD_TIMESTAMP
+            wsSend_P(PSTR("{\"weblog\": \"%s %s\"}"), timestamp, message);
```
mcspr: This can be just `wsSend(message)` and timestamp can always be in the client via `Date.toLocaleTimeString()`

lobradov: In theory, yes (and it would look better), but this is consistent with Telnet and Serial messages. 
I'm happy to add it as optional field before the actual text received from device - would this be a good idea?

Issue #104, opened 2017-03-05
general_custom.h


xoseperez:
Originally reported by: **P.B. (Bitbucket: [PieBru](https://bitbucket.org/PieBru), GitHub: [PieBru](https://github.com/PieBru))**

----------------------------------------

By adding a line at the very bottom of general.h one could set one or more customized builds without modifying the official repo file which contains the default values.

Here is my general_custom.h

```
#!arduino
#undef EMBEDDED_WEB
#define EMBEDDED_WEB            1

#undef NTP_SERVER
#define NTP_SERVER              "it.pool.ntp.org"

#undef MQTT_TOPIC
#define MQTT_TOPIC              "/{identifier}"

#undef NOFUSS_SERVER
#define NOFUSS_SERVER           "192.168.1.9"

#undef MQTT_REPORTS
#define MQTT_REPORTS            (MQTT_STATUS_REPORT | MQTT_UPTIME_REPORT | MQTT_FREEHEAP_REPORT | MQTT_RELAY_REPORT)

#undef ENABLE_DOMOTICZ
#define ENABLE_DOMOTICZ         0

#undef ENABLE_FAUXMO
#define ENABLE_FAUXMO           0
```

Thanks, Piero.

Issue #468, opened 2018-01-25
custom.h include order


mcspr:
Using dev branch, *custom.h* is now defined before the default headers.
Not all definitions are guarded by ifndefs, so when custom header got settings like `ADMIN_PASS`, `NTP_SERVER`, `MQTT_TOPIC` they get overridden back to default ones.

xoseperez:
There are two different options here: parse custom.h first and use #ifndef everywhere in general.h or parse custom.h last and used #undef before to avoid warnings by the compiler.

Problem with the second option is that some definitions depend on others, even inside the same general.h file. I don't know if this is actually a design error but first option above looks better. That, of course, means adding #ifndef guards everywhere. Some of the settings already have them, basically those I had to redefine at some point.

mcspr:
My current custom.h (without couple of _SUPPORT defines):
```
#define ADMIN_PASS=...
#define WEB_FORCE_PASS_CHANGE=0
#define WIFI_SCAN_NETWORKS=0
#define NTP_SERVER=...
#define NTP_TIME_OFFSET=0
#define MQTT_TOPIC="prefix/{identifier}"
```

I would agree that first option is better. Plus with ifndefs there is possibility to also use PLATFORMIO_BUILD_FLAGS to sometimes change settings per build (debug, pins etc.)

lobradov:
Trying to figure this out is like playing "Rock paper scissors lizard Spock" :) 

Rules we have: 
- command line preprocessor flags get processed first
- defaults should set the general scene before "customizations". 
- customization should override defaults, but not command flags 
- command line flags should have the highest preference, overriding anything else
- dependency resolution should come last to make config consistent

I can think of a very complicated set of intermediary flags to signal if parameter was defined in command line or from default, and how should later steps be taken, but that would make code unreadable beyond any recognition.

Other option would be to restrict the use of command line to only point to custom configuration (`-Iconfig/my_config.h`), that would then be processed after defaults, but before dependency resolving set of macros... 

In either case, it would make sense to split types from defaults and dependency resolving macros, and have them in separate files for easier manipulation. 

@xoseperez - I can give it a try if you agree with overall approach.

xoseperez:
Yes, it makes sense. The general.h file is a mixture of everything (types, defaults, dependencies) and it sure needs some cleaning up. I started some time ago splitting the hardware.h file into definitions (current hardware.h) and defaults (defaults.h) file.

@lobradov If you want to propose a new structure for the header files go ahead.

[...]

PR #660, opened 2018-03-11
Fixes USE_CUSTOM_H to allow to override (back) any default settings

soif:
This PR :

**(1)** allows to really override ANY settings , because this has been broken by 1087134 . Commit 1087134 only allowed to override settings that are previously enclosed by #ifndef macros. 

But as you can see, most of the VERY useful settings (in fact most if not all, listed as example by the user in #104 that leaded to get the custom.h file implemented) are just NOT currently enclosed by #ifndef, and thus can no longer be overridden. 

So either  ALL settings must be enclosed by #ifndef, which make the code unnecessarily long and uggly, either this PR revert to the situation before 1087134, where absolutely ALL macros can be overridden, easily.

**(2)** it also adds a detailed explanation on how to use custom.h as well as an example file.

HTH

### DIFF
--- a/code/espurna/config/all.h
+++ b/code/espurna/config/all.h
@@ -19,10 +19,6 @@
 
 */
 
-#ifdef USE_CUSTOM_H
-#include "custom.h"
-#endif
-
 #include "version.h"
 #include "arduino.h"
 #include "hardware.h"
@@ -36,3 +32,7 @@
 #endif
 
 #include "build.h"
+
+#ifdef USE_CUSTOM_H
+#include "custom.h"
+#endif
--- a/code/espurna/config/all.h
+++ b/code/espurna/config/all.h
@@ -3,19 +3,22 @@
     If you want to modify the stock configuration but you don't want to touch
     the repo files you can define USE_CUSTOM_H in your build settings.
 
-    Arduino IDE:
-    define it in your boards.txt for the board of your choice.
+    * Using Arduino IDE: **********************************************************
+    Define it in your boards.txt for the board of your choice.
     For instance, for the "Generic ESP8266 Module" with prefix "generic" just add:
 
         generic.build.extra_flags=-DESP8266 -DUSE_CUSTOM_H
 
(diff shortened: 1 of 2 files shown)


mcspr:
See https://github.com/xoseperez/espurna/issues/468#issuecomment-363774907

It makes sense to also move all of
```
#if <SOME_DEFINITION>
#define <SOME_OTHER_DEFINITION>
#endif
``` 
from *general.h* into separate header file, included after *custom.h*

soif:
I agree about the order as described by @lobradov in #468.

> move all of .... from general.h into separate header file, included **after** custom.h

And thus, the same problem would happen : custom.h could no longer be able to override ANY settings. 😶 . While if it is the last include, it's up to the user to decide if he want settings in custom.h, to be optional (ie mastered by the command line, or previous declarations) or not...

soif:
I guess I'd not understood exactly what you stated. You're right some (carefully choosen) macros could go after, the custom.h.
But this is rather a large re-think, re-order of the whole #define files, which is needed.

In the meantime, putting custom.h at the end does not solve all, but most user needs. And wherever some macros like the one you listed are happening in file included before, you can still undefine, redefine whatever needed, by just copy pasting code from previous includes.

This PR aims just to be a temporarily fix, waiting someone to better design the whole macros files....

mcspr:
Comes the question - why even have custom.h then. You can just copy whole *config* tree elsewhere and just edit that. Then, the either *espurna.ino* or *all.h* could reference that with a flag like USE_CUSTOM_CONFIG

mcspr:
#781 did add bunch of `#ifndef` for settings defaults, no need to reorder custom.h (config.h?)
Sample file could be helpful though.

xoseperez:
After @mcspr changes with #781 and @lobradov refactor of the whole .h files this should be rethinked.
Adding #ifdef guards in the main config files might be tedious but frees the user from doing #undef in her custom.h file.

PR #665, opened 2018-03-12
Buttons should not retain in MQTT

lobradov:
If you set "MQTT Retain" to on, than all values are retained and kept for later.
While this is good for things that have state (relays, switches), it's not necessary good for pushbuttons which are fire-and-forget.

To make sure buttons are sent wihtout retain flag, regardless of general settings, I had to monkey around with few overloads of `mqttSend()` and `mqttSendRaw()`, so hopefully I didn't break anything.

@xoseperez  - counting you will double check before merging.


xoseperez:
Looks good to me. Will merge it tonight and test it.

PR #677, opened 2018-03-14
WebUI checkboxes improvements

mcspr:
* replaced `is(":checked")` with `prop("checked")`
* replaced `$(":checkbox")` with native selector
* change `showPanel` to resize checkboxes only in active view.
* same for scheduler **Enabled** checkbox


xoseperez:
@mcspr Thank you, Max!

PR #679, opened 2018-03-15
Export max system check for build configuration

phuonglm:
My kid loves to play with this switch ( turn on and off a lot) and sometimes it reaches the max unstable count so I need increase that number at build time.


xoseperez:
@phuonglm  Thank you!

PR #701, opened 2018-03-20
Make switch schedules work again

wysiwyng:
Apparently JS makes a difference between 1 and "1"

### DIFF
--- a/code/espurna/scheduler.ino
+++ b/code/espurna/scheduler.ino
@@ -26,7 +26,7 @@ void _schWebSocketOnSend(JsonObject &root){
         scheduler["schEnabled"] = getSetting("schEnabled", i, 1).toInt() == 1;
         scheduler["schSwitch"] = getSetting("schSwitch", i, 0).toInt();
         scheduler["schAction"] = getSetting("schAction", i, 0).toInt();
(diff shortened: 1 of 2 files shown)


xoseperez:
Mmmm... didn't notice that problem when testing. What about doing a parseInt to be sure about the output type?

wysiwyng:
That could work too I think. When I was testing, regardless of the button I clicked in the schedule tab, a light schedule was always added. This change fixed it for me, though I must say I am not the greatest JS artificionado...

mcspr:
Maybe this is happening because of this? It is actually a string when sent to the client.
https://github.com/xoseperez/espurna/blob/8777d29f7c780bc14396a795c947d777b6d97262/code/espurna/scheduler.ino#L29
Like here, that probably missing `toInt`
https://github.com/xoseperez/espurna/blob/8777d29f7c780bc14396a795c947d777b6d97262/code/espurna/scheduler.ino#L128

wysiwyng:
@mcspr, yes that is also part of the problem, it is however nested in somewhere else also. 1bab8e0 should fix this properly, I changed all occurrences of the variable schType to an integer.

zafrirron:
Thanks @wysiwyng & @xoseperez 

this fix Works like a charm...

PR #711, opened 2018-03-22
Update utils.ino

gn0st1c:
SPIFFS_SUPPORT info.
why don't we free some bytes here


xoseperez:
OK, I guess this is kind of a YANGNI (you are not gonna need it). I had this code around because I think SPIFFS might be interesting in the future when implementing some kind of local log. But it could be deleted now and recovered in the future...

PR #717, opened 2018-03-23
new devices & am2320 sensor

gn0st1c:
new devices & am2320 sensor


gn0st1c:
wil resolve conflicts and add new sensor..

PR #728, opened 2018-03-25
Resize WebUI elements for narrow screen

mcspr:
It says that `div .header > h1` inherits line-height value of 1.6em from .content class. And it looks like this in portrait mode:

### DIFF
--- a/code/html/custom.css
+++ b/code/html/custom.css
@@ -30,8 +30,11 @@
     color: #777777;
 }
 
-.header > h1 {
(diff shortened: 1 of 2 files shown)


mcspr:
While at it - upgrade hint unsquashed and headers can globally be slightly smaller to fit the screen.
Also, settings backup/restore/reset buttons can be one row at the top, nowhere near other clickable things.
beforebeforeafter

PR #736, opened 2018-03-28
External js css

lobradov:
I gave it a try to externalize some of the .js and .css and managed to save 43344 bytes on the final image. 

(I also bumped versions of jquery wheelcolorpicker and jquery but that's less important).

The only side effect would be accessing virgin Espurna on separate AP without internet access, for the first time on a browser that doesn't have necessary .js and .css already cached. Theoretically possible but very unlikely.  I saw a trick to access a page that just loads necessary assets so that they are cached. 

@xoseperez - don't merge just yet, trying to see if we can have conditional build, so that external .js/.css would be gulp-inlined on demand. Still, here's the PR to play with and see if it conceptually works.


gn0st1c:
also building our own jQuery removing effects, deprecated etc should save space.

lobradov:
yeah, but trouble with "out own" jQuery is that someone needs to maintain it and improve over time :)

gn0st1c:
[it can be automated with grunt](https://github.com/jquery/jquery)

lobradov:
give it a go, it's beyond my skillset anyway :) I just discovered gulp :)
I'll follow / help as much as time/skill allow me :)

gn0st1c:
@lobradov : added to my to do queue :)

btw, thank you.
you're the reason i'm using espurna. i decided to use espurna after i read [FOSS-Firmware-comparison-overview](https://lobradov.github.io/FOSS-Firmware-comparison-overview/)

xoseperez:
I played with the jQuery builder some time ago and I couldn't save much space. Current web UI uses too many features, mainly the color picker. So maybe a good option would be to find a cool but slimmer color picker and reduce size both in the color picker and maybe in the jQuery.

mcspr:
Also this https://github.com/fujaru/jquery-wheelcolorpicker/issues/27
I wish `<input type="color">` [[demo](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color)] - the ultimate space saver - was implemented by browser and not OS to at least be consistent between devices.

Issue #695, opened 2018-03-18
Add support of dual white/cct led strips (H801)


soprater:
Hi there,

thanks for your awesome work!
I wanted to use your firmware to control my cct led strips (24V+, CW, WW) with the H801.
When working with Home Assistant (MQTT) I can only work with the _color_temp_command_topic_ when under _Lights>Use color_ is enabled. Unfortunately the values are getting sent on Channel 0/1/2 (RGB) and it would be nice to have a feature to get the values for Cold White and Warm White (on 2 channels).

Example:
![image](https://user-images.githubusercontent.com/15094202/37569092-bddea1de-2add-11e8-97ee-7fabf16355c3.png)
Cold White/Channel 1: 255
Warm White/Channel 2: 0

Skaronator:
The current RGBW implementation isn't that good either. I've enabled this setting in the webinterface:
![image](https://user-images.githubusercontent.com/2886913/38176775-40a1679c-35f5-11e8-828a-ccba56ee1125.png)

And when I set it in the espurna webinterface to perfect white it'll disable the RGB strip but if it isn't perfectly white it will still show the RGB strip at 100% brightness.

@xoseperez You need to convert the RGB value in the webinterface to a RGBW value. Here is a example code in python:
https://github.com/home-assistant/home-assistant/blob/98e4d514a5130b747112cc0788fc2ef1d8e687c9/homeassistant/util/color.py#L372-L381

It is basically get the lowest number and subtract it from all numbers and use it as white value.

Skaronator:
Okay just testing with H801 via MQTT in Home Assistant and IMO the `white_value_state_topic` and `white_value_command_topic` command must be removed to get rid of this slider:

![image](https://user-images.githubusercontent.com/2886913/38176933-5c780248-35f8-11e8-9541-79c76d501a32.png)


And the white value must be extracted from the RGB value and when you have the white color value then you can easily calculate the color temperature to set W1 between 0-255 and W2 between 0-255.

I can help you with the maths if you want :P

soprater:
The icing on the cake would be to "abuse" controllers with more than 4 channels (e.g. H801) to use 2 cct strips at once (2x warm white, 2x cold white). I'd definitely help, but unfortunately I can't program c.

Skaronator:
@soprater 
You can do this already... just disable the RGB thing and control each channel on its own.
[...]

PR #749, opened 2018-04-03
 Implement RGBW correctly

Skaronator:


xoseperez:
Current approach was that (when _light_use_whit) white channel was used when RGB channels had the same value. Your proposal is to use white always for the "common" (minimum) value of all three channels, and use the color LEDs for the rest. 

Does it make the difference? What is your testing hardware?

Skaronator:
Yes it makes a difference. You can have the White LED turn on and add a little red for example. Also if you switch from white to a color it works much smoother since it doesn't instantly turn off the white strip.

My Hardware is a H801 with a RGB strip and a white strip. I'm ordering a CCT (Cold white + Warm white LED) Strip and a RGB strip once the color temperature is working properly.

I noticed a problem with the current approache: When you switch the color and go over the white spot it'll turn off the RGB strip and show white only. Technically this is correctly but doesn't look that good. Sadly, I don't have my Hue LED Strip anymore to compare it against Phillips implementation.

It might be better to calculate the RGBW values before the transition and not while the transition.

image

Skaronator:
I've updated the RGB transition code. It looks much better now. I still need to implement this in the HSV functions.

And #695 should be pretty easy to implement now.

xoseperez:
Fine

I'll try it before merging. I was concerned about breaking previous user experience with the light. Setting Use White to OFF will actually turn OFF all these changes, right?

Skaronator:
If you turn off the White channel setting it doesn't change anything.

This is probably a better explanation for old code vs this PR:

image
_________
image

Skaronator:
PR should be ready for testing. Works fine for me and I didn't spot any more bugs.

code/espurna/light.ino
```diff
@@ -474,8 +474,16 @@ void _lightProviderUpdate() {
 // -----------------------------------------------------------------------------
 
 void _lightColorSave() {
+
+    //TODO: Remove this once this code was in a stable release
+    // This force set ch3 to 0 which could be not zero when you update from a old version
+    if (_light_use_white) {
+      setSetting("ch", 3, 0);
```
Skaronator: @xoseperez I'm not sure about this. We need to reset this setting when a user upgrade from a old to a new version.

Should I leave it like this? Or do you have a script that run once a user flash/upgrade? Or does it wipe these settings anyway when you flash it?

Skaronator: I've rewritten most of the code so this is no longer needed ;)

PR #753, opened 2018-04-04
Handle switch toggle events;

zafrirron:
The main goal is to allow defining reset events also for toggle type buttons.

Allow to add events for Switch button, by defining number of switch toggles in time frame (based on the same doubleclick delay definition).
This change depends on "DebounceEvent" small code change (file added here)...
The result is getting same button events:
CLICK
DBLCLICK
LNGCLICK
LNGLNGCLICK

also for toggle type switches

two parameters added
BUTTON_LNGCLICK_CNT
BUTTON_LNGLNGCLICK_CNT

to allow defining how many toggles will trigger the events.

### DIFF
--- a/code/espurna/button.ino
+++ b/code/espurna/button.ino
@@ -86,7 +86,7 @@ uint8_t mapEvent(uint8_t event, uint8_t count, uint16_t length) {
         }
         if (count == 2 ) {
             DEBUG_MSG_P(PSTR("[BUTTON] : DBLCLICK"));
-            return BUTTON_MODE_NONE;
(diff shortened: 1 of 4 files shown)


mcspr:
*about bitbucket: forks are hidden behind plus button on the top left. changes to the library will be easier to see*

But doesn't just using pushbutton mode already allows this?
And debug code from mapEvent can be translating the numeric mode i.e. `Button #1 event 2` will be `Button #1 event CLICK`
https://github.com/xoseperez/espurna/blob/fa20c47f2905c46628cbdde475e899ecc30368a3/code/espurna/button.ino#L91

PR #754, opened 2018-04-04
Implementation of support for dual white/cct led strip

soprater:
My first pull request and start into c. I think it's a good start, but definitely improvable.

index.html: Added a button to activate support of cct led strips on channel 4 and 5
general.h: Added 3 global variables. Two of them define the range in kelvin which to expect.
homeassistant.ino: Added Brightness and Colortemp Topic to config if useCct is activated
light.ino: When RGB and useCct is enabled only the white values will be calculated for color_temp_topic.
-> Here the calculation of kelvin to channel values definitely needs to be checked.
	modified:   code/espurna/config/general.h
	modified:   code/espurna/data/index.html.gz
	modified:   code/espurna/homeassistant.ino
	modified:   code/espurna/light.ino
	modified:   code/espurna/static/index.html.gz.h
	modified:   code/html/index.html


Skaronator:
It might be better to define the Kelvin of the Coldwhite and the Warmwhite LED strip and then do a proper calculation which also include the RGB strip to "extend" the color-range.

PR #763, opened 2018-04-06
ThingSpeak Relay support

zafrirron:
For better visualization of relay status on ThingSpeak graphs (relay is a step function data type).
I added a current state data point immediately followed by the new state of relay (for ThingSpeak only).

This creates a nice step function graphs for relays (the very basic graph options of ThingSpeak allows limited data presentation options, so this allows selecting line presentation avoiding more complex data manipulations).


zafrirron:
I've got some issues with later tests on TS...

PR #771, opened 2018-04-09
Disable CSE7766 by default

krosk93:
CSE7766 was left probably by mistake enabled by default. This caused garbage in the serial while monitoring.


xoseperez:
You are right. My bad.

PR #772, opened 2018-04-09
private setting wifi mqtt

alexgavs:


xoseperez:
@alexgavs Not sure this is actually what you wanted to do...

PR #780, opened 2018-04-10
.h splitting and reorg

lobradov:
- Isolated dependencies
- Sensors dependencies in individual sensor files.
- enums, macros and prototypes.


code/espurna/config/dependencies.h
```diff
@@ -0,0 +1,76 @@
+#ifndef DEPENDENCIES_H
```
mcspr: Just as sensors this should be pragma once.

lobradov: good point, changed.

xoseperez:
It's a good approach, but I'm worried about the dependencies. For instance: modifying ADC_VCC_ENABLED is useless since this setting is used to set de ADC_MODE at the bottom of the sensors.h file... Maybe that call (ADC_MODE) should be in the analog sensor classes, same thing you did for I2C.

lobradov:
I see what you mean. I can do it - give me a minute.

xoseperez:
Let's go!

xoseperez:
I had to change some stuff (main problem was with DEBUG_MSG definition order but got it working. I think it's a good step. Things look more rational now.

PR #785, opened 2018-04-11
Implement CCT warm white and cold white LEDs for devices with 5 channels.

### LINKED ISSUE #695: Add support of dual white/cct led strips (H801)

Skaronator:
Tested and it works fine. **Only thing that is missing is a slider to set the Mired value in the Webinterface.**

The current approach is to use the white value of the RGB to RGBW calculation and split it in warmwhite and coldwhite based on the Mireds (153-500). It works fine but is not perfect. We just assume that you have a 2000 Kelvin and 6500 Kelvin LEDs. I might improve this in a future PR so we can include the RGB strip but we'll see.

Close #754
Fix #695


xoseperez:
On some 5 channels devices defined in ESPurna 4th channel is cold white and 5th is warm white. Does this PR make sense for those devices (SONOFF_B1, for instance)?

Skaronator:
Basically yes but it depends. If you use the lamp mostly as white light it'll be perfect otherwise not so. Currently it just set the Color Temp for the white values. So RGB values are preferable`[255,255,255]` and then you can set the MiReds.

(It would be better to set the MiReds and know the cold white Kelvin and warm white Kelvin and merge everything together including the RGB Strip but no clue how to calculate that.)

Should I switch the warm white and cold white channel?

Skaronator:
I've updated this PR, it has now a Slider in the Webinterface and setting the Kelvin/MiReds color will match more.

image

Channel 4 is coldwhite
Channel 5 is warmwhite

_Tested in Home Assistant and Espurna Webinterface_

PR #786, opened 2018-04-11
new board ESP_DIN_V1

pilotak:
I will update board info page soon, board is Open Source Hardware


xoseperez:
Great. I see there is nothing in the repo yet, please let me know when it's ready. I love Open Source :smile:

pilotak:
source files added, only readme to do

pilotak:
@xoseperez could this be merged?

pilotak:
just before it gets merged, do you think it woulv be better if i move LED to GPIO15 and opto-coupler to GPIO16 so you can have IN and RF at the same time?

xoseperez:
Notice that GPIO15 must be asserted low on boot or the board will not boot. But this would be a revision of the board, right? I might as well merge your config now.

pilotak:
I know IO0, IO2 and IO15 make it difficult. I have 4 pcb of version v1.0 done, assembled and ready to use. But I discovered that i needed to add RC filter for the relays in order to turn on/off induction loads, thats why i made v1.1 and published it online. I have not made PCBs yet so i can still add to it. The question is whether to merge this commit and keep it for v1.0 (but basically only for me) or whether to commit changes for v1.1 (with RF and LED pin change) and have only one version of FW for this board

xoseperez:
You are right. I would probably add support only for v1.1.

pilotak:
@xoseperez can it merged please?

PR #789, opened 2018-04-12
Private API

sametflo:
I think it will be great to be able to edit the thinkspeak's full url.

The main goal is to allow to set something like this:
api.mydomain.com/sonoff:8080

Modifying the minimum interval would be useful too.
This way we coul'd have our own interface with no foreign middleware

Sam


xoseperez:
Are you sure you wanted to do a PR?
You can change THINGSPEAK_HOST, THINGSPEAK_PORT and THINGSPEAK_URL in `general.h`.

sametflo:
I know that.
I can also redirect api.thingspeak.com in my dns.
But these are manipulations that are not accessible to everyone and I think it is a little change for a great feature.

I show espurna to my friends who have a sonoff, and we all agree that the missing feature in this firmware is a web notification.
I think even a simple webhook would be heavy, but add a simple input on thingspeak url could do the trick.
I tell myself that perhaps we are not the only ones to think so.

I apologize for my english.
Have a nice day.

Sam

12 avril 2018 13:51 "Xose Pérez" <notifications@github.com (mailto:%22Xose%20P%C3%A9rez%22%20<notifications@github.com>)> a écrit:
	Are you sure you wanted to do a PR?
You can change THINGSPEAK_HOST, THINGSPEAK_PORT and THINGSPEAK_URL in general.h.

xoseperez:
Fine. Anyway this is not a PR. I have created an enhancement issue. Move the conversation to #792. Closing here.

PR #790, opened 2018-04-12
espurna plugins hooks

zafrirron:
A very basic to support plugin 3rd party code integration (https://github.com/xoseperez/espurna/wiki/3rd-Party-Plugins), including hooks.

usage:
For every core hook (reasonable ask), just add 
```
#if USE_EXTRA
espurnaDoHook(hookName);
#endif
``` 
(with any hook name.)

3rd party plugin code only needs to register to any hook by 
```
espurnaRegisterHook(_pluginCallbackFunction, _hookName);
```

The plugin template and hook system will allow to divert some none core enhancement requests to plugins, and just give easy support by adding hook calls.
plugin developers can add hook call until core supports the hooks they need, (very small diversion from core code).

### DIFF
--- a/code/espurna/espurna.ino
+++ b/code/espurna/espurna.ino
@@ -24,6 +24,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 std::vector<void (*)()> _loop_callbacks;
 
+#if USE_EXTRA
+std::vector<void (*)()> _hooks_callbacks;
+std::vector<std::string> _hooks_names;
+#endif
 // -----------------------------------------------------------------------------
 // REGISTER
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


xoseperez:
Any reason why this should be limited to only when USE_EXTRA is 1?

xoseperez:
I think this is an interesting feature, it could probably replace existing hooks like wifiRegister or mqttRegister but not completely... There is also the broker module which can be used for the same purpose and adds a value field too...
So adding this one would add a third way to send messages internally. I think we should first clarify what model we want to follow and then choose.

mcspr:
@zafrirron Due to the age, suppose this is no longer viable... though, any comments on the above?

Broker is gone in the current build, everything exchanges data using module-specific callbacks / timers / events / etc. If there is a better alternative, some example would be appreciated. (or, maybe an existing app, which handles this plugin external code in some way)

PR #811, opened 2018-04-24
Minor: Change libs from URLs to names

lobradov:
PlatformIO tracks most of the libs by name now, so we can use it.
Kept strict versioning of the libs, to avoid issues.


xoseperez:
Still, I think it adds value to have the repo paths there...

PR #818, opened 2018-04-26
Refine PMSX003 to support PMS5003T/PMS5003ST.

Yonsm:
1. Support PMS5003T/PMS5003ST;
2. Rewrite data reading method;
3. Remove PMS library dependency (move control command into espurna);
4. Support smart sleeping mode to extend the sensor’s lifetime.

Tested successfully for PMS5003T and PMS5003ST sensor in NodeMCU. Imagine working on other PMSX003 devices as well.

Need help: I use MAGNITUDE_ANALOG for HCHO, since I don't known how to rebuild compressed web package. Could you please add HCHO MAGNITUDE support in custom.js, Thank you.

### DIFF
--- a/code/espurna/sensors/PMSX003Sensor.h
+++ b/code/espurna/sensors/PMSX003Sensor.h
@@ -218,7 +218,6 @@ class PMSX003Sensor : public BaseSensor, PMSX003 {
             _serial = new SoftwareSerial(_pin_rx, _pin_tx, false, 64);
             _serial->enableIntTx(false);
             _serial->begin(9600);
-            //_serial->enableRx(true);  // TODO: Test
             passiveMode();
 
             _startTime = millis();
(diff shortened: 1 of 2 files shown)


xoseperez:
I have added the HCHO magnitude (without units, is that right?).
One thing I think should be fixed is that the type of sensor (which I have moved to sensors.h) should be set dinamically, with a setType method. So all the defies at the top of the sensor file should be reworked too.
This will be necessary when the user could change the sensor settings on the fly from the web interface...

Yonsm:
Got it. Thank you.

The HCHO's unit is "mg/m³". And another suggestion for temperature unit is:

        PROGMEM const char magnitude_celsius[] =  "°C";   // Line 80
        PROGMEM const char magnitude_fahrenheit[] =  "°F";  // Line 80

 I'll follow your advice to support setType dynamically by program call, but I don't known how to support changing sensor type dynamically from Web Front End.

Yonsm:
I've submitted 'Support set PMS sensor type dynamically' on https://github.com/Yonsm/espurna/commit/f0ca139b52048d9d8b71c21f348e286f1dc53655. and will PR later once I tested on my true device.

PR #828, opened 2018-04-29
support for luani HVIO https://luani.de/projekte/esp8266-hvio

wildwiz:
This is a smart wall switch board, featuring 2x 5A mains SPST channels and 2x mains inputs (so it is suitable to replace two pulse relais)

more info on 
https://luani.de/projekte/esp8266-hvio/
https://luani.de/blog/esp8266-230v-io-modul/


xoseperez:
Thank you!

PR #870, opened 2018-05-18
removed unused classes. saves about 1kb

gn0st1c:
removed unused pure grid classes. saves about 1kb (index.html.gz)


gn0st1c:
coded a simple script to list all classes and check if they're used.
noticed sm, md and xl were not used.
just removing them saved ~1k

```
#!/bin/bash
all_class_file="css_class_list.txt"
unused_class_file="css_class_unused.txt"

cat $(find vendor/ -type f -name '*.css') | grep -Eo '\.[a-z]+[a-z0-9_-]*' | sort | uniq | sed s/.// > $all_class_file
echo "** unused classes **\r\r\r\r" > $unused_class_file

while IFS= read -r CSS_CLASS
do
  if ! grep -Erqi "([^(remove|has)]?class[(:|=|[:space:]*=>[:space:]*)]*[[:space:]\W]*[(\"|')]*[-a-z0-9[:space:]]*$CSS_CLASS|\\.$CSS_CLASS\b)" index.html custom.js; then
    echo $CSS_CLASS >> $unused_class_file
  fi
done < $all_class_file
```

xoseperez:
That's good. My only concern is that we are touching a 3rd party file, which means that this patch should be re-applied if we update the library in the future...

gn0st1c:
that's why i didn't touch the main pure.css file. in case the unused classes are needed in the future.

but in pure grid, unused classes are for different screen sizes and only lg (>= 1024px) is used.

mcspr:
May document this with https://github.com/pure-css/pure builder directly?

Is this the same thing?
```patch
index 7398e1f..a7f0b5b 100644
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -212,10 +212,7 @@ grunt.initConfig({

             options: {
                 mediaQueries: {
-                    sm: 'screen and (min-width: 35.5em)',   // 568px
-                    md: 'screen and (min-width: 48em)',     // 768px
                     lg: 'screen and (min-width: 64em)',     // 1024px
-                    xl: 'screen and (min-width: 80em)'      // 1280px
                 }
             }
         }
```

Different builds:

Size | Default
--- | ---
15K | build/grids-responsive.css
7.9K | build/grids-responsive-min.css

Size | Stripped
--- | ---
3.8K | build/grids-responsive.css
2.1K | build/grids-responsive-min.css

gn0st1c:
@mcspr : yes, it's the same thing.

xoseperez:
Yes, I'm with @mcspr, using the tools the developer provide seems a better idea.
I have updated the wiki page on how to build the web interface with this hack.

PR #881, opened 2018-05-24
fix #880 bad hostname regex

### LINKED ISSUE #880: can't accept any setting because of hostname checking
in any settings i always got this error message and can't accept any valid parameters:
```
Hostname cannot be empty and may only contain the ASCII letters ('A' through 'Z' and 'a' through 'z'), the digits '0' through '9', and the hyphen ('-')! They can neither start or end with an hyphen.
```
the reason for this that the default hostname is ESPURNA_XXXXXX, where XXXXXX is last 3 octets of chipID (actually 6).
but the hostname check in the code/html/custom.js validateForm do NOT allow underscore "_". So after that if i try to change anything than it's wall with tha above error because of the general/hostname by default will be ESPURNA_3727CF which is not valid according to RFC.
IMHO the best fix would be to change the default hostname in espurna/code/espurna/utils.ino:
```
void setDefaultHostname() {
    if (strlen(HOSTNAME) > 0) {
        setSetting("hostname", HOSTNAME);
    } else {
        setSetting("hostname", getIdentifier());
    }
}
```
to something better eg. replace _ to - of the getIdentifier() and set it as hostname.

lfarkas:
it seems you missed the dot both from hostname and ip address.

### DIFF
--- a/code/html/custom.js
+++ b/code/html/custom.js
@@ -164,9 +164,9 @@ function validateForm(form) {
     // No other symbols, punctuation characters, or blank spaces are permitted.
 
     // Negative lookbehind does not work in Javascript
-    // var re_hostname = new RegExp('^(?!-)[A-Za-z0-9-]{1,32}(?<!-)$');
+    // var re_hostname = new RegExp('^(?!-)[A-Za-z0-9-.]{1,32}(?<!-)$');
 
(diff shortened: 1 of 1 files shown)


lfarkas:
ok it turns out the hostname regex check only the short hostname and the bug is the default hostname

PR #890, opened 2018-05-26
Migrate from {identifier} to {hostname}

mcspr:
It is marked as deprecated since 1.12.2
Maybe do this instead?

### DIFF
--- a/code/espurna/migrate.ino
+++ b/code/espurna/migrate.ino
@@ -1001,6 +1001,15 @@ void migrate() {
 
     }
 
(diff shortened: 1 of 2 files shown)


xoseperez:
I think we should simply remove that line, that's what deprecated means. The key question here is "when"?. According to the semantic versioning approach this should be done with 2.0.

mcspr:
I did just notice that this will not run bc of cfgversion check at the top. Should cfg version be updated or just move it before the check?

Issue #831, opened 2018-04-30
Feature request: api for open relay for n miliseconds


dddsssasa:
I would like use sonoff basic for electric door lock. If I use standard api for relay {root topic}/relay/0/set and call it by complicated central unit (raspberry j2ee) which connected throw not 100% reliable wifi, the system could fall and the lock would remain open.

Best solution should be api(mqtt, web) such as 
{root topic}/relay/0/pulse  => { 5000}

wildwiz:
maybd you want to use the pulse feature ? on your espurna, go to switches, choose pulse mode normally off and give it a pulse time of, say, 0.25 s ?

dddsssasa:
eventuelly yes, but api would be more comfortable. Some clients also wont short and long pulse for door lock.

xoseperez:
Thanks to @mcspr  for #902. I've added the HTTP API entry point to his work. Now you can pulse any relay for an arbitrary time. To toggle the relay and toggle it back after 10 seconds just send this request (notice times are in milliseconds):

```
curl "http://192.168.1.123/api/pulse/0?apikey=AABBCCDDEEFF0011&value=10000"
```

or using MQTT:

```
mosquitto_pub -t ESPURNA_001122/pulse/0/set -m 10000
```

xoseperez:
Staging for release

PR #902, opened 2018-05-30
MQTT dynamic pulse timer for relays

### LINKED ISSUE #816: Change relay pulse time while active
Changing pulse time from Web does not reset currently running timer i.e. using *Normally off* / 60s:
*Turn on -> Change time to 10s -> ... wait 60 seconds... -> relay turns off*
It works afterwards ofc:
*Turn on -> ... wait 10 seconds ... -> relay turns off*

Also, because _relayConfigure is only called by websockets code, there is no way to set `relayTimeN` using terminal commands besides rebooting device.

mcspr:
Resolves #816 (ticker revert) and #831 (/pulse/set)
Add `{root}/relay/N/pulse/set` setter to toggle relay for specified time.
Display pulse time and mode in terminal.

Overrides relay pulse settings and starts ticker with them. Resets to values from settings after starting.


xoseperez:
Merged but I did some changes:
* Pulse payload in millis to allow under 1s pulses
* Added same functionality for HTTP API
* Topic set to `pulse/+` instead of `relay/+/pulse`

mcspr:
But it did use floats as an input - just like web does. Settings stored as floats too thus the conversion for getSetting. Subsecond values for web: 0.1s === 100ms

HTTP API is nice, yes 👍

xoseperez:
You are right, do you think we should use the same units in web interface and mqtt/api

mcspr:
Maybe just use ms everywhere? Gets rid of multiple divisions / multiplications and web would not require some clever hint to denote 0. value usage (it is obvious only if implementation is known beforehand)

Issue #896, opened 2018-05-27
Backup settings are truncated


meyergru:
Whenever I backup my settings, the resulting JSON file is truncated at some arbitrary point, mostly around 1500 bytes.

This results in a corrupted JSON that cannot be used to restore the settings afterwards. This happens on a Sonoff RF Bridge, BTW and I have used the 1.12.6 release, but 1.12.5 had the issue as well.

xoseperez:
The max size of the file might be limited by the available memory since right it created before being sent. Can you check how much free heap you have before trying to do the backu? You can use `heap` command from the debug window.

mcspr:
@xoseperez I did test this yesterday and noticed that content-length is wrong - it is for the not-yet-pretty json. `prettyPrintTo` returns real size that should be used.
Tested with fresh device, after saving once from web. Around 14k heap backup was successful - 2313 bytes, not cut randomly. With lower heap (12952) it did not even start file download.

xoseperez:
I guess the best option would be to "stream" the contents as they are generated...

meyergru:
I could not look at the heap size before now, since apparently, the web UI is not reachable via a proxied connection?

The free heap is 17904 bytes when the output gets truncated.

mcspr:
Amending my previous reply... I did some tests since, and have some things to note:
- AsyncResponseStream already does keep printed json string in `cbuf` object. Changing it to chunked response does not make much difference.
- Removing `webLog()` from the top of config method avoids truncation. Debug code sends websocket message and immediately begins sending large config, which somehow causes tcp stack to run out of memory (?)
This is the most direct workaround for this.
- There is similar truncation when running `keys` in telnet and web. Adding nop delay by calling `optimistic_yield(100)` at the end of `debugSend()` in case it has TELNET or WEB sessions active allows all data to pass.

And some calls to DEBUG_MSG can be optimized to happen less frequently *(plus, they have some interesting logic for timestamping...)*

xoseperez:
Interesting... @mcspr do you mind issuing a PR with your suggested changes?
What do you mean by "interesting logic"?

mcspr:
@xoseperez Sure. Just need to check if weblog can still work.

*// about DEBUG_MSG*
It allows line continuation and requires remembering how data is presented from the code before. Accidentally, while trying to reduce debugsend calls, I did replace much of these with string formatting.
[...]

PR #930, opened 2018-06-13
Free jsonBuffer after printing

mcspr:
Solving heap problems from #896. Limited testng prefilling eeprom with bunch of "a"*64 strings and accessing /config when websocket is active.
For example, _onGetConfig method when settings size is `3236/4096`

|state|getFreeHeap()|jsonBuffer.size()|
|---|---|---|
|initial|22512|40|
|after settingsGetJson()|14768|4370|
|after printTo()|12528|4370|
|after clear()|20544|0|

### DIFF
--- a/code/espurna/homeassistant.ino
+++ b/code/espurna/homeassistant.ino
@@ -44,6 +44,7 @@ void _haSendMagnitudes() {
             JsonObject& config = jsonBuffer.createObject();
             _haSendMagnitude(i, config);
             config.printTo(output);
+            jsonBuffer.clear();
         }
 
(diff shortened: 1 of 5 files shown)


code/espurna/homeassistant.ino
```diff
@@ -123,6 +124,7 @@ void _haSendSwitches() {
             JsonObject& config = jsonBuffer.createObject();
             _haSendSwitch(i, config);
             config.printTo(output);
+            jsonBuffer.clear();
```
mcspr: This is not technically required bc buffer will be deconstructed each iteration. Just a reminder that it can be removed after serialization in done.

PR #931, opened 2018-06-13
Use ArduinoJson to print to weblog

mcspr:
Use ArduinoJson to format weblog message. Avoid doing json->html escape.
Use `Text` in the client to avoid accidental element additions. (if for whatever reason `x => "<div>hello</div>` append() will treat it like jquery element creation)

### DIFF
--- a/code/espurna/debug.ino
+++ b/code/espurna/debug.ino
@@ -68,17 +68,20 @@ void _debugSend(char * message) {
 
     #if DEBUG_WEB_SUPPORT
         if (wsConnected() && (getFreeHeap() > 10000)) {
-            String m = String(message);
-            m.replace("\"", "&quot;");
-            m.replace("{", "&#123");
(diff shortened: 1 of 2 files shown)


code/espurna/debug.ino
```diff
@@ -68,17 +68,20 @@ void _debugSend(char * message) {
 
     #if DEBUG_WEB_SUPPORT
         if (wsConnected() && (getFreeHeap() > 10000)) {
-            String m = String(message);
-            m.replace("\"", "&quot;");
-            m.replace("{", "&#123");
-            m.replace("}", "&#125");
-            char buffer[m.length() + 24];
+            DynamicJsonBuffer jsonBuffer(JSON_OBJECT_SIZE(1) + strlen(message) + 17);
```
mcspr: 17 = `weblog` + `\0` + timestamp

PR #949, opened 2018-06-16
Added security headers for each HTTP response

ITNerdbox:
In web.ino three security headers were added.


xoseperez:
Merged, thank you very much!

PR #959, opened 2018-06-20
added index.htm.gz file size display

gn0st1c:
added index.htm.gz file size display.
this helps me when working on the optimizations.


xoseperez:
I'm changing the gulpfile.js to use a pipe to build the .gz.h files. I will be adding your code there, but using `log` module instead of `console.log`.

PR #968, opened 2018-06-22
WIP Web backup from form data

mcspr:
Instead of calling `/config`, reuse already received data from websocket to prepare backup.json

- Only keys that are received from the device will be used
- Ignore rare `readonly` input when generating data
- Show otherwise empty mqtt clientID

I did not touch the part that sends data back via websocket. That could be adapted too.
*(And I did not quite understand why scheduler needs resetting each time)*

### DIFF
--- a/code/espurna/mqtt.ino
+++ b/code/espurna/mqtt.ino
@@ -316,7 +316,7 @@ void _mqttWebSocketOnSend(JsonObject& root) {
     root["mqttServer"] = getSetting("mqttServer", MQTT_SERVER);
     root["mqttPort"] = getSetting("mqttPort", MQTT_PORT);
     root["mqttUser"] = getSetting("mqttUser", MQTT_USER);
-    root["mqttClientID"] = getSetting("mqttClientID");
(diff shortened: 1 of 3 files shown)


xoseperez:
Not all configuration data is sent (or must be sent) to the web UI so this method will not be able to recreate the full configuration...

mcspr:
Not only that, I hadn't noticed at first that appended elements are actually cloned and do not reference live onces. This still has some useful bits - `readonly` and empty checks for values, anchor downloader. I'll add that separately then.

mcspr:
Missing config keys are (that i noticed)
- `adminPass` - never gets sent to the web, may be different than built in (although, this can just be aske
- `cfg` - same, only gets set by the `migrate.ino` once
- list entities `keyN` are mapped to inputs by the specific key matching done earlier in the processing cycle

I still sort of think this may be viable in some other capacity, maybe sort of like "profile" system (not device specific). The only problem is modifying current system to store state outside the dom and make dom get data from there

PR #970, opened 2018-06-22
Use scons to track html changes

mcspr:
- Extract possible module candidate from headers (as mentioned in gitter)
- gulpfile modified to support building only one .gz.h file with command line switch
- Use scons to track changed files in html/ dir and call gulp when they are changed.
- Scons has couple of useful methods to check program existence, command execution etc. Using them instead of raw python methods.
- Checks for node and gulp.js, with warnings

codacy check does not like subprocess usage... but somehow does env.Execute in `build_webui`, which is `sh -c`

TODO: check textual parser from pio to avoid subprocess completely.
TODO2: chunked webui build. gziped files can be easily (hopefully) concat-ed together to have even more granular build even with arduino ide


mcspr:
I did notice the [softcfg](https://github.com/xoseperez/espurna/tree/softcfg) branch making stock config.json bundled within firmware instead of using sets of defines. Scrap headers parser then?

xoseperez:
Sorry, missed this last comment. I have not yet decided how to "ship" configuration with the binaries.

mcspr:
Oops. I did forget about this.

Considering this once more, simple check for 'html' changes each time does only work per-environment.
Maybe something like 'dot file_name dot hash_string/mtime_of_htmldir' nearby should be checked instead.

I did also try lib_ignore via same header values to shorten the build. Which did instantly break on sensors. Trying to reorder headers did not work and broke i2c definitions (and rebasing broke commit author, somehow) :/ Current layout has things pulling Arduino.h which does require prior framework location knowledge (core/esp8266 include dir etc.) that is not available in pre: extra script. SCons.cpp.PreProcessor does better at this, but needs changing every method to handle #defines instead of #includes (it does check but does not stop on #if/#else and ends up reading every #define statement in order)

*OT* And about softcfg... wouldn't same .bin files work? repo keeps jsons as both reference (which should work uploading to /config) and hardware profiles for compilation. Library headers are still bound to defines.

Issue #950, opened 2018-06-18
Pre-release binaries


mcspr:
As mentioned in #944 and #896 - there is a way to build `dev` branch firmware outside of repository itself and release it using CI services.
Currently testing via Travis and Gitlab (... still contemplating CircleCI)

https://github.com/mcspr/espurna-travis-test/releases
https://travis-ci.org/mcspr/espurna-travis-test/builds/402353371
Using latest HEAD to trigger the build. Because it starts on schedule - check for latest release commit (description of release) and stop if it is already built. Ignores espurna/code/build.sh and uses separate script.

https://gitlab.com/mcspr/espurna-travis-test
https://gitlab.com/mcspr/espurna-travis-test/pipelines/25418527
Same as travis. Releases go to release.tar.xz file attached as artifact and github.
Note: `package:release` / release.tar.xz is visible in the job log and can be downloaded from web or using API. It does expire after some time though.

Current implementation triggers build on schedule or api and tags release as current UTC date. Both gitlab and travis can release to any repo (that API token has access to). All CI services support triggering a build on demand by using API so it would be possible to pass some variables to build specific commit, change ci script logic or even change pio logic (for example, change platform 1.5.0/1.6.0/1.7.1 etc.)

mcspr:
Kudos to #986, there is saner way to build releases. So scratch OP part about totally parallel gitlab/e-t-t repos. 

$TRAVIS_REPO_SLUG is changeable - [releases deploy script](https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/releases.rb#L40) uses it and happily deploys wherever auth token allows. Triggering can be scheduled based on time (first build + every 24 hours, 7 days or 1 month) via travis cron or using it's API and doing manually.

Because `releases` provider is bound to git tags, it is either `$(date)-$(git rev-parse --short HEAD)` or single `latest` with some name mangling to overwrite old files.

Removing old releases either from web or using api (maybe through deploy key and additional script for after_deploy stage)

mcspr:
Builder repo now has 2 working implementations for gitlab and travis.

Scheduling builds for gitlab is more direct - syntax is simple cron (0 7 * * * is 7:00 every day) but builds on are much slower than travis - [23min](https://travis-ci.org/mcspr/espurna-travis-test/builds/402353371) vs [53min](https://gitlab.com/mcspr/espurna-travis-test/pipelines/25418527). Firmware build time is twice as long. Maybe need to try non-shared runners...
[...]

PR #986, opened 2018-06-25
Travis parallel builds

lobradov:
Make travis run in parallel. 
Minor improvement for dev branch, but major for full releases. 
Based on months-old work by @mcspr and myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,21 @@ sudo: false
 cache:
   directories:
   - "~/.platformio"
+  - "$TRAVIS_BUILD_DIR/code/.piolibdeps"
+  - "$TRAVIS_BUILD_DIR/code/espurna/node_modules"
 install:
 - pip install -U platformio
 - cd code ; npm install --only=dev ; cd ..
+env:
+    global:
+        - BUILDER_TOTAL_THREADS=4
+    matrix:
+        - BUILDER_THREAD=0
+        - BUILDER_THREAD=1
+        - BUILDER_THREAD=2
+        - BUILDER_THREAD=3
 script:
-- cd code && ./build.sh &&  cd ..
+- cd code && ./build.sh -p &&  cd ..
 before_deploy:
 - mv firmware/*/espurna-*.bin firmware/
 deploy:
--- a/code/build.sh
+++ b/code/build.sh
@@ -9,15 +9,36 @@ travis=$(grep env: platformio.ini | grep travis | sed 's/\[env://' | sed 's/\]/ 
 available=$(grep env: platformio.ini | grep -v ota  | grep -v ssl  | grep -v travis | sed 's/\[env://' | sed 's/\]/ /' | sort)
 
 # Parameters
+while getopts "lp" opt; do
+  case $opt in
+    l)
+        echo "--------------------------------------------------------------"
+        echo "Available environments:"
+        for environment in $available; do
+            echo "* $environment"
+        done
+        exit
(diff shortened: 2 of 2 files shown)


mcspr:
🎊 Hello.

What about explicit names for matrix?
When script is `./build.sh $BUILD_ME`, env BUILD_ME=travis01, BUILD_ME=travis02 ... will buil explicit test binaries by name. That way you can just use BUILD_ME=itead-sonoff-basic for release (plus branch condition ofc)

Also see #950 (where ./build.sh is wrapped, but doing that explicit env call) for working parallel release test. I wonder if there is a way bundle hardware config instead of building arduino core 90 times for each binary :/

lobradov:
You can still use name of the build as parameter, but in case you omit it AND you add `-p`, builds will go parallel. I personally am against manual parallelization (manually splitting target builds into groups), but than again, it's just me.

The only change compared to old build script is the `-l` instead of `list`, for the sake of consistency.

mcspr:
It does show OK but awk actually broke

Well, awk does splits jobs by number already. Consider if ini is parsed and pre push commit does awk job listing all the jobs into the travis yml. Why stop at 4 jobs :) *...limit is 200...*

lobradov:
> It does show OK but awk actually broke

I must have missed something... what do you mean?

lobradov:
pfft, have no clue where that backslash came from. anyway, it's fixed now, but I see Travis randomly failing, which I think was the original issue.

lobradov:
It would be super-good if we had a bucket with S3 storage somewhere, so that we can create build environment in one task, store pio + pio libs + npm + web UI, and then use that in parallel tasks that depend on the first one.
 
I guess, without it, this is the best/fastest we can get with this logic of building. We're still left with alternatives: 
- Your wrapped build.sh, potentially massively parallelized (which I'm sure Travis folks would love ;)
- Docker builder 
- Circle-ci or some other CI service.

mcspr:
https://github.com/esp8266/Arduino/commit/a1a9b99af664b0ab084c7a1ffddecd53888e2e41 arduino/esp8266 does it this way too, using bash math instead of awk.

Because there are 3 travis firmwares one job does nothing but build webui. job count can still be fixed for test, just use this split for releases until there are less of them to build. Other ci (i mean, all of them) do have saner options for doing this chain, but still require building N images.

mcspr:
@lobradov What I meant by splitting tests from release:
https://github.com/lobradov/espurna/compare/parallel-travis...mcspr:checkme

https://travis-ci.org/mcspr/espurna/builds/397608343 - untagged commit (push, pr)
https://travis-ci.org/mcspr/espurna/builds/397609136 - tagged commit adds 4 builder jobs, does the matrix thing

xoseperez:
This is a massive improvement when releasing! 
How would it work for pre-release versions or something like nightly versions?

lobradov:
Nightly builds are a matter of Travis settings (trigger nightlies) and publishing.

I'm out on a vacation with only but a phone access to emails, but I'm sure Max can help if you merge this PR and let him work on it.

//_ on phone

mcspr:
True. For nightlies deploy provider can put releases somewhere else, despite what docs / issues say (jeez, I hope that will not be changed) - $TRAVIS_REPO_SLUG controls where to put releases. `export ...` before_deploy can change that because environment is not reset.

So I'll put that branch as PR. *I did open pr against this pr, but I figure that is not going to be merged soon :)*

Issue #886, opened 2018-05-25
REST API: one common status page


lfarkas:
currently is very complicated to query all sensors data ie.

- first  call /apis
- then call all api endpoint and each call gives only one value.

eg it'd be very useful if i can call a /api/status which gives a "real" json result with all sensor data.

lfarkas:
of course it could be a graphql api which would be more user friendly but in this simple case may a rest api would be enough. for the only the query will be important (and wouldn't like to call 3-10 http request just to get 3-10 simple value. i review the current influxdb interface and decided that i'd rather write my own telegraf input plugin. but with this so many call it's rather annoying...
anyway setting all parameters which one can also modify on the web interface would be useful.

thebradleysanders:
Enhancement: Add RSSI, Uptime, avg load... To rest api similar to how data is transmitted to MQTT.
This would be a great addition as I want to build a status panel in PHP to see all my Sonoffs without the need of MQTT. Thank you.

Cabalist:
RSSI would be great!

mcspr:
ref. https://github.com/xoseperez/espurna/pull/2332

PR #1022, opened 2018-07-03
[wip] HTTP API device info and status (json only)

### LINKED ISSUE #446: REST API: status entry point
Originally reported by: **mtufan (Bitbucket: [gn0st1c](https://bitbucket.org/gn0st1c), GitHub: [gn0st1c](https://github.com/gn0st1c))**

----------------------------------------

**READ:**
api/status returning the values on status web page (_wsOnStart function)
*especially mac/rssi/firmware info/uptime are helpful*

and maybe?
**WRITE:**
if need be, setting a new ssid/password

mcspr:
Resolves #446, #886 

Add two api endpoints - one to get device info and another one to get it's status.

JSON only. Albeit having methods to register apis, they are limited to 15 bytes of data. Because ws methods do so nicely fit here, maybe http api can use them too (for example, different endpoints for sensors and one that merges them all via simple loop feeding json object)

### DIFF
--- a/code/espurna/api.ino
+++ b/code/espurna/api.ino
@@ -134,10 +134,32 @@ ArRequestHandlerFunction _bindJsonAPI(unsigned int apiID) {
 
         AsyncJsonResponse *response = new AsyncJsonResponse();
         JsonObject& root = response->getRoot();
-        (api.getFn)(root);
 
-        response->setLength();
-        request->send(response);
+        if (request->method() == HTTP_GET) {
+            (api.getFn)(root);
+
+            response->setLength();
+            request->send(response);
+
+            return;
+        }
+
+        if (request->method() == HTTP_PUT) {
+            if (!request->hasParam("value", true)) {
+                root.set("error", F("Missing \"value\" parameter"));
+                response->setCode(400); // Bad request
+            } else {
+                AsyncWebParameter* p = request->getParam("value", true);
+                (api.putFn)((p->value()).c_str(), root);
+            }
+
+            response->setLength();
+            request->send(response);
+
+            return;
+        }
+
+        request->send(405); // Method not allowed
     };
 }
 
--- a/code/espurna/config/prototypes.h
+++ b/code/espurna/config/prototypes.h
@@ -14,7 +14,7 @@ extern "C" {
     typedef std::function<void(char *, size_t)> api_get_callback_f;
     typedef std::function<void(const char *)> api_put_callback_f;
     typedef std::function<void(JsonObject&)> json_api_get_callback_f;
-    typedef std::function<void(JsonObject&)> json_api_put_callback_f;
+    typedef std::function<void(const char *, JsonObject&)> json_api_put_callback_f;
     void apiRegister(const char * key, api_get_callback_f getFn, api_put_callback_f putFn = NULL);
 #else
     #define api_get_callback_f void *


mcspr:
I need to finish this <_>. Either as strictly json api or current combined form-data / json

One thing i noticed about arduinojson - some class implementing Print can be used to output data. In this case it can be used to print form-data. *If* this is still a requirement.
https://github.com/bblanchon/ArduinoJson/blob/fa1a40ac6e63c18583540806d4fc63297ab5f99e/src/ArduinoJson/Serialization/JsonWriter.hpp#L16-L24

code/espurna/utils.ino
```diff
@@ -346,6 +346,50 @@ void info() {
...
+    root["app_revision"] = APP_REVISION;
+    root["manufacturer"] = MANUFACTURER;
+    root["chipid"] = String(chipid);
+    root["mac"] = WiFi.macAddress();
+    root["bssid"] = String(bssid_str);
+    root["channel"] = WiFi.channel();
+    root["device"] = DEVICE;
+    root["hostname"] = getSetting("hostname");
+    root["network"] = getNetwork();
+    root["deviceip"] = getIP();
```
soif: If you get  the API answer, you ALREADY know the IP address, don't you?

soif:
It would be great if you could merge this. 👍 💯 

BTW would you please add the Domoticz_idx: 
When owning a lot of Espurna boxes (like me), its really a pain to maintain a list of Domoticz Ids <-> Espurna IPs. Adding this would allow to easily build a script to build such a list.

I might even be able to develop a nice [icingaweb2](https://icinga.com/) module, to print these IDs in each host information, in Icinga web interface.

Keep Up the good work! Cheers!

code/espurna/utils.ino
mcspr: Yeah, but that would mean additional step to get (m)DNS response?

mcspr:
Hey, thanks, glad to hear it :)

This just needs a small rebase to be able to run with refactored API (373fc535452a5e91b8fb81954d58c471eede3724). I see that the PR adds some code into the lambda, which was considerably heavy on memory.

Will domoticz stuff reuse the same model and integrate `_domoticzWebSocketOnConnected()` as (for example) `/api/domoticz` endpoint?
https://github.com/xoseperez/espurna/blob/a2599de4d11481ba9f9712087472ee63bc7e9a7f/code/espurna/domoticz.ino#L203-L220

Also note of the [`/discover` endpoint](https://github.com/xoseperez/espurna/blob/a2599de4d11481ba9f9712087472ee63bc7e9a7f/code/espurna/web.ino#L75-L94) (without any api or http auth, ref https://github.com/xoseperez/espurna/pull/1933#issuecomment-562182418)

code/espurna/utils.ino
soif: I'm not really used with mDNS timing because I use a BIND dns server (with 2 slaves) at home ...yes, i'm a geek ;-)...and in this case query times are totally negligibles. Anyway mDNS queries should not be that long too... whenever someone really need to resolve the IP. 

I'm just trying to save on precious ESP memory.

soif:
You might also integrate this directly in the `/api/device` endpoint?
Maybe as a nested array, ie:
```
.....
device: "DEVICE"
.....
domoticz: {
     Enabled: true,
     TopicIn: "/topic/xxx",
     TopicOut: "/topic/yyy",
     RelayIds:[100,101]
}
....
```
Thus avoiding multiple queries to the API, to get all the  device Information... Those ESPs webservers are not that fast, and I guess that it stress them a litte bit to do all the http stuff, while still trying to get good timings for much important features, like being reactive on mqtt commands.

mcspr:
But does this mean to stuff all of the data from ws callbacks into a single endpoint? Any way to just use ws directly if that's the case? It already works and will stream all of the available modules data + status
Nested stuff also looks almost as in vue refactor, so i'd look at the callback structure there.

HTTP also allows some args passing, client could just request what it wants?
e.g. `/api/device?modules=dcz,relay,led,...`, device will just ignore any unknown ones

soif:
The idea is to get the maximum of information from a single endpoint (if it is possible, regarding memory issues on the esp), thus to get all the configuration (ie all settings, and maybe also runtime values like load, used memory... and maybe also sensor values or relay states) you would only need to send _one and only one_ http request, ie `/api/info?key=xxx`, instead of having to send multiple different requests (separate endpoint or separate GET queries), to not stress the esp too long.

From a client perspective to fetch API response,  it is as simple as (ie in php):
```
$array=json_decode(file_get_content("http://esp.local/api/info?key=xxx"));
```
that's it!

I guess (i'm not experienced with WS) that relying on websocket, from the client perspective, needs a **much** longer/complicated code ([example](https://stackoverflow.com/a/16608429) not tested) to perform the same thing, and you'd need to document the ws port, ws key, and ws messages used... not really easy.

IMHO the HTTP API  is the right way to go.

Splitting into different enpoints would be only required if it cause memory problem on the Esp to send all data from a single endpoint.

Then other endpoints would only be needed to POST actions to be performed by the API, ie "switch relay0 ON", "reboot esp", "Change BSSID" etc...

mcspr:
Resulting JSON is big though.

Async server is a bit weird here, because the general approach is to just dump serialized JSON string in a separate buffer (JsonBuffer keeps data -> resulting string, so we take twice the space. 5...10KB total depending on the device flavour). This was already done for /ws in 1.13.5 and the problem was that we can't do a second request, because we hit OOM most of the time. Thus now we only send small objects and one at a time.

I do see a solution though, but I am a bit worried that we stretch ESP8266 threading model again :>
AsyncWebServer stuff can periodically poll some buffers for data to send back, but the main issue is that ArduinoJson serializer expects us to finish all of it in the same serialize() call. Arduino Print abstractions can do some work here, but I am not yet sure how

mcspr:
Re: 
> I do see a solution though, but I am a bit worried that we stretch ESP8266 threading model again :>
AsyncWebServer stuff can periodically poll some buffers for data to send back, but the main issue is that ArduinoJson serializer expects us to finish all of it in the same serialize() call. Arduino Print abstractions can do some work here, but I am not yet sure how

#2247 API handler for command line does exactly this. We capture `request` object when something connects to us and create a 'task' that gives us a generic [Arduino's Print interface](https://github.com/arduino/ArduinoCore-API/blob/master/api/Print.h), which we can use to print / printf / write(byte) / etc. into. HTTP response is chunked, so we don't need to know the resulting size beforehand.

Following existing implementations with JSON in the name, https://github.com/me-no-dev/ESPAsyncWebServer/blob/master/src/AsyncJson.h does not implement chunked encoding and is limited by the 1KB (lwip v2 536 mss) / 3KB (lwip v1 & v2 high bandwidth) response, including headers.

I don't think I happened to find any JSON generator that can work in-place, most (if not all) implement a temporary buffer. Writing raw bytes will work (e.g. `printer.printf("{\"a\": %d, \"b\": %d}", getA(), getB());`), nesting might become tricky though.

PR #1032, opened 2018-07-06
Show APP_REVISION value in terminal and web

### LINKED ISSUE #1023: APP_REVISION value in UI
APP_REVISION is sent via websockets, but is not shown anywhere. Terminal does not display it too.
https://github.com/xoseperez/espurna/commit/3cd58e7f7c06652a4d92c9ca822f91ee003b62e6#diff-d55db8bfadc1744610632df7386409d3L223 even removed commented out section with it.

Instead of build.sh, extra_script could add this value to the build flags instead of modifying header.

mcspr:
Resolves #1023 
Show revision in terminal and web.

#1008 added APP_REVISION as build flag for build.sh builds

### DIFF
--- a/code/espurna/utils.ino
+++ b/code/espurna/utils.ino
@@ -260,7 +260,11 @@ void _info_print_memory_layout_line(const char * name, unsigned long bytes) {
 void info() {
 
     DEBUG_MSG_P(PSTR("\n\n"));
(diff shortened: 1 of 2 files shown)


mcspr:
Hm. Need to check `projenv` / `env` difference. 3.5.4 is current, tried with 3.6.0a...

PR #1034, opened 2018-07-06
Try to find wifi again

piotrpz:
In case lost of power if espurna cant find defined wifi network it is switching to own AP mode.
Id like to ask about new option - try to connect to defined network after defined time (or reboot after defined time).  In this case if booting time of AP is longer then espurnas it will connect in next try.


mcspr:
@xoseperez ping
@piotrpz please open issue instead if this is still a problem

PR #1048, opened 2018-07-12
add support for custom VCC monitoring

poulch74:
custom VCC monitoring with correction. enabled with defines:
   // for example
    #define ADC_MODE_VALUE      ADC_VCC_CUSTOM
    #define ADC_VCC_CUSTOM_MUL (15.63)
    #define ADC_VCC_CUSTOM_ADD (0)

in  hardware section description (hardware.h)

### DIFF
--- a/code/espurna/config/sensors.h
+++ b/code/espurna/config/sensors.h
@@ -646,6 +646,17 @@
 #define ADC_MODE_VALUE                  ADC_VCC
 #endif
 
+// custom Vcc mode (ADC_TOUT mode with correction)
+#define ADC_VCC_CUSTOM                  1
+
+#ifndef ADC_VCC_CUSTOM_MUL
+    #define ADC_VCC_CUSTOM_MUL          (1.0)
+#endif
+
+#ifndef ADC_VCC_CUSTOM_ADD
+    #define ADC_VCC_CUSTOM_ADD          (0.0)
+#endif
+
 // -----------------------------------------------------------------------------
 // I2C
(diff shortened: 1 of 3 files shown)


code/espurna/config/sensors.h
```diff
@@ -646,6 +646,17 @@
 #define ADC_MODE_VALUE                  ADC_VCC
 #endif
 
+// custom Vcc mode (ADC_TOUT mode with correction)
+#define ADC_VCC_CUSTOM                  1
```
lobradov: why not make this a configuration parameter, guarded by ifndefs?

code/espurna/utils.ino
```diff
@@ -191,9 +190,8 @@ void heartbeat() {
                 lightMQTT();
             #endif
             #if (HEARTBEAT_REPORT_VCC)
-            #if ADC_MODE_VALUE == ADC_VCC
-                mqttSend(MQTT_TOPIC_VCC, String(ESP.getVcc()).c_str());
-            #endif
+                int Vcc = custom_getVcc(ADC_MODE_VALUE);
```
lobradov: if would be great to introduce this feature with conditions, like: 
```
#if ADC_VCC_CUSTOM
int Vcc = custom_getVcc()
#else
int Vcc = ESP.getVcc()... 
#endif
```

lobradov: ah, I see that it's actually done in `custom_getVcc()`

PR #1055, opened 2018-07-14
Button

InduPrakash:
Added support for using button as door sensor. One can define the topic to where open/closed messages are published. Included a test setup in travis03.


lbdroid:
This looks like it has 2 purposes;
1) To add the GPIO state to the web UI,
2) To publish the open/closed state to MQTT.

Seems like its a lot more complicated than it needs to be, for what it does, and in other ways, inadequate.

First, for the excessive complication;
- Add a relay to a NON-CONNECTED GPIO.
- Add a switch to control the relay as;
```
    #define BUTTONx_PIN         p
    #define BUTTONx_MODE        BUTTON_PUSHBUTTON | BUTTON_SET_PULLUP | BUTTON_DEFAULT_HIGH
    #define BUTTONx_PRESS	BUTTON_MODE_ON
    #define BUTTONx_CLICK	BUTTON_MODE_OFF
    #define BUTTONx_DBLCLICK	BUTTON_MODE_OFF
    #define BUTTONx_LNGCLICK	BUTTON_MODE_OFF
    #define BUTTONx_LNGLNGCLICK	BUTTON_MODE_OFF
    #define BUTTONx_RELAY	y
```

Now, the state of RELAY "y" will follow the state of GPIO "p", and will be displayed both on the web UI, as well as being published to MQTT.

For the inadequacies;
- Garage door openers have more states than just "open" or "closed" and will take more than one GPIO to read. In particular, there is OPEN, there is CLOSED, there is PARTLY OPEN (when OPEN == CLOSED == false), there is RUNNING and STOPPED, there is UP and DOWN. I think that as a minimum level of input, you should implement OPEN, CLOSED, and RUNNING, which will take 3 GPIOs. Full implementation will take 4 GPIOs.
- Should implement some form of security in the event of losing MQTT. In particular, there should be logic in place to CLOSE the door upon loss of communications and/or TIMEOUT, i.e., if you open it remotely, it probably should close automatically after a couple of minutes in the event of a loss of WAN.

InduPrakash:
I added the support not to make the Espurna device a garage controller on its own but instead facilitate the same through SmartThings app (mine is at https://github.com/InduPrakash/SmartThings). The brain is the app and I would expect automatic closing to implemented there and not at Espurna end. Note that the simple garage door would back out on obstruction so just attempting a close after some delay is not sufficient. You are correct about the other states: opening and closing, they match the door event definition at SmartThings end and I implemented them in the app. Yes a contained operation would need 2 inputs but it turned out I already had a garage title sensor in place. I have been using this branch's code base on my garage with no issues since I submitted it but I do see your point; I don't see a need to merge this request.

lbdroid:
I think you missed the point about the auto-close feature. If you have a network failure, smart things will NOT be able to command it to close the door.

InduPrakash:
I do understand your point and I gave it a little bit of thought. Currently communication is only through well defined events send over MQTT and I would very much like to pass this action as a notification to the mobile app. I start feeling that this needs a more dedicated app. The merge request was to accomodate the MQTT events which SmartThings understands, more work would be neede to complete all what you mentioned. I do have a sort of backup as well, a daily nightly routine closes all doors at dusk which gets executed by the local SmartThings hub.

InduPrakash:
Showing GPIO state on web GUI was not my end goal, the change there is to define the MQTT message and indicate if the door sensor state is considered open or closed. The available magnetic sensors can be normally closed or opened.

The relay always needs to be momentarily toggled regardless of the button states. I do not understand your suggestion about tieing the relay to button. If the door was manually opened/closed, then the garage controller sends the MQTT message.  If the operation is invoked from SmartThings end, then garage controller responds by clicking the button.

Overall the MQTT subscriptions need to be specific to be recognized, currently I have 2 sensors subscribed at "smartthings/Garage Closed Sensor/contact" and "smartthings/Garage Open Sensor/contact" and a switch subscribed at "smartthings/Garage Door/switch". The contact sensors have states open/closed and the switch has on/off. The opening/closing are pseudo states determined on based of contact sensors.

PR #1084, opened 2018-07-27
Support for basic thermistor sensor

InduPrakash:
It uses the default analog input


xoseperez:
There is already support for a thermistor... check the NTCSensor code...

InduPrakash:
Well that was a waste then .. but at least now I have experience writing a sensor.

PR #1092, opened 2018-07-31
OTA Progress Debug Msg only on change

BauerPh:
no duplicate messages while OTA Update throug PlatformIO or Arduino IDE


xoseperez:
Is this really worth it? What harm do they do?

BauerPh:
They flood the weblog.. but it's only eye candy ;)

PR #1093, opened 2018-07-31
Change Button Settings in Web Interface

BauerPh:
It's working, but somone should review this please.


xoseperez:
Thank you. This should probably go to the `softcfg` branch where I'm working on configuring different hardware from the web interface.

Valcob:
@xoseperez, do we really need to change the GPIO pin from the UI. Lets say I have a PCB that has a button tight to GPIO 0 how wise is to allow user to change the pin assignment? What is the benefit of this being on the UI?

BauerPh:
I agree to that. You cant chance the GPIO with my commits. You can only change the connected button type (Pushbutton/Toggle Switch), enable/disable Pullups, set default high or low and the set actions for each event (Click, Double Click,...)
GPIO and the number of Buttons is set during compiling.

xoseperez:
The main reason to allow GPIO configuration from the web UI is that most boards will use the same binary image with just different configuration. I am trying to move from an "every board has his binary file" to "most boards share the same binary but using different configuration settings". The configuration settings can be downloaded and applied on top of a fresh installation.

xoseperez:
Sure. I meant the basic configuration for a given device can be applied on top of the fresh installation but after that, you will still be able to modify any configuration to suit your needs.

mcspr:
Closing due to age. API changes for configuration are in `dev`, WebUI stuff is still missing though.

Issue #1105, opened 2018-08-05
New device: Homecube 16A


hyteoo:
I got these from Amazon: https://www.amazon.de/gp/product/B07D7RVF56/ref=oh_aui_detailpage_o00_s01?ie=UTF8&psc=1
![homecube](https://user-images.githubusercontent.com/1378206/43683857-938e1346-989d-11e8-9e8b-e45975a9017d.PNG)
Nice little sockets, they have the following features:
- extremely compact
- 16A relay
- 4x RGB LEDs
- energy monitoring with HLW8012

Tested them with espurna, created a new hardware profile, so far they work great. It is similar to Blitzwolf but some pins differ.
I'll submit a PR shortly

There are 3 boards, sandwiched together like a H.
The ESP8266 is on a board based on [WT8266-S1](https://www.seeedstudio.com/WT8266-S1-FCC%26amp%3BCE%26amp%3BRoHS%26amp%3B-Wi-Fi-Module-based-on-ESP8266-p-2521.html).
The outer ring is illuminated by the 4 RGB LEDs, making the whole socket look very nice. Maybe some PWM would work nice here.

Here are some internal shots:

![20180804_184041](https://user-images.githubusercontent.com/1378206/43684137-298fc8d0-98a3-11e8-8638-200db44792d5.jpg)
![20180804_184101](https://user-images.githubusercontent.com/1378206/43683870-c2c7ea38-989d-11e8-8f0d-059c087c8c62.jpg)
![20180804_184106](https://user-images.githubusercontent.com/1378206/43683871-c2e4745a-989d-11e8-98eb-1f1b58ee3d7d.jpg)
![20180804_184120](https://user-images.githubusercontent.com/1378206/43683872-c3041c24-989d-11e8-8ee8-ebf055b63f43.jpg)
![20180804_184129](https://user-images.githubusercontent.com/1378206/43683873-c321d7a0-989d-11e8-82a1-d6d85310c7d5.jpg)
![20180804_213827](https://user-images.githubusercontent.com/1378206/43683875-c3448ca0-989d-11e8-95d2-27a90eb8e0c4.jpg)
![20180804_213901](https://user-images.githubusercontent.com/1378206/43683876-c362bd7e-989d-11e8-8a80-2098f4a5ef5d.jpg)

xoseperez:
Merged! I'm sure this is a great device.

flashy02:
One question: How can I open this Device without damaging?

hyteoo:
[...]

PR #1106, opened 2018-08-05
Added Homecube 16A

hyteoo:
Added Homecube 16A Wifi socket with RGB LED ring
See #1105


xoseperez:
Nice device!!! I think I should get one myself

pbriesch:
Where can I find the RX, TX, GND, VCC pins on the board?

hyteoo:
You can check the pinout here https://m.seeedstudio.com/productDetail/2521

PR #1107, opened 2018-08-05
Add support for VL53L1X sensor

ruimarinho:
Support for VL53L1X Time-of-Flight Distance Sensor Carrier with Voltage Regulator, 400cm Max.

> This sensor is a carrier/breakout board for ST’s VL53L1X laser-ranging sensor, which offers fast and accurate ranging up to 4 m. It uses the time of flight (ToF) of invisible, eye-safe laser pulses to measure absolute distances independent of ambient lighting conditions and target characteristics like color, shape, and texture (though these things will affect the maximum range). The VL53L1X also features a programmable region of interest (ROI), so the full field of view can be reduced or divided into multiple zones. Distance measurements can be read through a digital I²C interface. The board includes a 2.8 V linear regulator and level-shifters that allow it to work over an input voltage range of 2.6 V to 5.5 V, and the 0.1″ pin spacing makes it easy to use with standard solderless breadboards and 0.1″ perfboards. Learn more at [Pololu](https://www.pololu.com/product/3415).

### DIFF
--- a/code/espurna/config/types.h
+++ b/code/espurna/config/types.h
@@ -279,6 +279,7 @@
 #define SENSOR_SDS011_ID            0x27
 #define SENSOR_MICS2710_ID          0x28
 #define SENSOR_MICS5525_ID          0x29
+#define SENSOR_VL53L1X_ID           0x30
 
 //--------------------------------------------------------------------------------
 // Magnitudes
--- a/code/espurna/sensors/VL53L1XSensor.h
+++ b/code/espurna/sensors/VL53L1XSensor.h
@@ -1,0 +1,119 @@
+// -----------------------------------------------------------------------------
+// VL53L1X Sensor over I2C
+// Copyright (C) 2017-2018 by Xose Pérez <xose dot perez at gmail dot com>
+// -----------------------------------------------------------------------------
+
+#if SENSOR_SUPPORT && VL53L1X_SUPPORT
+
+#pragma once
+
+#undef I2C_SUPPORT
+#define I2C_SUPPORT 1 // Explicitly request I2C support.
+
+
+#include "Arduino.h"
+#include "I2CSensor.h"
+#include "VL53L1X.h"
+
+class VL53L1XSensor : public I2CSensor {
+
+    public:
+
+        // ---------------------------------------------------------------------
+        // Public
+        // ---------------------------------------------------------------------
+
+        VL53L1XSensor(): I2CSensor() {
+            _count = 1;
+            _sensor_id = SENSOR_VL53L1X_ID;
+            _vl53l1x = new VL53L1X();
+        }
+
+        ~VL53L1XSensor() {
+          delete _vl53l1x;
+        }
+
+        // ---------------------------------------------------------------------
+
+        void setDistanceMode(VL53L1X::DistanceMode mode) {
+          _vl53l1x->setDistanceMode(mode);
+        }
+
+        void setMeasurementTimingBudget(uint32_t budget_us) {
+          _vl53l1x->setMeasurementTimingBudget(budget_us);
+        }
+
+        void setInterMeasurementPeriod(unsigned int period) {
+          if (_inter_measurement_period == period) return;
+          _inter_measurement_period = period;
+          _dirty = true;
+        }
+
+        // ---------------------------------------------------------------------
+        // Sensor API
+        // ---------------------------------------------------------------------
+
+        void begin() {
+          if (!_dirty) {
+            return;
+          }
+
+          // I2C auto-discover
+          unsigned char addresses[] = {0x29};
+          _address = _begin_i2c(_address, sizeof(addresses), addresses);
+          if (_address == 0) return;
+
+          _vl53l1x->setAddress(_address);
+
+          if (!_vl53l1x->init()) {
+            return;
+          };
+
+          _vl53l1x->startContinuous(_inter_measurement_period);
+
+          _ready = true;
+          _dirty = false;
+        }
+
+        // Descriptive name of the sensor
+        String description() {
+            char buffer[21];
+            snprintf(buffer, sizeof(buffer), "VL53L1X @ I2C (0x%02X)", _address);
+            return String(buffer);
+        }
+
+        // Descriptive name of the slot # index
+        String slot(unsigned char index) {
+            return description();
+        };
+
+        // Type for slot # index
+        unsigned char type(unsigned char index) {
+            if (index == 0) return MAGNITUDE_DISTANCE;
+            return MAGNITUDE_NONE;
+        }
+
+        // Pre-read hook (usually to populate registers with up-to-date data)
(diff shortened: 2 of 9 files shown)


code/espurna/sensors/VL53L1XI2CSensor.h
```diff
@@ -0,0 +1,86 @@
...
+
+        // Initialization method, must be idempotent
+        void begin() {
+            // I2C auto-discover
+            unsigned char addresses[] = {0x29};
+            _address = _begin_i2c(_address, sizeof(addresses), addresses);
+            if (_address == 0) return;
+
+            _sensor = new VL53L1X();
+            _sensor->setDistanceMode(VL53L1X_DISTANCE_MODE);
```
ruimarinho: @xoseperez, is there any way this could be configured using the web ui and stored as a setting on the device's flash?

code/espurna/sensors/VL53L1XI2CSensor.h
```diff
@@ -0,0 +1,86 @@
...
+        // Pre-read hook (usually to populate registers with up-to-date data)
+        void pre() {
+            _error = SENSOR_ERROR_OK;
+        }
+
+        // Current value for slot # index
+        double value(unsigned char index) {
+            if (index != 0) return 0;
+
+            return _sensor->read();
```
ruimarinho: @xoseperez this is a blocking call. I am not sure if espurna supports asyncs events. I also have no idea on best practices..

code/espurna/sensors/VL53L1XI2CSensor.h
```diff
@@ -0,0 +1,86 @@
...
+
+#undef I2C_SUPPORT
+#define I2C_SUPPORT 1 // Explicitly request I2C support.
+
+
+#include "Arduino.h"
+#include "I2CSensor.h"
+#include "VL53L1X.h"
+
+class VL53L1XI2CSensor : public I2CSensor {
```
ruimarinho: @xoseperez is the convention `VL53L1XI2CSensor` or `VL53L1XSensor`?

code/platformio.ini
```diff
@@ -90,6 +90,7 @@ lib_deps =
     rc-switch
     https://github.com/LowPowerLab/RFM69#1.1.3
     https://github.com/xoseperez/Time
+    https://github.com/pololu/vl53l1x-arduino
```
ruimarinho: @xoseperez I will pin this right before/if the PR is merged.

xoseperez:
Great, just one thing: I prefer to use International System Units always as return values from the sensor classes. After that, a filter can be applied for units conversion (current conversion filter is horrible, but still).

ruimarinho:
@xoseperez do you mean under `magnitudeUnits` and returning `MAGNITUDE_DISTANCE_METERS` on the sensor itself?

code/espurna/sensors/VL53L1XSensor.h
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        // Descriptive name of the slot # index
+        String slot(unsigned char index) {
+            return description();
+        };
+
+        // Type for slot # index
+        unsigned char type(unsigned char index) {
+            if (index == 0) return MAGNITUDE_DISTANCE_MILLIMETERS;
```
xoseperez: Sensor should return meters for distance (that's the IS unit)

code/espurna/sensors/VL53L1XSensor.h
```diff
@@ -0,0 +1,108 @@
...
+        // Pre-read hook (usually to populate registers with up-to-date data)
+        void pre() {
+            _error = SENSOR_ERROR_OK;
+        }
+
+        // Current value for slot # index
+        double value(unsigned char index) {
+            if (index != 0) return 0;
+
+            return _vl53l1x->read();
```
xoseperez: This is a blocking call, how long does it take on average?

code/espurna/sensors/VL53L1XSensor.h
```diff
@@ -0,0 +1,108 @@
...
+          _inter_measurement_period = period;
+          _dirty = true;
+        }
+
+        // ---------------------------------------------------------------------
+        // Sensor API
+        // ---------------------------------------------------------------------
+
+        void begin() {
+          if (!_dirty) {
```
xoseperez: I2C sensors can benefit from address auto-discover feature. Check the BMX280 for instance.

ruimarinho: Will take a look. Mind if I add this as a separate PR?

ruimarinho:
@xoseperez addressed some comments. Let me know before merging so I can squash the commits.

code/espurna/sensors/VL53L1XSensor.h
```diff
@@ -0,0 +1,108 @@
...
+        // Pre-read hook (usually to populate registers with up-to-date data)
+        void pre() {
+            _error = SENSOR_ERROR_OK;
+        }
+
+        // Current value for slot # index
+        double value(unsigned char index) {
+            if (index != 0) return 0;
+
+            return _vl53l1x->read() / 1000.0;
```
ruimarinho: @xoseperez I believe this blocks for a maximum of `_inter_measurement_period` (which should be always as long as the timing budget - I may add this guard). We can also set a timeout. Is there any example for using interrupts for reading data in espurna?

mcspr: hmm. reading about read(), it has an optional arg blocking=true/false, which you do not specify and it defaults to true. _vl531lx->dataReady() is a check for non-blocking mode (which could be in pre() ?) and value() will just return cached value.

code/espurna/config/types.h
```diff
@@ -269,36 +269,37 @@
...
+#define MAGNITUDE_PM2dot5                16
+#define MAGNITUDE_PM10                   17
+#define MAGNITUDE_CO2                    18
+#define MAGNITUDE_LUX                    19
+#define MAGNITUDE_UV                     20
+#define MAGNITUDE_DISTANCE               21
+#define MAGNITUDE_HCHO                   22
+#define MAGNITUDE_GEIGER_CPM             23
+#define MAGNITUDE_GEIGER_SIEVERT         24
+#define MAGNITUDE_COUNT                  24
```
xoseperez: This should be 25

code/espurna/sensors/VL53L1XSensor.h
xoseperez: @mcspr suggestion is better than using interrupts (which you can, check the EventSensor, for instance). Non-blocking calls are always preferred.

ruimarinho:
@xoseperez @mcspr all comments addressed! Reads are now non-blocking.

ruimarinho:
@xoseperez if you want to merge this one before EZO pH then I can rebase it on top.

PR #1112, opened 2018-08-06
Add support for VEML6075

ruimarinho:
> The VEML6075 senses UVA and UVB light and incorporates photodiode, amplifiers, and analog / digital circuits into a single chip using a CMOS process. When the UV sensor is applied, it is able to detect UVA and UVB intensity to provide a measure of the signal strength as well as allowing for UVI measurement.

> The VEML6075 provides excellent temperature compensation capability for keeping the output stable under changing temperature. VEML6075’s functionality is easily operated via the simple command format of I2C (SMBus compatible) interface protocol. VEML6075’s operating voltage ranges from 1.7 V to 3.6 V. VEML6075 is packaged in a lead (Pb)-free 4 pin OPLGA package which offers the best market-proven reliability.

*via [datasheet](https://www.vishay.com/docs/84304/veml6075.pdf)*

screen shot 2018-11-02 at 00 18 38


code/espurna/config/types.h
```diff
@@ -269,6 +269,7 @@
 #define SENSOR_SENSEAIR_ID          0x24
 #define SENSOR_GEIGER_ID            0x25
 #define SENSOR_NTC_ID               0x26
+#define SENSOR_VEML6075_ID          0x28
```
ruimarinho: @xoseperez assumes VL53L1X gets merged first :)

code/html/custom.js
```diff
@@ -60,7 +61,7 @@ function magnitudeType(type) {
         "Current", "Voltage", "Active Power", "Apparent Power",
         "Reactive Power", "Power Factor", "Energy", "Energy (delta)",
         "Analog", "Digital", "Event",
-        "PM1.0", "PM2.5", "PM10", "CO2", "Lux", "UV", "Distance" , "HCHO",
+        "PM1.0", "PM2.5", "PM10", "CO2", "Lux", "UVA", "UVB", "UV Index", "Distance" , "HCHO",
```
ruimarinho: @xoseperez not sure if we should keep `UVI` for consistency here.

code/espurna/sensors/VEML6075Sensor.h
```diff
@@ -0,0 +1,96 @@
...
+        double value(unsigned char index) {
+            if (index == 0) return _veml6075->a();
+            if (index == 1) return _veml6075->b();
+            if (index == 2) return _veml6075->index();
+
+            return 0;
+        }
+
+        void setIntegrationTime(VEML6075::veml6075_uv_it_t integration_time) {
+          _veml6075->setIntegrationTime(integration_time);
```
ruimarinho: @xoseperez is my understanding correct that the `_dirty` flag is used when initialising a sensor requires pre-configuring certain attributes?

code/espurna/config/progmem.h
```diff
@@ -233,7 +238,9 @@ PROGMEM const char magnitude_pm2dot5_topic[] = "pm2dot5";
 PROGMEM const char magnitude_pm10_topic[] = "pm10";
 PROGMEM const char magnitude_co2_topic[] = "co2";
 PROGMEM const char magnitude_lux_topic[] = "lux";
-PROGMEM const char magnitude_uv_topic[] = "uv";
+PROGMEM const char magnitude_uv_a_topic[] = "uva";
+PROGMEM const char magnitude_uv_b_topic[] = "uvb";
+PROGMEM const char magnitude_uv_index_topic[] = "uv";
```
ruimarinho: @xoseperez should this be `uvi`? Probably a BC..

code/espurna/config/sensors.h
```diff
@@ -607,6 +607,25 @@
...
+#define VEML6075_SUPPORT                  0
+#endif
+
+#ifndef VEML6075_INTEGRATION_TIME
+#define VEML6075_INTEGRATION_TIME         VEML6075::IT_100MS        // The time, in milliseconds, allocated for a single
+#endif                                                              // measurement. A longer timing budget allows for more
+                                                                    // accurate results at the cost of power.
+
+#ifndef VEML6075_DYNAMIC_MODE
+#define VEML6075_DYNAMIC_MODE             VEML6075::DYNAMIC_NORMAL  // The dynamic mode can either be normal or high. In high
```
ruimarinho: @xoseperez can these options be configured via the web ui?

code/espurna/sensors/VEML6075Sensor.h
```diff
@@ -0,0 +1,96 @@
...
+        // ---------------------------------------------------------------------
+
+        VEML6075Sensor(): I2CSensor() {
+            _count = 3;
+            _sensor_id = SENSOR_VEML6075_ID;
+            _veml6075 = new VEML6075();
+        }
+
+        ~VEML6075Sensor() {
+          delete _veml6075;
```
ruimarinho: @xoseperez I don't know if this is enough to have on the destructor..

code/espurna/config/progmem.h
```diff
@@ -208,7 +211,9 @@ PROGMEM const unsigned char magnitude_decimals[] = {
     3, 0, 0, 0, 0, 0, 0, 0, // Power decimals
     0, 0, 0, // analog, digital, event
     0, 0, 0, // PM
-    0, 0, 3, 3, 0,
+    0, 0,
+    0, 0, 3, // UVA, UVB, UV Index
```
ruimarinho: @xoseperez UVA, UVB and UVI as returned as floats. Should we keep them with decimals when presenting?

lobradov:
Would be good to compile this sensor in one of the travis builds to understand if they compile, but also to ensure that future updates will not break anything.

code/espurna/sensors/VEML6075Sensor.h
xoseperez: Yes, set _dirty to true on when a parameter change requires the master code to call begin again

code/espurna/sensors/VEML6075Sensor.h
xoseperez: Should be. I have not dive into the VEML6075 library code.

code/espurna/config/types.h
xoseperez: Mmm... ok....

code/espurna/config/sensors.h
xoseperez: Yes, but for now it's OK. I will run a migration process on all sensors sometime in the future.

code/espurna/config/progmem.h
xoseperez: I have seen it called UVI somewhere, so I guess it's better to say UVI

code/espurna/config/progmem.h
xoseperez: Even if they are returned as floats, the decimals value must be that that makes more sense. If UVA and UVB are integers then set it to 0.

ruimarinho:
@xoseperez all comments addressed and PR rebased.

PR #1113, opened 2018-08-06
Add support for wemos-d1mini dev board

ruimarinho:
I don't know if this is actually required since espurna already has the `wemos-d1mini-relayshield` environment but since I don't operate my wemos with this shield, I thought it could be helpful for others to have this one available as well.


ruimarinho:
Actually, I think if the board is set `board: esp12e` I can't get it to work properly.

PR #1116, opened 2018-08-07
Add support for NewPing

ruimarinho:
* Works with many different ultrasonic sensor models: HC-SR04, SRF05,  SRF06, DYP-ME007, JSN-SR04T & Parallax PING)))™.
* Built-in digital filter method ping_median() for easy error correction (use `SONAR_ITERATIONS` build flag to configure).
* Allows setting of a maximum distance where pings beyond that distance are read as no ping or clear (use `SONAR_MAX_DISTANCE` build flag).

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -132,6 +132,9 @@
     #define I2C_SDA_PIN         12  // D6
     #define I2C_SCL_PIN         14  // D5
 
+    #define SONAR_TRIGGER       12  // D6
+    #define SONAR_ECHO          13  // D7
+
 #elif defined(WEMOS_D1_TARPUNA_SHIELD)
 
     // Info
@@ -2665,10 +2668,10 @@
     #define EVENTS_SUPPORT      1
     #define EVENTS_PIN          6
 
-    // HC-RS04
-    #define HCSR04_SUPPORT      1
-    #define HCSR04_TRIGGER      7
-    #define HCSR04_ECHO         8
+    // Sonar
+    #define SONAR_SUPPORT       1
+    #define SONAR_TRIGGER       7
+    #define SONAR_ECHO          8
 
     // MHZ19
     #define MHZ19_SUPPORT       1
--- a/code/espurna/sensor.ino
+++ b/code/espurna/sensor.ino
@@ -476,11 +476,13 @@ void _sensorLoad() {
     }
     #endif
 
-    #if HCSR04_SUPPORT
+    #if SONAR_SUPPORT
     {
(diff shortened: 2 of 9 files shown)


mcspr:
Quick note: TRAVIS02 (in hardware.h) still has HCSR04_SUPPORT.

code/espurna/sensor.ino
```diff
@@ -476,11 +476,13 @@ void _sensorLoad() {
...
 
-    #if HCSR04_SUPPORT
+    #if SONAR_SUPPORT
     {
-        HCSR04Sensor * sensor = new HCSR04Sensor();
-        sensor->setTrigger(HCSR04_TRIGGER);
-        sensor->setEcho(HCSR04_ECHO);
+        SonarSensor * sensor = new SonarSensor();
+        sensor->setEcho(SONAR_ECHO);
+        sensor->setIterations(SONAR_PING_ITERATIONS);
```
mcspr: Typo. Should be `SONAR_ITERATIONS`

code/espurna/config/hardware.h
```diff
@@ -112,6 +112,9 @@
     #define I2C_SDA_PIN         12  // D6
     #define I2C_SCL_PIN         14  // D5
 
+    #define SONAR_TRIGGER       12  // D6
+    #define SONAR_ECHO          13  // D7
+
```
mcspr: Maybe just keep as defaults in sensor.h?

xoseperez: Agree, these should not be here.

code/espurna/sensor.ino
xoseperez: Right

code/espurna/config/hardware.h
mcspr: Oops. Pins from wemos do not include hw.h 7 and 8 (6-11 are not mapped to pins at all so they do nothing): https://github.com/esp8266/Arduino/blob/master/variants/d1/pins_arduino.h
What i meant to say, change the default to these and just not define it only for wemos hw :) Sry for confusion

ruimarinho: @xoseperez I don't think these will work on Wemos — https://github.com/xoseperez/espurna/blob/dev/code/espurna/config/hardware.h#L2669-L2671. Wasn't this change correct after all?

mcspr: @ruimarinho nah you linked travis-only hw that is only a built test, no real boards involved. these are fine https://github.com/xoseperez/espurna/blob/4f61d3aa7e2e6526fdb518a9f84785b055ac0ab7/code/espurna/config/sensors.h#L570-L576

PR #1135, opened 2018-08-14
Add support for Lohas 9W bulbs

steveway:
It seems to be very similar to the ITead Sonoff B1, so it's based on that.
This should be the official product page if I'm not mistaken:
http://www.lohasled.com/9w-a19-p00064p1.html

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -743,6 +743,25 @@
     #define MY92XX_MAPPING      4, 3, 5, 0, 1
     #define LIGHT_WHITE_FACTOR  (0.1)                    // White LEDs are way more bright in the B1
 
+#elif defined(LOHAS_9W)
+
+    // Info
+    #define MANUFACTURER        "LOHAS"
+    #define DEVICE              "E27_9W"
+    #define RELAY_PROVIDER      RELAY_PROVIDER_LIGHT
+    #define LIGHT_PROVIDER      LIGHT_PROVIDER_MY92XX
+    #define DUMMY_RELAY_COUNT   1
+
+    // Light
+    #define LIGHT_CHANNELS      5
+    #define MY92XX_MODEL        MY92XX_MODEL_MY9231
+    #define MY92XX_CHIPS        2
+    #define MY92XX_DI_PIN       13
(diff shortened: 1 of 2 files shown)


Wikibear:
Tried out today... Found ITead is similar but not same.

`#elif defined(ITEAD_SONOFF_B1)

    // Info
    #define MANUFACTURER        "LOHAS"
    #define DEVICE              "RGBW 9W"
    #define RELAY_PROVIDER      RELAY_PROVIDER_LIGHT
    #define LIGHT_PROVIDER      LIGHT_PROVIDER_MY92XX
    #define DUMMY_RELAY_COUNT   1

    // Light
    #define LIGHT_CHANNELS      5
    #define MY92XX_MODEL        MY92XX_MODEL_MY9231
    #define MY92XX_CHIPS        2
    #define MY92XX_DI_PIN       13
    #define MY92XX_DCKI_PIN     15
    #define MY92XX_COMMAND      MY92XX_COMMAND_DEFAULT
    #define MY92XX_MAPPING      0, 1, 2, 3, 4
    #define LIGHT_WHITE_FACTOR  (0.1)             `

You have then 5 Channels: R,G,B and 2 White LED's and at this Point i don't understand if that is well or not.

solbergw:
I tried this out on E26 9W, RGB Works, White temperature does not change.  Are there further possible outputs that can be mapped beyond 0-4.

steveway:
Mhh, I moved recently and can't test my light at the moment.
I need to reflash it for that, currently I don't have the ability because I'm missing a place to solder here.
I think if I remember correctly the temperature is controlled by two different sets of white LEDs, I'm not sure if Espurna does support controlling that this way yet.

Wikibear:
Are you sure that there is a real white channel??? This Lamp ist RGB only...

solbergw:
I cracked open the bulb to find out what was really happening.  Channels 0-2 control the RGB LEDs, channel 3 controls (4 white LEDs) and channel 4 controls (4 white LEDs).  All 8 of the white LEDs are the same temperature.   With the factory installed firmware, I set the bulb to a warm temperature and it uses the white LEDs + the red LEDs.  When I set the bulb to a cool temperature it only uses the white LEDs.

PR #1176, opened 2018-09-02
added SDS011 sensor support

derlucas:
i added support for the SDS011 sensor. It can measure P2.5 and P10 particulates.
Basically i cloned MHZ-19 and put code from https://github.com/ricki-z/SDS011 onto it.


xoseperez:
Great! Thank you

PR #1188, opened 2018-09-10
Define HOMEASSISTANT_PAYLOAD_AVAILABLE for payload_available (instead of HOMEASSISTANT_PAYLOAD_ON)

Yonsm:
Since `payload_on` and `payload_available` is not the same on HomeAssistant default implementation, I suggest use HOMEASSISTANT_PAYLOAD_AVAILABLE for payload_available (instead of HOMEASSISTANT_PAYLOAD_ON).

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -919,6 +919,14 @@
 #define HOMEASSISTANT_PAYLOAD_OFF   "0"         // Payload for OFF and unavailable messages
 #endif
 
+#ifndef HOMEASSISTANT_PAYLOAD_AVAILABLE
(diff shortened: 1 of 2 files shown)


Yonsm:
Furthermore, the default value in HomeAssistant [mqtt.py](https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/light/mqtt.py) are list as below:
`payload_on`: "ON"
`payload_off`: "OFF"
`payload_available`: "online"
`payload_not_available`: "offline"

That is not same as our Espurna's `1`/`0`. Maybe you can change these default value in the future:)

PR #1191, opened 2018-09-11
All DST Zones

Diramu:
I've added a new DSTZone OTHERS.  setDSTZone has a second form, that take start month/week/day and end month/week/day and hour.  week == 5 means the last week of the month.
Calls to setDSTZone with DST_ZONE_EU or DST_ZONE_USA is map to the new fully expanded version of for setDSTZone().  Instead of saving the _dstZOne, time_t values of _dstStart & _dstEnd (starting and ending time for summer time) are saved in the class.  This make the testing  of summer time simple, and works for the southern hemisphere.   This change is required by a similar change to NtpClientlib.


mcspr:
I would suggest slightly different approach at keeping timezone data. Take a look at how arduino/sntp does it by using TZ variable, which describes exactly the things you want:
https://github.com/mhightower83/Arduino-IDE-Tweaks/wiki/SNTP,-Time,-TZ,-libc
http://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

Diramu:
New version ready, with pull posix generated timezones, and ntpclient to match

PR #1195, opened 2018-09-12
Rename cont_t for Core 2.4.2 and higher

mcspr:
https://github.com/xoseperez/espurna/commit/88e2eccccef3846845c2b43d5b58e3a24c61ef2d#diff-bd6de869be74ac5eaf90f23e4b75b857 added extern g_cont:
https://github.com/xoseperez/espurna/blob/88e2eccccef3846845c2b43d5b58e3a24c61ef2d/code/espurna/utils.ino#L9-L12
but it is no longer defined in 'master' of esp8266/Arduino, now it is `cont_t* g_pcont`:
https://github.com/esp8266/Arduino/commit/5d5ea92a4d004ab009d5f642629946a0cb8893dd#diff-3fa12668b289ccb95b7ab334833a4ba8L35

This PR:
- added define check for core version when using g_cont/g_pcont
- updated travis envs to use newest platform, so this is tested

Issue #1205, opened 2018-09-17
Build project error - utils.ino:23:5: error: expected unqualified-id before '||' token


icevoodoo:
Build project error see below:

Processing itead-sonoff-th (framework: arduino; platform: espressif8266@1.5.0; board: esp01_1m)
-------------------------------------------------------------------------------------------------------------------------------
Verbose mode can be enabled via `-v, --verbose` option
PLATFORM: Espressif 8266 > Espressif Generic ESP8266 ESP-01 1M
SYSTEM: ESP8266 80MHz 80KB RAM (1MB Flash)
Converting espurna.ino
Library Dependency Finder -> http://bit.ly/configure-pio-ldf
LDF MODES: FINDER(chain) COMPATIBILITY(soft)
Collected 51 compatible libraries
Scanning dependencies...
Dependency Graph
|-- <ArduinoJson> 5.13.2
|-- <AsyncMqttClient> 0.8.1 #ddbf4d1
|   |-- <ESPAsyncTCP> 1.1.3 #55cd520
|   |   |-- <ESP8266WiFi> 1.0
|-- <Brzo I2C> 1.3.2
|-- <DebounceEvent> 2.0.4 #4681547
|-- <EEPROM_Rotate> 0.9.1 #9c7d14f
|   |-- <EEPROM> 1.0
|-- <Embedis> 1.2.0
|-- <Encoder>
|-- <EspSoftwareSerial> 3.4.1 #5378868
|-- <ESPAsyncTCP> 1.1.3 #55cd520
|   |-- <ESP8266WiFi> 1.0
|-- <ESP Async WebServer> 1.1.0 #05306e4
|   |-- <Hash> 1.0
|   |-- <ESPAsyncTCP> 1.1.3 #55cd520
|   |   |-- <ESP8266WiFi> 1.0
|   |-- <ArduinoJson> 5.13.2
|   |-- <ESP8266WiFi> 1.0
|-- <FauxmoESP> 3.0.1 #9a03a61
|   |-- <ESP8266WiFi> 1.0
|   |-- <ESPAsyncTCP> 1.1.3 #55cd520
|   |   |-- <ESP8266WiFi> 1.0
|-- <HLW8012> 1.1.0 #cc3baec
|-- <IRremoteESP8266> 2.2.0 #4d57a2b
|-- <JustWifi> 2.0.2 #1367a8b
|   |-- <ESP8266WiFi> 1.0
[...]

PR #1208, opened 2018-09-17
Inline getFreeStack and mention 2.5.0 helper method

mcspr:
Amending #1195 with new ESP.getFreeContStack
Definition instead of function, -6 lines

### DIFF
--- a/code/espurna/utils.ino
+++ b/code/espurna/utils.ino
@@ -6,31 +6,6 @@ Copyright (C) 2017-2018 by Xose Pérez <xose dot perez at gmail dot com>
 
 */
 
-// Core version 2.4.2 and higher changed the cont_t structure to a pointer:
-// https://github.com/esp8266/Arduino/commit/5d5ea92a4d004ab009d5f642629946a0cb8893dd#diff-3fa12668b289ccb95b7ab334833a4ba8L35
-#if defined(ARDUINO_ESP8266_RELEASE_2_3_0) \
-    || defined(ARDUINO_ESP8266_RELEASE_2_4_0) \
-    || defined(ARDUINO_ESP8266_RELEASE_2_4_1)
-extern "C" {
-    #include <cont.h>
-    extern cont_t g_cont;
-
-}
-
-unsigned int getFreeStack() {
-    return cont_get_free_stack(&g_cont);
-}
-#else
-extern "C" {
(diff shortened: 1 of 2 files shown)


code/espurna/utils.ino
```diff
@@ -6,31 +6,6 @@ Copyright (C) 2017-2018 by Xose Pérez <xose dot perez at gmail dot com>
...
-extern "C" {
-    #include <cont.h>
-    extern cont_t g_cont;
-
-}
-
-unsigned int getFreeStack() {
-    return cont_get_free_stack(&g_cont);
-}
-#else
```
Misiu: can we add elif here just like in https://github.com/xoseperez/espurna/pull/1208/files#diff-52135e36de9de9ff791c1184a170b897R50, this way for 2.5.0 we won't have that extern "C" and rest.
Sorry if this is a bumb comment, I recently start learning Arduino and stuff.

mcspr: much less things on the screen was the main reason.
i am not an expert on c++, just felt like a better approach :) and as it turns out from #1205 all that multiline constructs inside .ino broke compilation when on Windows. moving to header also helped with that

mcspr:
After reading the source for [cont_init](https://github.com/esp8266/Arduino/blob/1a44f79a9e26ebb72900caef8e03a2b7df78724d/cores/esp8266/cont_util.c#L29-L42) and [cont_get_free_stack](https://github.com/esp8266/Arduino/blob/1a44f79a9e26ebb72900caef8e03a2b7df78724d/cores/esp8266/cont_util.c#L50-L61) I think getFreeStack() should be renamed to something like getTopUsedStack(), because it does not actually return currently used stack, but max amount used since startup.

Looking at postmortem code, current stack pointer can be retrieved via asm("a1") (and note the check for system / cont stack location). My basic understanding of this - sp/"a1" is "top", so just subtracting it with stack start will get used amount.
https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_postmortem.c#L150-L159
Poking around tasmota code I did find section using it for exactly that:
https://github.com/arendst/Sonoff-Tasmota/blob/e4ebee738b437fcc101f80e2f8d7df119aea9f02/sonoff/xdrv_99_debug.ino#L219-

PR #1219, opened 2018-09-22
NPM maintenance

mcspr:
* [Use 'npm ci' instead of 'install --only=dev'](https://docs.npmjs.com/cli/ci)
> This command is similar to npm-install, except it's meant to be used in automated environments such as test platforms, continuous integration, and deployment. It can be significantly faster than a regular npm install by skipping certain user-oriented features. It is also more strict than a regular install, which can help catch errors or inconsistencies caused by the incrementally-installed local environments of most npm users.
* bump npm to latest version in .travis.yml, because 'npm ci' is not available using it's default npm@5.5.1
* general update for package.json + package-lock.json (also has new version/resolve keys. can't just use old one) and remove unused uglify + del
* store datatables.js images that gulp is always complaining about https://cdn.datatables.net/1.10.16/images/sort_asc_disabled.png
https://cdn.datatables.net/1.10.16/images/sort_desc_disabled.png


code/package.json
```diff
@@ -1,25 +1,23 @@
 {
   "name": "esp8266-filesystem-builder",
-  "version": "0.2.1",
+  "version": "0.2.2",
```
inverse: Is version even required?

mcspr: npm docs says no. and neither is name. both can be safely removed.

PR #1220, opened 2018-09-22
Fix-up style and form validation logic

mcspr:
Amend #1182 and #1203 
-  attr() returns undefined with no attribute, method never set 'original'. 🤦‍♂️ likely happened mistaking getAttribute, which does return null for both present and unset attr (and not checking which branch was used locally, which was with native js getAttribute/hasAttribute)
- restoring old check for first password field length, which was accidentally removed. now '#layout' screen wont always fail password check
- also check for password validity before sending. this will issue client-side error instead of sending to device on '#password' screen

*I'll start looking into how to test ui interaction/presentation to avoid this in the future... And be more careful*

*edit:* minor css issue that i noticed from password ui update, it did not apply properly and neither '#layout' or '#password' were targets. default selector centered everything
*edit2:* forcepushed over wrong change. type='submit' introduced some questionable 'features', so that css is not needed anymore

### DIFF
--- a/code/html/custom.js
+++ b/code/html/custom.js
@@ -164,21 +164,35 @@ function validatePassword(password) {
     );
 }
 
-function validateForm(form) {
+function validateFormPasswords(form) {
+    var passwords = $("input[name='adminPass1'],input[name='adminPass2']", form);
+    var adminPass1 = passwords.first().val(),
+        adminPass2 = passwords.last().val();
 
-    // password
-    var adminPass1 = $("input[name='adminPass']", form).first().val();
-    if (!validatePassword(adminPass1)) {
-        alert("The password you have entered is not valid, it must be 8..63 characters and have at least 1 lowercase and 1 uppercase / number!");
-        return false;
+    var formValidity = passwords.first()[0].checkValidity();
+    if (formValidity && (adminPass1.length === adminPass2.length === 0)) {
+        return true;
     }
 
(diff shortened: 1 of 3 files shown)


code/html/custom.js
```diff
@@ -164,21 +168,26 @@ function validatePassword(password) {
...
+function validateFormPasswords(form) {
+    var passwords = $("input[name='adminPass'],input[name='adminPass_confirm']", form);
 
-    // password
-    var adminPass1 = $("input[name='adminPass']", form).first().val();
-    if (!validatePassword(adminPass1)) {
+    var adminPass1 = passwords.first().val(),
+        adminPass2 = passwords.last().val();
+
+    if ((adminPass1.length > 0) && !validatePassword(adminPass1)) {
```
inverse: It would skip this check if `adminPass1` was an empty string right?

mcspr: True, I was following original check logic. On '#password' layout it falls through, sends the data and returns 'No changes detected' from device.

General '#layout' can have empty password fields, no need to alert. But I think '#password' should have 'required' attrs on inputs, so this check becomes 'required' && length>0 or just 'form/input .checkValidity()' (which btw I remember finding out about and briefly trying it at #888 - form validation without alerts, just highlighting hints)

inverse: I totally think going for the HTML5 input validation being the best bet! This validation is still JS and can be turned off by the user so ideally there should be some backend validation going on too.

PR #1237, opened 2018-09-30
(partial) fix for YiDian XS-SSA05 configs

ducky64:
Developed against a [SM800-type device](https://fccid.io/2ANOO-SM800/Internal-Photos/Internal-Photos-3601477) (more-or-less equivalent Amazon products [1](https://www.amazon.com/Monitoring-Control-Smartphone-Compatibles-Assistant/dp/B0793RSYTV) [2](https://www.amazon.com/gp/product/B07837B9C1)), which has a similar circuit structure (ESP8266 module, pin assignments, LED polarity, switch polarity, energy monitoring chips and resistor values) to the XS-SSA05.

Changes:
- The LEDs are active-low (they are connected to the MCU pin on the low side)
- There are two LEDs, one red, one blue. I don't know if you have a convention for which one should be LED1 (which seems the only one editable from the web UI)
- The switch is active-low (pressing shorts to ground, pulled high by external circuitry)
- The energy monitor is a HJL-01 chip, and the SEL / interrupt polarity constants are changed accordingly. The voltage readings seem accurate.

Problems remaining:
- The energy monitor reading is highly inaccurate (comparing against a Kill-A-Watt P4400.01)
  - The current reads about half of what the Kill-A-Watt reads (and the apparent power is ~half of the Kill-A-Watt in VA mode)
  - The active power is twice of what the Kill-A-Watt reads in Watt mode
  - The reactive power just reads zero, even on loads where the Kill-A-Watt reports a PF of <0.6. This might just be an artifact of the active/apparent power being super off
- **This suggests that there's some different calibration constants between the HJL-01 and HLW8012 chips. I can't find any information for the HJL-01 chip, though...**

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -1932,7 +1932,7 @@
 
     // Buttons
     #define BUTTON1_PIN			13
-    #define BUTTON1_MODE		BUTTON_PUSHBUTTON
+    #define BUTTON1_MODE		BUTTON_PUSHBUTTON | BUTTON_DEFAULT_HIGH
     #define BUTTON1_RELAY		1
 
     // Relays
@@ -1940,9 +1940,14 @@
     #define RELAY1_TYPE			RELAY_TYPE_NORMAL
 
     // LEDs
-    #define LED1_PIN			4
-    #define LED1_PIN_INVERSE	0
-
+    #define LED1_PIN			0  // red
+    #define LED1_PIN_INVERSE	1
+    #define LED1_MODE           LED_MODE_WIFI
+    
+    #define LED2_PIN			15  // blue
+    #define LED2_PIN_INVERSE	1
+    #define LED2_MODE           LED_MODE_RELAY
+    
     // HLW8012
     #ifndef HLW8012_SUPPORT
     #define HLW8012_SUPPORT     1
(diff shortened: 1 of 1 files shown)


mcspr:
Hey! It seems there are already custom ratios for HJL defined. See:
https://github.com/xoseperez/espurna/blob/daf786421262d558ca9bee6a1815d9ed6372cbba/code/espurna/config/hardware.h#L2511-L2523

ducky64:
Those don't seem particularly accurate either, I trued both by setting the ratio and also by additionally defining the `HLW8012_VOLTAGE_R_UP` and `HLW8012_CURRENT_R`.

Read voltage is 0.77 of the Kill-A-Watt, read active power is 0.6 of the Kill-A-Watt, and read apparent power is 0.66 of the Kill-A-Watt. Power factor is in the right ballpark though (within 10%).

I also measured the voltage sense resistors on both one of the SM800-type boards and a XS-SSA05, they're within 1% of the labeled resistance, so this inaccuracy doesn't appear to be from bad component tolerances.

ducky64:
(that being said, I don't know how much to trust the Kill-A-Watt, since its voltage consistently reads high by ~10% compared to a Fluke multimeter. I don't have anything to measure mains current safely though.)

xoseperez:
Those ratios in the Blitzwolf are based on experimental calculations and might be different from one device to another, even of the same branch. But they will probably provide a close approximation and then you can use the calibration features in the web interface.

Thank you for your contribution!

Issue #1239, opened 2018-10-01
"Clear counts" on rfm69 does not reset node count properly.


Trickx:
Issue: If the button "Clear counts" is pressed, the node count on the status page does not start to count anymore.

Reason: The reset code does not reset .count.
```
void _rfm69Clear() {
    for(unsigned int i=0; i<255; i++) {
        _rfm69_node_info[i].duplicates = 0;
        _rfm69_node_info[i].missing = 0;
    }
    _rfm69_node_count = 0;
    _rfm69_packet_count = 0;
}
```

Solution:

```
void _rfm69Clear() {
    //for(unsigned int i=0; i<255; i++) {
    for(unsigned int i=0; i<(sizeof(_rfm69_node_info)/sizeof(_node_t)); i++) {
        _rfm69_node_info[i].duplicates = 0;
        _rfm69_node_info[i].missing = 0;
        _rfm69_node_info [i].count = 0;
    }
    _rfm69_node_count = 0;
    _rfm69_packet_count = 0;
}
```

lobradov:
Good catch!
Care to raise a PR for a fix?

__
//_

Trickx:
I would, but my espurna-fork is currently being used for other features - not ready for PR.
That's why I posted the fixed code here.

lobradov:
You can always have a new branch :) 
Or even use online editor for small changes... 
__
//_

Trickx:
-> PR #1241

xoseperez:
Merged into `dev`. Thank you!

PR #1241, opened 2018-10-01
-Fix for issue #1239 

Trickx:
"Clear counts" on rfm69 does not reset node count properly.

### DIFF
--- a/code/espurna/rfm69.ino
+++ b/code/espurna/rfm69.ino
@@ -235,9 +235,10 @@ void _rfm69Loop() {
 }
 
 void _rfm69Clear() {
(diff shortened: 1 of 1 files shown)


code/espurna/rfm69.ino
```diff
@@ -235,9 +235,10 @@ void _rfm69Loop() {
 }
 
 void _rfm69Clear() {
-    for(unsigned int i=0; i<255; i++) {
+    for(unsigned int i=0; i<(sizeof(_rfm69_node_info) / sizeof(_node_t)); i++) {
```
xoseperez: `_rfm69_node_info` is a static size array, wouldn't this always be 255?

PR #1247, opened 2018-10-05
Add hyphen to valid wifi password chars

djwmarcx:
Currently is missing in the validation regexp, but it is acceptable at WPA2 passwords


djwmarcx:
Need to verify, but seems this password validation is just for the admin pass not for de WPA 2

Issue #1244, opened 2018-10-05
Save button in Web UI is unresponsive on latest dev


xoseperez:

mcspr:
This form checking haunts me.

https://github.com/xoseperez/espurna/blob/ba4d4be8108ec843a0037db80055301dc3ae541c/code/html/custom.js#L173
evaluates to `true === 0` -> `false`

```js
var zeroLen = [adminPass1.length, adminPass2.length].every(function(val) { return val === 0; });
if (formValidity && zeroLen) {
```
Just to verify, this should be equivalent in real js (not python-like...)

xoseperez:
Fixed in #1249

PR #1249, opened 2018-10-06
Admin pass validation condition always failing. Adding hyphen to chars

djwmarcx:
Admin pass validation condition always failing. Adding hyphen to chars


xoseperez:
See #1244

PR #1295, opened 2018-10-21
All timeZones and DaylightSaving 

Diramu:
added fulling expanded setTime() function to take all parameters from a posix database

https://raw.githubusercontent.com/nayarsystems/posix_tz_db/master/zones.csv

to set all time zones include day light saving.

html/TZupdate is a perl script that reads this script, generating a selection tree in TS.js, and encoding of the timezone parameters in TZ.h included by ntp.ino

Requires my version of NtpClient: https://github.com/Diramu/NtpClient 
A PR also has been issue on this repro.


Diramu:
Thanks for links. Very useful.
If TZall[] is put in the ROM/flash, do need to make sure the elements are  a multiple of 4 bytes, or does the compiler take care  of this?

Aside: are the web pages (html) in the flash?

The advantage of the generated  TZall[], the posix notation has already been decoded ( and probably is smaller, and even much smaller because I removed duplicates). 

Still think (out loud)....

mcspr:
Embedis / Settings API is treating everything as `String`s so it must be somehow serialized. If not as TZ, maybe just plain values separated by some delimiter. Unless I am missing some other option to store it 🤔
(previous PR used multiple keys but single one should suffice) 

About WebUI:
[`PROGMEM`](https://github.com/esp8266/Arduino/blob/42c977bd4d14281e46ff5d01f8f860dd0c195a08/cores/esp8266/pgmspace.h#L24) is just a marker to store value in separate binary section, so it is not placed in RAM. Web is stored as byte array made from .gz file:
https://github.com/xoseperez/espurna/blob/88a7076ad3b906f8aeee0ae6978332f0e61b4335/code/espurna/static/index.small.html.gz.h#L1-L2
https://github.com/xoseperez/espurna/blob/88a7076ad3b906f8aeee0ae6978332f0e61b4335/code/espurna/web.ino#L200
"You must read from the flash 32 bit aligned.", hence the use of ..._P functions instead of directly using 
 it as char pointer.

mcspr:
*Little sidenote*
I had found an analogue to NtpClient, including timezone and timestamp fractions support:
https://github.com/ropg/ezTime

Instead of having prebuilt listing it can fetch tz data on-demand. By default, from author's own server:
https://github.com/ropg/ezTime/blob/760309591781f5c0b55439f11a96982c0e182228/src/ezTime.h#L143
https://github.com/ropg/ezTime/blob/3327f816ca89c68f13c8919f960de3c269bc2ace/src/ezTime.cpp#L772-L829

Diramu:
exTime looks good, but I still want to complete this exercise. Great learning.
This last update, addresses many  of the problem. 
I changed the selection technique. A single <select> statement, with a <optgroup> for each major zone. All <options> are hidden, until a <optgroup> is clicked, then the <option>s of this group are show.

The TZall table is now in EEPROM.

mcspr:
Some things to note:
- because you are hijacking `ntpOffset`, NTPClient::begin(ntpServer, ntpOffset) is called with "wrong" offset when it is using large indices (like 65) and it does not finish initializing. sync functions are never set and ntp request does not happen. simple solution is just not use ntpOffset for 'begin()' and rely on setting it later.
- there is a problem with strings like Asia/Tehran `<+0330>-3:30<+0430>,J80/0,J264/0`
it does not set explicit month/week/day but uses days as counter for dst start/end.
TZ.h does interpret it as `{"<+0330>",210, "<+0430>", 270, 0, 3, 0, 360, 30, 0, 0, 120}` which seems wrong.

BTW ezTime does implement posix string parser. Although, it does not support J... dst format.
https://github.com/ropg/ezTime/blob/3c0c86dd45f685355dce46ebfc554cd40abb2d0a/src/ezTime.cpp#L568-L766
e.g. NoNetwork example from that repo, with slightly modified string to trigger DST now (`<+0330>-3:30<+0430>,J80/0,J320/0`):
```
Local time   :  Sunday, 11-Nov-2018 02:25:03 MSK
Pacific time :  Saturday, 10-Nov-2018 15:25:03 PST
Tehran time  :  Sunday, 11-Nov-2018 02:55:03 <+0330>
UTC          :  Saturday, 10-Nov-2018 23:25:03 UTC
```

Diramu:
I've removed the timeZone offset to NTP.begin() and let it default.

Good catch on the 'J' notation.  I never saw it.  It seem to mean the day in the year.  At leat that match my search for Tehran dst dates.

Created a new TZ.h (from TZupdate.pl) with Tehran using day offsets to chose the start/end dates.
Modified the NtpClient to take the new method.  Just overrode the mean of the day of week. If 7, then the sum of month & week, is the day count.  Bit bit ugly, but...

Diramu:
BTW. The /usr/share/zoneinfo entry for Iran is wrong.  It should be:
<+0330>-3:30<+0430>,80/0,264/0

The 'J' means don't count 29 Feb, and start count from 1.  Without 'J', do count 29 Feb, and start count from 0.

Because it specified with a 'J',  we are out by a day on non leap years.

From https://en.wikipedia.org/wiki/Iran_Standard_Time

> The Iranian DST starts on March 22 and ends on September 22 each year with the exception of leapyears in which the DST starts and ends one day prior.

Diramu:
I did post an issue to https://github.com/nayarsystems/posix_tz_db/issues/3#issuecomment-440762339  about the 'j', but it's in the source  /usr/share/zoneinfo/

It appears stringrule(), is omitting the J when writing dates for Jan & Feb, because it not effected by a leap year and it saves a character, but it does modify the count by -1, to match the definition.

I'll see about posting an issue to IANA..

In the meantime, I've think I've taking this far enough.  The only issue is Tehran, the only non 'M' posix tz  in the tz db.  I'm sure an Tehran users must be dealing with this issue everywhere if its wrong in the source.

I'll look a changing to ezTime at some time, but I have a few other project to complete first.

mcspr:
> I'm sure an Tehran users must be dealing with this issue everywhere if its wrong in the source.

Not really though. zoneinfo files also contain transition times and this only happens when directly using POSIX tz variable. So, on modern system using zoneinfo this would not happen (...or until 2037 where it becomes approximation, but rules should be extended before that) 
For example, using gnu coreutils `date`:
```sh
# Using zoneinfo file
$ export TZ='Asia/Tehran'
$ date --date '21 Sep 2019 19:29:00 UTC'
Sat Sep 21 23:59:00 +0430 2019
$ date --date '21 Sep 2019 20:29:00 UTC'
Sat Sep 21 23:59:00 +0330 2019

# POSIX string
$ export TZ='<+0330>-3:30<+0430>,J80/0,J264/0'
$ date --date '21 Sep 2019 19:29:00 UTC'
Sat Sep 21 22:59:00 +0330 2019
$ date --date '21 Sep 2019 20:29:00 UTC'
Sat Sep 21 23:59:00 +0330 2019

$ date --date '20 Sep 2019 20:29:00 UTC'
Fri Sep 20 23:59:00 +0330 2019
$ date --date '20 Sep 2019 19:29:00 UTC'
Fri Sep 20 23:59:00 +0430 2019
```

mcspr:
😴 
Due to various reasons this got sidelined. Core support had added the exact same thing in the meantime, so there is no need to maintain this kind of thing here / library. Closing as current `dev` supports POSIX TZ. List of TZs is also maintained in the Core.

PR #1301, opened 2018-10-24
Add support for EZO™ pH Circuit

ruimarinho:
Add basic reading support for the [pH Circuit from Atlas Scientific](http://atlas-scientific.com/product_pages/circuits/ezo_ph.html) in UART mode.

screen shot 2018-11-26 at 01 39 20

Assumes device is using UART with in the default continuous mode (one reading per second). You can check in which reading mode the device is by sending the command `C,?`.

For a more responsive sensor read, enter the sensors tab on the Web UI and change "Reading interval" to 1 and "Report every" to 1 as well.


code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,142 @@
...
+
+            while ((_serial->available() > 0)) {
+                char ph_char = (char)_serial->read();
+                ph_serial += ph_char;
+
+                if (ph_char == '\r') {
+                  break;
+                }
+
+                yield();
```
ruimarinho: Is this considered best practice?

code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,142 @@
...
+
+        // ---------------------------------------------------------------------
+        // Protected
+        // ---------------------------------------------------------------------
+
+        void _read() {
+            String ph_serial = "";
+            ph_serial.reserve(30);
+
+            while ((_serial->available() > 0)) {
```
ruimarinho: The device is in continuous read mode by default. 1 reading per second. Is this the right way to read as fast as possible?

mcspr: #referencing https://github.com/plerup/espsoftwareserial/blob/master/SoftwareSerial.cpp
If I understand this right, actual reading is done by isr handler and you just advancing the buffer with read()

code/espurna/sensors/EZOPHSensor.h
mcspr: Maybe not. Real serial [`.available()`](https://github.com/esp8266/Arduino/blob/42c977bd4d14281e46ff5d01f8f860dd0c195a08/cores/esp8266/HardwareSerial.cpp#L90-L97) already yields, SW one see link above.

Note: [`optimistic_yield()`](https://github.com/esp8266/Arduino/blob/42c977bd4d14281e46ff5d01f8f860dd0c195a08/cores/esp8266/core_esp8266_main.cpp#L109-L115) checks if loop() `now() - loop_start_timestamp` is bigger than given usec interval

code/espurna/sensors/EZOPHSensor.h
ruimarinho: Ah, indeed. I need to look a bit more closely but there is a chance that a larger content of the buffer is returned in a single `read()` right?

code/espurna/sensors/EZOPHSensor.h
ruimarinho: Exactly, and apparently the actual rxRead method already has `WAIT` which is a micro for 
`optimistic_yield()`.

Thanks for the comments, it has been a great learning exercise for me.

code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,142 @@
...
+            return MAGNITUDE_NONE;
+        }
+
+        void pre() {
+            _read();
+        }
+
+        // Current value for slot # index
+        double value(unsigned char index) {
+            if (index == 0) return _ph;
```
ruimarinho: @mcspr how many times is `_pre` called? It looked like it was only once during sensor initialization but the value on the UI is being updated. Maybe the `_read()` function should also be called inside `value()`?

code/espurna/sensors/EZOPHSensor.h
mcspr: Tnx to google for reminder - 9600 baud =~ 960 chars per second.
Datasheet does not really mention any other things but short ~5-9char data chunks once in a ~second being `va.lue\r` + `*OK\r`
BTW it should also check response code (which is on by default?)

ruimarinho: In default continuous mode (UART), the output data are always the readings. There is no `*OK\r`, as that is only in response to commands, which this code does not send.

As far as `*ER`, `*OV` and the other error response codes, I'm thinking of testing the first chard and if it starts with `*`, I will simply set `_error = SENSOR_ERROR_OTHER`.

WDTY?

code/espurna/sensors/EZOPHSensor.h
ruimarinho: Ok, removed.

code/espurna/sensors/EZOPHSensor.h
ruimarinho: I will suggest passing `-DSENSOR_READ_INTERVAL=1` then.

ruimarinho: @mcspr even with `-DSENSOR_READ_INTERVAL=1 -DSENSOR_REPORT_EVERY=1`, `pre()` is only being called every 7-8 seconds. Any idea?

ruimarinho: Both defines are not modifiable after all. No one ever had to change these 🤔?

mcspr: Maybe `snsRead` is set from settings? Define is only the default value for it

ruimarinho: That's right. Is there anything else missing from this PR?

code/espurna/sensors/EZOPHSensor.h
mcspr: One thing I had noticed in similar [CSE7766 sensor](https://github.com/xoseperez/espurna/blob/88a7076ad3b906f8aeee0ae6978332f0e61b4335/code/espurna/sensors/CSE7766Sensor.h#L291-L336) is that it does the check the buffer every available opportunity using `.tick() -> ._read()`, so it waits only `loop()` time + `delay("loopDelay" / LOOP_DELAY_TIME / 10ms default)` instead. And also discards data on known intervals, counting since the last reading.

ruimarinho:
@xoseperez can you take a look at the revised PR?

mcspr:
@ruimarinho in the meantime ...did you forget to push?

ruimarinho:
@mcspr Yep, indeed. Pushed.

xoseperez:
LGTM. At least from the code point of view. Can you confirm it is working OK @ruimarinho ?

ruimarinho:
Sure. I will see if I can do that change tonight @xoseperez.

ruimarinho:
@xoseperez I have rebased the code but I think `tick()` is called too quickly. It also skips ESPurna sensor settings, which I'm not sure is a good idea. Any disadvantage with keeping pre?

mcspr:
Main advantage is separate sensor reporting from actual reading, which you know how fast (or slow) it is beforehand.
Given you want the report every 10 seconds - there are 10 reads in the serial buffer that are already 10 seconds older that you want.

ruimarinho:
@mcspr got it. I am comparing a manual sync time to a timestamp so I can skip readings if the sync interval is less than that.

However, I would like to issue a command `C,?` and parse it so that the sync time would come directly from the configured sensor. I just have no idea on how to do that as my attempts at issuing a `_serial->write("C,?"); _serial->write(13)` on `void begin()` have been fruitless. Any tips? Seems like my TX connection is being ignored.

mcspr:
Mm I think the problem is sending char pointer does not really work. "C,?" should be sent byte-by-byte like the '\r'
*(and i'm not really sure how it is interpreted char*->uint8_t. is it only first char sent?)*

ruimarinho:
Tried with the following as well — not sure where to send it though (during `begin()`, during `pre()`, ...).

```
_serial->write(67); // C
 _serial->write(44); // ,
_serial->write(63); // ?
_serial->write(13); // \r
```

Also tried adding `_serial->flush()` at the end.

Right now RX is connected to GPIO13 (D7, wemos) and TX is connected to GPIO15 (D8, wemos), but sometimes it just freezes the wemos. Sometimes it won't boot just by connecting the TX cable.

Pretty sure I am doing something very wrong :-)

mcspr:
Booting issue is easily explained, actually. [GPIO15 is read at startup and would switch the board into the SDIO mode instead of normal boot sequence](https://github.com/esp8266/Arduino/blob/d7ca8840070a629d40c6b6d0bacd11d4e652286e/doc/boards.rst#boot-messages-and-modes)

idk, did you try switching pins? D1 (5) and D2 (4) for example.
and one definitive way to check if it is actually sending the right data - just use external usb-uart and some terminal program with binary view (or wire to a second esp sw-gpio X hw-uart)

ruimarinho:
@mcspr great stuff! I am now able to receive the response to the command sent! I have move the write call to `begin()` but I think it's too soon as I get an `*ER` back. Should I just try moving to `pre()` or `tick()`?

ruimarinho:
@mcspr, @xoseperez thanks for all the guidance so far. I had never experimented with SoftwareSerial. The updated code now supports detecting the continuous mode interval and schedule accordingly. Since the default sensor comes configured in this mode, it will be easier for newcomers to just plug and play. In the future, if continuous mode is disabled, I can add support for a manual read using `pre()`'s schedule. 

I'm pretty sure there are more efficient ways to do what I just wrote (i.e. buffering more chars to get a full response with value + `*OK`). Open to suggestions and improvements.

code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,210 @@
...
+              char sync_char = (char)_serial->read();
+              sync_serial += sync_char;
+
+              if (sync_char == '\r') {
+                break;
+              }
+          }
+
+          if (sync_serial.startsWith("?C,")) {
+              _sync_interval = sync_serial.substring(sync_serial.indexOf(",") + 1).toInt() * 1000;
```
mcspr: indexof("?C,")?

code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,210 @@
...
+
+          _error = SENSOR_ERROR_OK;
+        }
+
+        void _read() {
+            if (_error != SENSOR_ERROR_OK) {
+              return;
+            }
+
+            if (millis() - ts <= _sync_interval) {
```
mcspr: This is harmless right now, but isn't it comparing `millis() <= _sync_interval` before first read? _setup should set it initially.

code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,210 @@
...
+        void _setup() {
+          if (_sync_responded == true) {
+            return;
+          }
+
+          _error = SENSOR_ERROR_WARM_UP;
+
+          String sync_serial = "";
+          sync_serial.reserve(30);
+
```
mcspr: Just to be safe at this point it can also .flush() the serial to drop possibly read data.

ruimarinho: Done, thanks.

code/espurna/sensors/EZOPHSensor.h
ruimarinho: It shouldn't because I have `if (_error != SENSOR_ERROR_OK) {` just before that.

code/espurna/sensors/EZOPHSensor.h
ruimarinho: I'm already making sure the string starts with `?C,` and that is the data format they use so I think it's good as it is. Alternative would be `indexOf("?C,") + 3` which wouldn't look as good.

code/espurna/sensors/EZOPHSensor.h
```diff
@@ -0,0 +1,210 @@
...
+            if (ph_serial == "*ER") {
+              _error = SENSOR_ERROR_OTHER;
+              return;
+            }
+
+            _ph = ph_serial.toFloat();
+            _error = SENSOR_ERROR_OK;
+        }
+
+        boolean _sync_requested = false;
```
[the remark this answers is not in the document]
ruimarinho: Fixed, thanks!

ruimarinho:
I was trying to leave some sensor debug messages such as:

```
          #if SENSOR_DEBUG
              DEBUG_MSG_P(PSTR("[SENSOR] Continuous reading mode interval is set to every %d seconds\n"), _sync_interval / 1000);
          #endif
```

...but I am always getting this:

```
In file included from espurna/config/all.h:33:0,
from code/espurna/espurna.ino:22:
.platformio/packages/framework-arduinoespressif8266@1.20300.1/cores/esp8266/pgmspace.h:21:51: error: __c causes a section type conflict with __c
#define PSTR(s) (__extension__({static const char __c[] PROGMEM = (s); &__c[0];}))
^
espurna/config/debug.h:11:42: note: in definition of macro 'DEBUG_MSG_P'
#define DEBUG_MSG_P(...) debugSend_P(__VA_ARGS__)
^
espurna/config/../sensors/EZOPHSensor.h:211:29: note: in expansion of macro 'PSTR'
DEBUG_MSG_P(PSTR("[SENSOR] pH reading: %f\n"), _ph);
^
.platformio/packages/framework-arduinoespressif8266@1.20300.1/cores/esp8266/pgmspace.h:21:51: note: '__c' was declared here
#define PSTR(s) (__extension__({static const char __c[] PROGMEM = (s); &__c[0];}))
^
espurna/config/debug.h:11:42: note: in definition of macro 'DEBUG_MSG_P'
#define DEBUG_MSG_P(...) debugSend_P(__VA_ARGS__)
^
code/espurna/ws.ino:391:21: note: in expansion of macro 'PSTR'
DEBUG_MSG_P(PSTR("[WEBSOCKET] #%u pong(%u): %s\n"), client->id(), len, len ? (char*) data : "");
^
*** [.pioenvs/wemos-d1mini/src/espurna.ino.cpp.o] Error 1
```

Any idea what am I doing wrong?

mcspr:
it is a esp8266 arduino core bug.
https://github.com/xoseperez/espurna/pull/1370/commits/103d8836df5320cf97b5fec0a814b4afbe89f98d has a port of recent changes that fix this. DEBUG_MSG_P / PSTR() / F() etc. should work in classes

ruimarinho:
I think this is ready for being merged @xoseperez.

PR #1307, opened 2018-10-31
typo error

salukadev:


xoseperez:
Is that only a space you are changing?

PR #1314, opened 2018-11-03
transparent serial-TCP bridge and flash-programming of attached AVR uControllers

tve:
This PR needs some discussion. I wrote a serial-TCP bridge and programmer for AVR processors a few years ago (https://github.com/jeelabs/esp-link) which uses the Espressif SDK directly and a crazy Makefile, etc. I am now porting the key functionality to Arduino and decided to use Espurna as the "framework" to plug the features into. I may add the flashing of AVR Megas and really would like to add the function of the Black-Magic-Probe (https://github.com/blacksphere/blackmagic/wiki) as well, which provides flashing and GDB access to ARM processors using jtag or single-wire-debug (SWD).

Now I realize that all this is pretty different from the current relays, sensors, actuators supported by Espurna, so my question is whether you are interested in adding this type of functionality or whether I should simply fork Espurna and go my own route. I suspect that the latter makes the most sense, but I didn't want to just go and fork silently without at least asking first.

In terms of pull request specifics, the bulk of the code is in separate libraries (https://github.com/jeelabs/esp-link-v4) and the PR needs some clean-up for sure. But I figured I'd wait for an overall answer before diving into splitting up and cleaning up the PR.

Note that the PR is against the v2 branch, not the dev branch. I hope v2 isn't abandoned...

image


swilson86:
I have been actively following Thorsten with esp-link/el-client and Xose on the WiFi products both for about 2-1/2 years.  To imagine the two of you coordinating things is incredible. I use both of your software products extensively.

I have a house that integrates both of your ideas, along with the node-red team and others now famous in these circles (Jcee, Hardill, knolleary, the Bald Engineer, etc). It would seem as if there are about 12 of us in the world highly involved in the same area of expertise. Forgive me if I left out your name for those of you in 3D printing, electronic designers, raspberry pi gurus, et al. I am not an expert in any singular area.

What I have developed over the last 2-1/2 years is a system that opens/closes the windows in my house when it rains. I can monitor garden moisture, human and animal movements around the house, Alexa integration, lighting, wind analysis (for closing windows partially when it gets windy, etc).

My point here is that I am thrilled when I see coordination between what i call "The Prime Developers" as I just mentioned.  Esp-link for example lacks in WiFi scanning and choosing the best signal where Xose's code does just that, while at the same time Hardills Alexa integration in node-red works somewhat well with Xose's Espurna and perfectly with Esp-Link.

May all of you meet each other and coordinate! I would love to attend a meeting of you all, whether online or a physical place. Barcelona 2019??

JohnOH:
+1 Nicely described @swilson86.

tve:
@swilson86: thanks for the enthusiasm! I'm getting the feeling that either Xose has moved on or is taking a break. Looks like he's involved in AllWize, perhaps all his time is going there. So my prediction is that sadly espurna is going to stagnate...

xoseperez:
Hi, this is a great PR. I've also been following you since a while ago. 

I understand your concerns about whether this functionality fits inside ESPurna, which is more oriented to so called "smart devices" and not development boards. But I can think of a use case where this would make sense. There are a number of connected devices, mostly sensors, that use an ESP8266 plus an ATMEGA328 (or alike) combination. The ESP8266 falls short in GPIOs, especially analog GPIOs, so teaming it with a "traditional" arduino microcontroller makes sense. I can think of the Sonoff SC or the 4Duino-24. Being able to flash the atmega from the ESP8266 makes a lot of sense in these cases. Of course, this should be an opt-in functionality. 

Allwize and The Things Network are stealing most of my time lately, that's true, but I will try to keep on with ESPurna. And yes, even thou there are a number of uncertainties in the v2 branch yet, that's the way to go.

tve:
Cool, I'll clean things up.

swilson86:
Are there any downloadable binary's yet on the esp-link/espurna project? I am really anxious to test it out!

PR #1322, opened 2018-11-07
Added ability to disable/enable buttons by MQTT.

vtochq:
Added ability to disable/enable buttons by MQTT.

For example it useful when automation enable light on door opening and you trying to turn on light manually. And you touch switch exactly after light turned on by automation and you mistakenly turn off lights.

### DIFF
--- a/code/espurna/button.ino
+++ b/code/espurna/button.ino
@@ -32,6 +32,56 @@ void buttonMQTT(unsigned char id, uint8_t event) {
     mqttSend(MQTT_TOPIC_BUTTON, id, payload, false, false); // 1st bool = force, 2nd = retain
 }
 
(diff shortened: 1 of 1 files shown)


vtochq:
Does nobody needs feature for temporary disable buttons?

mcspr:
Wouldn't  .../button/0/set be useful to have to "push" the button?
Not really sure how it may be better namespaced though.
.root./button/0/{action,relay,mode}/set , .root./config/button/0/{action,relay,mode}/set ?

There is floodwindow thing implemented for relays (5 switches max over 3 seconds window). Would it make sense to configure that instead? If automation switches relay on, it can block things based on relative time since the last switch just setting switch limit to 1. See RELAY_FLOOD_*. This may be exposed as a setting.

lancelon:
Is this feature coming?

PR #1324, opened 2018-11-07
fix _ir_results checks for Raw mode

vtochq:
Move _ir_results checks to non raw part.
Add _ir_results.rawlen check to raw part.


xoseperez:
Thanks!

PR #1326, opened 2018-11-08
Improvement on ANALOG_SENSOR

cconde:
Using defines allow now to scale the raw value and select the number of decimals.

It's useful when running on battery you want to check it by connecting the appropriate resistor from battery (+) to analog pin of the ESP8266


code/espurna/sensors/DHTSensor.h
```diff
@@ -204,7 +204,7 @@ class DHTSensor : public BaseSensor {
                 _humidity = dhtData[0];
             } else if (_type == DHT_CHIP_DHT12) {
                 _humidity = dhtData[0];
-				_humidity += dhtData[1] * 0.1;
+				        _humidity += dhtData[1] * 0.1;
```
Valcob: You may want to revert this file the changes in here are useless

xoseperez: Yes, please, revert changes to the DHTSensor.h file.

code/espurna/sensors/AnalogSensor.h
```diff
@@ -94,8 +115,28 @@ class AnalogSensor : public BaseSensor {
...
+          //Raw measure could also be a class variable with getter so that can
+          //be reported through MQTT, ...
+          unsigned int rawValue;
+          double scaledValue;
+          // Debugging doubles to string
+          //debugSend(("[ANALOG_SENSOR] Started standard read, factor: %s , offset: %s, decimals: %d \n"), String(_factor).c_str(), String(_offset).c_str(), ANALOG_DECIMALS);
+          rawValue = _rawRead();
+          //debugSend(("[ANALOG_SENSOR] Raw read received: %d \n"), rawValue);
+          scaledValue = _factor*rawValue  + _offset;
+          debugSend(("[ANALOG_SENSOR] Scaled value result: %s \n"), String(scaledValue).c_str());
```
xoseperez: Use DEBUG_MSG macro instead of debugSend.

cconde:
OK, I'll do that ASAP (I'm drowning on work right now).

Sorry, it's my first ever pull request, so forgive me all noob issues  :)

cconde:
Thank you very much, also for the advice.

Indeed it seems that my Atom wasn't indenting properly so it will be
reverted.

Also I don't have a clue on how the web internals so I can't expose the
three parameters for web configuration.
Having them in eeprom would be really nice as calibration tends to be done
on the final device, not at firmware level.

BTW I come from industrial automation so I have some good ideas on how to
detach measures from units in a clean way. I'll look for some spare time to
elaborate.
That should be done at the "Issues" part, isn't it? I'm pretty new to
Github, too...

Wonderful job for the whole Espurna project!

Valcob:
@cconde It's very simple to add a config item to WebUI all you need to do is to:
1. Add some code to void _sensorConfigure()
2. Add a bit to void _sensorWebSocketStart(JsonObject& root)
3. Add  some intput fields with the name set to be the same as the one used to persist these values in EEPROM   (check lines around <div class="pure-g module module-hlw module-cse module-emon"> for eg)   The module-[yournamehere] is the way to show or hide a block so when you pass from _sensorWebSocketStart()  a [yournamehereVisible] = 1 this will trigger the block to be visible 

just check the code around those lines it is pretty self explanatory and don't hesitate to ask if something is not clear.

@xoseperez I was thinking maybe we can move the config per each sensor from sensors.ino into their own modules ? and just have a generic js function that will add fields on the fly just to remove all the input fields from html  all we need we mostly have though we still would need to know to which module we need to add a specific setting request

cconde:
I'll give a try to these steps!

Is there any straightforward way to tell Github to use your file instead of
mine (revert) or shall I mess with my git?

Sorry, I've coded a lot but never within a VCS.


El 9 nov. 2018 17:14, "Vc" <notifications@github.com> escribió:

@cconde <https://github.com/cconde> It's very simple to add a config item
to WebUI all you need to do is to:

   1. Add some code to void _sensorConfigure()
   2. Add a bit to void _sensorWebSocketStart(JsonObject& root)
   3. Add some intput fields with the name set to be the same as the one
   used to persist these values in EEPROM (check lines around
   for eg) The module-[yournamehere] is the way to show or hide a block so
   when you pass from _sensorWebSocketStart() a [yournamehereVisible] = 1 this
   will trigger the block to be visible

just check the code around those lines it is pretty self explanatory and
don't hesitate to ask if something is not clear.

@xoseperez <https://github.com/xoseperez> I was thinking maybe we can move
the config per each sensor from sensors.ino into their own modules ? and
just have a generic js function that will add fields on the fly just to
remove all the input fields from html all we need we mostly have though we
still would need to know to which module we need to add a specific setting
request

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/xoseperez/espurna/pull/1326#issuecomment-437409679>,
or mute
the thread

code/espurna/sensors/DHTSensor.h
cconde: Reverted!

code/espurna/sensors/AnalogSensor.h
cconde: Should be correct now

mcspr:
@Valcob Please explain it in detail as separate issue.

I see DHTSensor is reverted, so treating @xoseperez review as resolved.

mcspr:
Oops. My bad not noticing `master` as target.
Later merge should figure it out anyways, but github should add a warning or something...

PR #1332, opened 2018-11-11
Add support for Lombex LX-BW09 smart bulb

kcghost:
MY9291 based smart lightbulb with 4 channels (RGB and warm leds). I have some photos and perhaps slightly dodgy information on how to flash it which I can add to the wiki if this gets merged in?

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -2875,6 +2875,28 @@
     #define LIGHT_CH4_INVERSE   0
 
 // -----------------------------------------------------------------------------
+// Lombex LX_BW09
+// https://www.amazon.com/Compatible-Automation-Dimmable-Equivalent-Changing/dp/B07587CNFV
+// -----------------------------------------------------------------------------
+#elif defined(LOMBEX_BW09)
+
+    // Info
+    #define MANUFACTURER        "LOMBEX"
+    #define DEVICE              "LX_BW09"
+    #define RELAY_PROVIDER      RELAY_PROVIDER_LIGHT
+    #define LIGHT_PROVIDER      LIGHT_PROVIDER_MY92XX
+    #define DUMMY_RELAY_COUNT   1
+
+    // Light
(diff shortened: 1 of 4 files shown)


kcghost:
Actually not sure if I have the MY92XX_MAPPING value correct? The interface seems to assume the fourth channel is cold LEDs, is the mapping value related to that? Not sure how that would translate to migrate.ino either...

xoseperez:
Yes, the MY92XX_MAPPING value is a map between the 4 MY9291 channels and the expected order in ESPurna, which is RGBWW. So "0,1,2,3" means the first channel in the MY9291 controls red, second green, and so on. Don't worry about the migrate.ino. I will do it when syncing the v2 branch.

code/espurna/config/hardware.h
```diff
@@ -2874,6 +2874,28 @@
     #define LIGHT_CH3_INVERSE   0
     #define LIGHT_CH4_INVERSE   0
 
+// -----------------------------------------------------------------------------
+// Lombex LX_BW09
+// https://www.amazon.com/Compatible-Automation-Dimmable-Equivalent-Changing/dp/B07587CNFV
+// -----------------------------------------------------------------------------
+#elif defined(LOMBEX_BW09)
```
xoseperez: The current version requires the board name to be MANUFACTURER_DEVICE, so this should be "LOMBEX_LX_BW09". Change it also in the platformio.ino file.

kcghost: K, I'll keep that in mind. I might need to cancel this request altogether and submit a new change later on due to two developments:

1. I blew up the bulb (Started making a repeated cap squeak noise and GNC and VCC are now shorted). It's possible it isn't designed to be run on both the RGB channels and warm LEDS full blast for very long, I think I had them on overnight. Have you experienced any problems like this?

2. I just went to experiment with a different newer variant of these Lombex bulbs "Lombex Lux Nova" that just have warm and cold LEDS, and found that the model number is frustratingly exactly the same.

PR #1335, opened 2018-11-12
fix for commit 30285b3

eschava:
fix for: we should send only buttons events that have actions assigned to them or set BUTTON_MQTT_SEND_ALL_EVENTS to 1 to send everything


mcspr:
👍 Thanks!

PR #1337, opened 2018-11-13
allow override of HA settings and WEB_REMOTE_DOMAIN

peterhoeg:
Overrides in ```custom.h``` do not otherwise work.


mcspr:
✨ Thank you!

PR #1370, opened 2018-11-24
Stubborn HASS MQTT Discovery

### LINKED ISSUE #1346: rfbridge only sending first 8 relays via MQTT for auto-discovery
While setting ```#define DUMMY_RELAY_COUNT 16``` in ```custom.h``` makes additional 8 relays available in the UI, only the first 8 are submitted to home assistant for auto-discovery.

Further to #1133.

Additional reference: #1308 

Cc: @mcspr @vogelfreiheit

mcspr:
`AsyncMqttClient::publish` method returns packet id > 0 when it is actually accepted by the network layer. worked for me with both QoS 0 (PID 1 is success) and QoS 1.
limited testing with lwip2/536mss+git, lwip2/1460mss+git and lwip1.4+2.3.0
pubsubclient works the same way, returning bool right from the .publish method.

Modified mqttSend... variants to check PID validity and check what they return in the discovery sender. Both ha relays and sensors methods modified into a helper class to store last sent internal array ID.

Additionally, this approach should also be applied to the config generator.

Depends on #1374 *// F() could be removed for it to work without it*
Fixes #1346
cc @peterhoeg

### DIFF
--- a/code/espurna/mqtt.ino
+++ b/code/espurna/mqtt.ino
@@ -496,25 +496,29 @@ String mqttTopic(const char * magnitude, unsigned int index, bool is_set) {
 
 // -----------------------------------------------------------------------------
 
-void mqttSendRaw(const char * topic, const char * message, bool retain) {
+bool mqttSendRaw(const char * topic, const char * message, bool retain) {
+    bool res = false;
 
     if (_mqtt.connected()) {
         #if MQTT_USE_ASYNC
             unsigned int packetId = _mqtt.publish(topic, _mqtt_qos, retain, message);
             DEBUG_MSG_P(PSTR("[MQTT] Sending %s => %s (PID %d)\n"), topic, message, packetId);
+            res = (packetId > 0);
         #else
-            _mqtt.publish(topic, message, retain);
+            res = _mqtt.publish(topic, message, retain);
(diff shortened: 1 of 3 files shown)


peterhoeg:
Should this be tested against dev or can I cherry-pick the commits on top of the latest release?

mcspr:
Idk if it is worth managing cherry picks, but it should work for master branch too.
homeassistant.ino had no changes since. There is one eeprom related method change in mqtt module, but it could be ignored and old one used instead.

code/espurna/mqtt.ino
```diff
@@ -34,7 +34,9 @@ WiFiClientSecure _mqtt_client_secure;
 #endif // MQTT_USE_ASYNC
 
 bool _mqtt_enabled = MQTT_ENABLED;
+bool _mqtt_reset = false;
```
mcspr: Oops. These are leftovers from the other branch.

abmantis:
Any reason why this is still pending?

mcspr:
IMO there should be generic one-shot callback function in the core to avoid juggling discovery helper
Otherwise, I remember this being pretty robust. Maybe one thing that should be tested is DEBUG_SUPPORT=0, because of implicit yields

PR #1374, opened 2018-11-27
Port PROGMEM from 2.5.0

mcspr:
Compilation no longer fails when using F() / DEBUG_MSG_P in some specific cases like sensor classes:
```cpp
/home/builder/.platformio/packages/framework-arduinoespressif8266@1.20300.1/cores/esp8266/pgmspace.h:21:51: error: __c causes a section type conflict with __c
#define PSTR(s) (__extension__({static const char __c[] PROGMEM = (s); &__c[0];}))
^
```
#1370 as an example of where this did happen in class method

### DIFF
--- a/code/espurna/config/prototypes.h
+++ b/code/espurna/config/prototypes.h
@@ -8,6 +8,14 @@ extern "C" {
     #include "user_interface.h"
 }
 
+// ref: esp8266/Arduino/cores/esp8266/pgmspace.h @ 2.5.0
+// __STRINGIZE && __STRINGIZE_NX && PROGMEM definitions port
+#define __TO_STR_(A) #A
+#define __TO_STR(A) __TO_STR_(A)
+
+#undef PROGMEM
+#define PROGMEM      __attribute__((section( "\".irom.text." __FILE__ "." __TO_STR(__LINE__) "."  __TO_STR(__COUNTER__) "\"")))
+
 // -----------------------------------------------------------------------------
 // API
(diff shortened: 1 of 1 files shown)


Valcob:
I don't think that is a good idea I overcame that by simply declaring all the string in progmem.h and then just reused them across the firmware there is no need to make constants per each file by wrapping in a modified F() with a counter when you can have them all in one place and just reuse across

mcspr:
Uniqueness is not a problem for debug messages though, which is what this solves:
https://travis-ci.org/mcspr/espurna/jobs/460520688 changes DEBUG_MSG -> DEBUG_MSG_P when SENSOR_DEBUG is enabled and builds ok
https://travis-ci.org/mcspr/espurna/jobs/460524020#L546 same as above, but without the unique section attr

Valcob:
@mcspr for the first link the warnings are telling you that comparing a  const == a & b is not safe and you should use const == (a & b) if that was what caught you attention 
for the second it's a legit thing when you are using F() in classes and that is solved the way I've described above

mcspr:
Added comments + defguards.

Have been wondering about PSTR comment from the link above. Testing size, can be removed if it is noticeable.

ruimarinho:
@mcspr is there any alternative for debugging inside sensor classes while this isn't merged?

mcspr:
DEBUG_MSG should work ok (as all SENSOR_DEBUG parts are using it)
Does this patch work ok though?

PR #1385, opened 2018-12-01
Add warning about TELNET_PASSWORD

mcspr:
#1382 had changed the define to TELNET_AUTHENTICATION
Seamlessly replace old -> new and show compile time warning

### DIFF
--- a/code/espurna/config/all.h
+++ b/code/espurna/config/all.h
@@ -28,6 +28,7 @@
 #include "arduino.h"
 #include "hardware.h"
 #include "defaults.h"
+#include "deprecated.h"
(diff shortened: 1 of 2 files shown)


xoseperez:
Not a bad idea, but why did you change the name of the file. I like `deprecated.h`, it clearly states what goes there...

mcspr:
I changed my mind after adding redefinition
+ migrate.ino / migrate() functions namesake

xoseperez:
I guess all the fuss is to provide nice notifications to users using `custom.h` files, right?
I still think `deprecated.h` is more suitable.

Issue #1367, opened 2018-11-23
Reduce ws / settings lists size


mcspr:
Current settings use `keyN` layout for lists:
https://github.com/xoseperez/espurna/blob/d2d92687c1b205b5ce02d4d9024b91c872fbc219/code/espurna/settings.ino#L335-L337
Same situation for websockets api. For example, relays, where `key` part is duplicated N times relays:
https://github.com/xoseperez/espurna/blob/d2d92687c1b205b5ce02d4d9024b91c872fbc219/code/espurna/relay.ino#L598-L606

To be more memory friendly - what about mapping them as simple arrays?
e.g. settings - for 8 relays `relayBoot => "0,0,0,0,0,0,0,0"`, similar for websockets `relayConfig[0] = ...`.

This saves *a lot* not duplicating `key` for every entry. Less EEPROM used, much less data transferred for rfbridge-like boards, much less RAM used for JSON.

Valcob:
I would go for a struct as tasmota does, yes I understand that it's level 100 cryptic but the saved space can make it worth(cause all the keys are gone also we can use unions for relay states, pins, lights, their values and so on) this way the saved space is of an order of magnitude from what we are using now, but that would require a much in depth redesign of how configs are stored, loaded, updated

xoseperez:
Still, cryptic code or data is not a good idea. One thing is packing the info is sent via WS (with the benefit of less memory usage doing JSON stuff) as @mcspr suggests in the PR. and another one is changing the way the data is stored in settings... 

@mcspr, let's discuss this further. I like the PR so far but I would limit the scope to the WS communication here and open a different one for EEPROM storage if we decide to change that too.

mcspr:
ref #1387, will update soon and link back to this
+note: js already used this way of packing lists to send back, but server side only did it sometimes

Will add another issue tracking settings then
(and regarding cryptic settings: one can always program embedis to serialize the data under the hood and have input/output be whatever)

Issue #1300, opened 2018-10-23
Slow websockets with Windows client


mcspr:
Thinking about https://github.com/xoseperez/espurna/commit/9f817ae2d0cc66bcbc0c683e9d51d28b19ee9c7a / #896 
`optimistic_yield()` did not do what I assumed it did - it just yields if loop spent more time that time value it is given (in usec).

Doing some testing with new ESPAsyncWebServer I got hit with queueing problem. As of me-no-dev/ESPAsyncWebServer@bed4146 queue limit for ws messages is 8, so messages dropped after some point.

<details>
<summary>Config difference from baseline</summary>

pio platform **feature/stage** (see #1260 for ldscripts generator)
**esp8266/Arduino@5493552** / **2.4.2-85-ge5493552** / **lwip2** high bandwidth
https://github.com/me-no-dev/ESPAsyncTCP#7e9ed22
https://github.com/me-no-dev/ESPAsyncWebServer#bed4146
```c
#define DEBUG_ESP_WIFI 1 // debug.ino setup calls .setDebugOutput(true);
#define WIFI_SLEEP_MODE WIFI_LIGHT_SLEEP // no difference NONE / MODEM
#define ALEXA_SUPPORT 0         // to avoid network activity besides asyncwebserver
#define NTP_SUPPORT 0           // ^
#define SCHEDULER_SUPPORT 0     // ^
#define HOMEASSISTANT_SUPPORT 0 // ^
#define MQTT_SUPPORT 0          // ^
```
</details>
<details>
<summary>
For example, using `keys`. `AsyncWebSocket.cpp` sections with os_printf / ets_printf are commented out.
Console output is filled with queue errors after some point and simple yield does not really help here as message objects are not yet purged.
</summary>

```
[159104] [WEBSOCKET] Requested action: dbgcmd
M: 51
W: 0 51
---> client->send() 51 bytes
S: 51 51
[159113] Current settings:
M: 32
[159115] > adminPass => "fibonacci"
M: 43
[159118] > alexaEnabled => "0"
M: 38
[159120] > board => "2"
M: 31
[159122] > boardName => "NODEMCU_LOLIN"
M: 47
[159124] > btnGPIO0 => "0"
M: 34
[159126] > btnRelay0 => "0"
M: 35
[159128] > cfg => "3"
M: 29
[159130] > dns0 => "10.0.0.1"
M: 39
ERROR: Too many messages queued
[159133] > gw0 => "10.0.0.1"
M: 40
[...]

PR #1387, opened 2018-12-01
Compact WS data

mcspr:
For #1367 

Do not duplicate keys for relay, sensor, schedule, rfbridge, thingspeak & domoticz ws objects
Convert bool -> int (3 and 4 char shorter!) for status arrays
Minor edit _magnitudes.size() -> magnitudeCount()
Reuse single JsonObject for all callbacks


mcspr:
I was distracted while doing the first merge.

Quick testing 4d5c0d7617ad0719efca72978ef9e76f1ded3f46 with Windows / Chrome and 16 switch rfbridge. Compared to this, it takes full 1-2 seconds for all of the rfb messages to come through, as it has to wait for ack from the client side to send the next one. This is strictly Windows problem, as any other client takes only 2ms per message and not 50ms, but I wanted to keep the same amount of messages sent to keep ws from needlessly buffering things in this case.
(ref: #1300 for debug messages, where each one was just sitting in the AsyncWebSocket internal buffer long after the wsSend was done. noticed only when new espasyncwebserver cut the buffering after 8 sends)

mcspr:
Will force-push https://github.com/xoseperez/espurna/compare/dev...mcspr:ws/optimize here to avoid doing merge + reduced buffer growth when using ws

Tangentially, this also lead up to trying to fix ws logging for the latest ESPAsyncWebServer by manually queueing messages. I wouldn't call it finished, as that approach only matters for ws. Telnet blows up with large payloads when using buffer and probably needs queuing like asyncwebsocket does internally. And it does not handle special case for '\r' char yet (OTA updates progress will fill the buffer with garbage)

mcspr:
I would argue that logging changes are related, but they can be split into a separate pr with ESPAsyncTCP / ESPAsyncWebServer versions update.

mcspr:
@xoseperez Generally, if you do not like approach at packing everything inside single ws message, there are two solutions for the ack problem at hand:
- force webui to keep state of which cfg it loaded, sending something like `{action: "load_cfg", {module: "relay"}` per each module and receive response from the esp. that way, only single module data is in the memory at once and queue is not kept filled. each time client responds with "action", ack is piggybacked inside, so the last message is removed from the queue whenever app yields into the system
- fork espasyncwebserver 🤷‍♂️ and introduce additional user callback that is called when tcp ACK is received. this will also help the solution from #1535 by making printer decisively send the data

PR #1390, opened 2018-12-04
Fix WS action lookup

Niek:
This fixes the "dbgcmd" messages in the websocket debug console.

### DIFF
--- a/code/espurna/ws.ino
+++ b/code/espurna/ws.ino
@@ -139,7 +139,7 @@ void _wsParse(AsyncWebSocketClient *client, uint8_t * payload, size_t length) {
 
     // Check actions -----------------------------------------------------------
 
-    const char* action = root["action"];
+    const char* action = root["data"]["command"];
     if (action) {
 
         DEBUG_MSG_P(PSTR("[WEBSOCKET] Requested action: %s\n"), action);


Niek:
I just noticed that this break the relay button in the UI. Maybe the code should look like this?
```c
if (strcmp(root["action"], "dbgcmd") == 0) {
  const char* action = root["data"]["command"];
} else {
  const char* action = root["action"];
}
```

code/espurna/ws.ino
```diff
@@ -139,7 +139,7 @@ void _wsParse(AsyncWebSocketClient *client, uint8_t * payload, size_t length) {
 
     // Check actions -----------------------------------------------------------
 
-    const char* action = root["action"];
+    const char* action = root["data"]["command"];
```
Valcob: As far as I can see we are still sending 
```
function sendAction(action, data) {
    websock.send(JSON.stringify({action: action, data: data}));
}
```
actions: .... in our custom.js so what this line actually fixes ?

Niek:
No matter what you type in the debug console, you get: ``[WEBSOCKET] Requested action: dbgcmd``, which looks incorrect to me (shouldn't it echo the command)? But maybe it need to be fixed at https://github.com/xoseperez/espurna/blob/de7ff7722c2c0eccd34ae0d7b08005fe38720b66/code/espurna/ws.ino#L145 as well as some code like https://github.com/xoseperez/espurna/blob/de7ff7722c2c0eccd34ae0d7b08005fe38720b66/code/espurna/rf.ino#L111-L113 which is now broken.

Valcob:
Actually dbgcmd is handled in debug.ino check the sutup function. And regarding the DEBUG_MSG_P logic well it only displays the action not the payload. The same is true for rfblearn   check the rf.ino check _rfWebSocketOnAction()

Niek:
@Valcob OK gotcha, I think it would be useful to display the payload as well in the debug console.

Re rf.ino: I thought the code was broken because the only way to issue the rfblearn/rfbforget/rfbsend commands is through the debug console. I see now that adding ``#define WEBUI_IMAGE WEBUI_IMAGE_RFBRIDGE`` adds the RF interface which executes the actions.

PR #1395, opened 2018-12-05
Add debouncing to pulse meter sensor

jackwilsdon:
I found debouncing to be required when using a "digital LDR" such as this one;

I copied the implementation from the event sensor, as it seemed pretty straightforward.

I don't think a configuration option is required for this, so it's hardcoded to 50ms right now (even then, if your power meter is flashing every 50ms then there's something wrong!).


xoseperez:
Thnak you!

PR #1400, opened 2018-12-07
add rssi to InfluxDB heartbeat

BuildTheRobots:
Add an option for logging RSSI (wifi signal level) as part of the heartbeat messages when using InfluxDB.

Ideally I'd make this a runtime option with a toggle on the InfluxDB web page, but I haven't worked out how to do that yet.

grafana-rssi-espurna


BuildTheRobots:
Thanks Max, that'd be great.

I did notice that enabling InfluxDB has lead to a little less stability in
the Sonoff Pow devices I've been testing with. Disabling some other
services seemed to help, so my blind assumption is that it's memory
related. I guess anything to make it more efficient would help.

Greatest of thanks :)

PR #1403, opened 2018-12-08
Update gulp@4

### LINKED ISSUE #1325: Upgrade to Gulp 4
Gulp 3.9 is used now, but is has some dependency vulnerabilities and it doesn't work with Node 10 (https://github.com/gulpjs/gulp/issues/2162). Upgrading to Gulp 4 would require some changes in the gulpfile.

mcspr:
Resolves #1325 

My only question is pinned "map-stream": "0.0.7" from 7b63521c80a37e854fc8941abb0aa21a48f60b86


xoseperez:
It is an old dependency. `map-stream'' was used to create the .h build pipe but was replaced with `through2` later on.

PR #1406, opened 2018-12-09
Runtime heartbeat configuration

mcspr:
As referenced in #1400 comment

`hbReport` bitset, mapped from `Heartbeat::Report` enum (1 << 17 right now in size)
Set for both MQTT and InfluxDB (as they are the only targets now)
Keep old config active + allow to redefine them at build time
LOADAVG_REPORT -> HEARTBEAT_REPORT_LOADAVG

*Maybe* WIP, as there is only one setting for now.
No mode setting (but can disable using 0 for value)
No UI config / doc etc.

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -185,23 +185,74 @@
 
 #define UPTIME_OVERFLOW             4294967295  // Uptime overflow value
 
-// Topics that will be reported in heartbeat
+// Values that will be reported in heartbeat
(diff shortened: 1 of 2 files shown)


xoseperez:
It looks good. Even thou there is no UI yet it's already more flexible than what we have now. Do you want to merge it?

mcspr:
Basically, yes.

As for web, this needs to be somehow reasonably compact to fit 17 labels + checkboxes

Issue #622, opened 2018-03-04
OBI Powerplug-Adapter (Sonoff S20 type of device)


Geitde:
The DIY store OBI in germany is selling a compatible power plug switch for 10 Euro, which is nice as you don´t have to wait for devices from china and as they are selling them in local shops it is guaranteed the quality has been somewhat checked. They can switch up to 3500W, which is at least more as the Dual/Basic/Touch switches can handle, which is nice.

Here is a link to arendst page.

https://github.com/arendst/Sonoff-Tasmota/wiki/Obi-Socket

Using the source below makes Espurna compatible with this power adapter. I bought 3 of them and they work fine incl LED, switch and relay. Thanks to K.H for working it out.

The ready to flash binary using 1.12.3 espurna can be found here: http://www.geit.de/tmp/espurna-1.12.3-obi.bin

Would be nice if the adapter could be added to the regular build system.

There is a little issue with this source. Alexa (WeMo/Belkin International inc. servers) are not detecting them by default. It is strange and seems to be related to some of the id fields. If you flash the normal S20 binary first, search for devices using alexa and flashing the right OBI firmware afterwards it works fine, so it has to be the device string or something else to break the finding process using this source.

Maybe you know what is required to allow proper WeMo detection. Here the source snipped to make it compatible.

Thanks for your hard work on this project.

#if defined(OBI_WIFISTECKDOSE)

// Info
#define MANUFACTURER "EUROMATE"
#define DEVICE "OBI_WIFI_STECKDOSE"

// Buttons
#define BUTTON1_PIN 14
#define BUTTON1_MODE BUTTON_PUSHBUTTON | BUTTON_SET_PULLUP | BUTTON_DEFAULT_HIGH
#define BUTTON1_RELAY 1

// Relays
// #define RELAY1_PIN 12
// #define RELAY1_TYPE RELAY_TYPE_LATCHED_INVERSE
// #define RELAY1_RESET_PIN 5
// hack: drive gpio12 low and use gpio5 as normal relay pin:
#define RELAY1_PIN 5
#define RELAY1_TYPE RELAY_TYPE_NORMAL
#define LED2_PIN 12 /* DUMMY: exploit default off state for GPIO12=low */
#define LED2_PIN_INVERSE 0

// LEDs
#define LED1_PIN 4
#define LED1_PIN_INVERSE 0
#elif defined(NODEMCU_LOLIN)

xoseperez:
[...]

PR #1408, opened 2018-12-09
Added configuration for the new WIFI device sold by Obi

arthurf1969:
There is a new version of the WIFI Schuko device sold by OBI (here in germany). This device uses an ESP-WROOM02 with two LED's (attached to GPIO12 and GPIO13) as well as a button attached to GPIO5.
I have also added an image and updated the readme file for this device.


code/espurna/config/hardware.h
```diff
@@ -3335,6 +3335,35 @@
     #define NETBIOS_SUPPORT     1
     #define SSDP_SUPPORT        1
 
+// -----------------------------------------------------------------------------
+// This configuration is for the second generation of devices sold by OBI.
+// https://www.obi.de/hausfunksteuerung/wifi-stecker-schuko-weiss/p/4077806
+// -----------------------------------------------------------------------------
+#elif defined(OBI_OBI_V2)
```
mcspr: Any reason why not name it the same as previous one?
`OBI_WIFI_STECKER_SCHUKO_V2`?

arthurf1969: No, I was just not aware of the brand behind the device. You're absolutely right this should be renamed (and probalby be relocated so that both are close together? What do you think?)

arthurf1969:
I have changed the name of the configuration to EUROMATE_WIFI_STECKER_SCHUKO_V2, relocated the configuration next to the older device EUROMATE_WIFI_STECKER_SCHUKO and fixed the spelling of the word "Schuko" also in the configuration of the older device.

mcspr:
Thank you!

@xoseperez 
For the v1, #622 is the only mention of the EUROMATE I could find. Maybe just have OBI for both? https://github.com/arendst/Sonoff-Tasmota/pull/3944 links to the same store page and actually tracked down the manufacturer (which is not the same for this one): http://www.cn-yidong.com/Home/En-products-cid-18.html

arthurf1969:
Just as an information for you: I originally did not take a closer look at the housing of the device. On the back of the device there is the name EUROMATE (injection molded in the housing - white on white ... silly is). Euromate is part of the OBI brand. Thus my original thought on OBI. But to me (similar to the version 1) this is just a rebranded device from some other chinese brand. In contrast to the first version it does not use a latching relay.

mcspr:
Well, then it makes sense to keep it like it is. 
Thank you for clarifying, as I was not aware of obi / euromate relation (slightly different google search would've revealed it:)

Issue #1265, opened 2018-10-10
SETTINGS_AUTOSAVE and new delayed eeprom commit


mcspr:
When autosave is disabled it removes the commit call from the saveSettings():
https://github.com/xoseperez/espurna/blob/88a7076ad3b906f8aeee0ae6978332f0e61b4335/code/espurna/settings.ino#L367-L371

But it still will eventually save it anyways. For example, with relaySave (when switching relay)
https://github.com/xoseperez/espurna/blob/88a7076ad3b906f8aeee0ae6978332f0e61b4335/code/espurna/relay.ino#L415-L422
Or relay boot event (setSetting would already set "hostname" and "board")
https://github.com/xoseperez/espurna/blob/88a7076ad3b906f8aeee0ae6978332f0e61b4335/code/espurna/relay.ino#L549-L551

Should it be more restrictive setting, like `EEPROM_AUTOCOMMIT`? Keeping the EEPROM buffer but do not save unless explicitly told to.
Because it will be possible to set such setting, would it make sense to keep some state in RTC? Like relay state between reboots. It should keep things around unless device completely powers off or gone into deep sleep

xoseperez:
AFAIK, RTC memory does not survive across reboots: "The stored data can be retained between deep sleep cycles. However, the data might be lost after power cycling the chip."

The main reason not to commit the data right away is to save writing cycles. Using a force commit to write the rest of the buffer is OK.

mcspr:
Oh, I confused RTC timer with RTC mem itself. Citing NONOS_SDK pdf:
> System time will return to zero because of `system_restart` (note: === `Esp.restart()`), but RTC still goes on.
If an external hardware reset the chip via EXT_RST or CHIP_EN (such as timed
wakeup from Deep-sleep), the RTC timer will be reset.
• Reset by pin EXT_RST: RTC memory won’t change; RTC timer starts to zero.
• Watchdog reset: RTC memory won’t change; RTC timer won’t change.
• system_restart: RTC memory won’t change; RTC timer won’t change.
• Power-on: RTC memory contains a random value; RTC timer starts from zero.
• Reset by pin CHIP_EN: RTC memory contains a random value; RTC timer
starts from zero.

And about data what I meant is having EEPROM in Read-Only state forever. Essentially just treat that flash segment as some initial state to work with.

mcspr:
[...]

PR #1420, opened 2018-12-11
Initial implementation of RTCMEM storage

mcspr:
Part of #1265 

Store system crash counter and reset reason in rtcmem instead of eeprom
Store relay state mask in rtc in addition to the eeprom
Store relay state in eeprom only when boot mode requires it
Simplify relay state mask calculation / reading via std::bitset

Missing:
energy counter (`EEPROM_ENERGY_COUNT`) - should it be saved like this?
mqtt json id counter (`EEPROM_MESSAGE_ID`) - can write here every send, no need for delay like for eeprom.

*Note:*
This also can directly use rtcmem as an array (example [from esp-open-rtos](https://github.com/SuperHouse/esp-open-rtos/blob/a8c60e096093e9e9f4a60b885676adc2cf5b790a/core/include/esp/rtcmem_regs.h))
```c
typedef volatile uint32_t rtcmem_array_t[96];
#define RTCMEM_USER (*(rtcmem_array_t *)(RTCMEM_ADDR))
...
RTCMEM_USER[...SOME_INDEX_DEFINE...] = 1; // indexes defined like for eeprom
RTCMEM_USER[97] = 2; // compiler will yell because of rtcmem_array_t definition
```

And idk if consistency check is required at all, and just checking magic value would suffice
\+ tracking `resetInfo->reason` for `REASON_DEFAULT_RST` / `REASON_SOFT_RESTART`


xoseperez:
We are not checking for consistency in the flash atm.
I would add energy count too and light colors also (same update frequency than relays).

mcspr:
for energy counter the goal here is to protect readings in between save intervals from disappearing?

lights are straightforward enough, save 5 channels + brightness + mireds state. even fits inside 64 bit. only unlike relay it always saves - should it have an option like `relayBoot` to optionally do it?

mcspr:
Removed crc, added mqtt counter, added general methods for energy total and lights handling.
Last thing - check when it is appropriate to save them.

For lights: LIGHT_SAVE_ENABLED should be a setting and somehow wired into this?
For sensor: one can also save uptime() in rtc and reference that in energy ts code, since now it becomes 2 states - uptime under power with occasional reboots and complete shutdown with eeprom saves.

code/espurna/light.ino
```diff
@@ -469,6 +469,39 @@ void _lightProviderUpdate() {
...
+        uint8_t channels[5];
+        uint8_t brightness;
+        uint16_t mired;
+    } packed;
+    uint64_t value;
+};
+
+void _rtcmemLightSave() {
+    rtcmem_light_t light;
+    for (unsigned int i=0; i < lightChannels(); i++) {
```
xoseperez: We don't support any device with more than 5 channels (are there any?) but maybe we should check here the `lightChannels()` return value.

mcspr: Yeah, that is true.
Unlike pwm channels init this cannot really depend on LIGHT_CHANNELS to init array size. `rtcmem_light_t::channels` size is still hardcoded, but this can be `std::min(lightChannels(), sizeof(rtcmem_light_t::channels))` instead of simply using it.

mcspr:
Light issue resolved via static check of array size.

To reiterate - breaking changes are:
mqtt counter (no longer persisted between power-offs)
emon sensor total persists between reboots

I wonder if ntp time can be saved too.

mcspr:
No obvious things stand up. Merging in the interest of moving milestone forward.

Exception system might need some tweaking though:
- maybe lock the boards in safe mode until reboot. don't reset system stability counter to 0 on the 5th time, and force manual intervention (terminal command / web switch / power cycle)
- allow to store "small" crash log, instead of the full one

PR #1426, opened 2018-12-14
Fix compile error when TERMINAL_SUPPORT is disabled

Niek:
I noticed an issue in my PR of yesterday https://github.com/xoseperez/espurna/pull/1424: when ``TERMINAL_SUPPORT`` is disabled but ``OTA_MQTT_SUPPORT`` is enabled, there was a build error. This should fix that.

In addition, I added a small enhancement to the ``GPIO`` debug command, to get the state of all valid pins at once.


Niek:
Re-opened, up-to-date with upstream

code/espurna/settings.ino
```diff
@@ -213,7 +213,12 @@ void _settingsInitCommands() {
 
     settingsRegisterCommand(F("GPIO"), [](Embedis* e) {
         if (e->argc < 2) {
-            DEBUG_MSG_P(PSTR("-ERROR: Wrong arguments\n"));
+            DEBUG_MSG_P(PSTR("Printing all GPIO pins:\n"));
```
mcspr: This makes sense in general, but maybe save it for later to show used gpios instead?

Niek: Actually the PR was for 2 things at once ;) But yes I understand your concern, it's better to split them. I have updated the PR now.

PR #1454, opened 2018-12-27
Fix: Add Debug flag for compilation of wifiDebug() function

debarko:
Won't compile the Debug code in the final build which will help reduce the size of the build.

### DIFF
--- a/code/espurna/wifi.ino
+++ b/code/espurna/wifi.ino
@@ -467,10 +467,9 @@ void _wifiWebSocketOnAction(uint32_t client_id, const char * action, JsonObject&
 // INFO
 // -----------------------------------------------------------------------------
 
-#if DEBUG_SUPPORT || TERMINAL_SUPPORT
-
 void wifiDebug(WiFiMode_t modes) {
 
+    #if DEBUG_SUPPORT || TERMINAL_SUPPORT
     bool footer = false;
 
     if (((modes & WIFI_STA) > 0) && ((WiFi.getMode() & WIFI_STA) > 0)) {
@@ -511,14 +510,13 @@ void wifiDebug(WiFiMode_t modes) {
     if (footer) {
         DEBUG_MSG_P(PSTR("[WIFI] ----------------------------------------------\n"));
     }
+    #endif //DEBUG_SUPPORT
 
 }
 
 void wifiDebug() {
     wifiDebug(WIFI_AP_STA);
 }
-
-#endif //DEBUG_SUPPORT
 
 // -----------------------------------------------------------------------------
 // API
--- a/code/espurna/wifi.ino
+++ b/code/espurna/wifi.ino
@@ -467,6 +467,8 @@ void _wifiWebSocketOnAction(uint32_t client_id, const char * action, JsonObject&
 // INFO
 // -----------------------------------------------------------------------------
 
+#if DEBUG_SUPPORT
+
 void wifiDebug(WiFiMode_t modes) {
 
     bool footer = false;
@@ -515,6 +517,8 @@ void wifiDebug(WiFiMode_t modes) {
 void wifiDebug() {
     wifiDebug(WIFI_AP_STA);
 }
+
+#endif //DEBUG_SUPPORT
 
 // -----------------------------------------------------------------------------
 // API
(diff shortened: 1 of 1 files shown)


mcspr:
Can you give config when you encountered this?

I see different issue altogether with following attempt to disable both debugging and terminal input:
```c
#define DEBUG_WEB_SUPPORT 0
#define DEBUG_TELNET_SUPPORT 0
#define TERMINAL_SUPPORT 0
```
```
jhdtl@DESKTOP-I8S9V0N:/mnt/c/Users/Maxim/Projects/espurna/wifidebug/code$ pio run -s -e nodemcu-lolin ; ls -l .pioenvs/nodemcu-lolin/firmware.bin
.pioenvs/nodemcu-lolin/src/espurna.ino.cpp.o: In function `webSetup()':
/mnt/c/Users/Maxim/github/espurna/wifidebug/code/espurna/ws.ino:76: undefined reference to `wifiDebug()'
/mnt/c/Users/Maxim/github/espurna/wifidebug/code/espurna/ws.ino:76: undefined reference to `crashDump()'
.pioenvs/nodemcu-lolin/src/espurna.ino.cpp.o: In function `function<void (*)(AsyncWebServerRequest*), void>':
/mnt/c/Users/Maxim/github/espurna/wifidebug/code/espurna/ws.ino:76: undefined reference to `crashClear()'
.pioenvs/nodemcu-lolin/src/espurna.ino.cpp.o: In function `_Function_base':
/mnt/c/Users/Maxim/github/espurna/wifidebug/code/espurna/ws.ino:76: undefined reference to `wifiDebug()'
.pioenvs/nodemcu-lolin/src/espurna.ino.cpp.o: In function `_M_init_functor':
/mnt/c/Users/Maxim/github//espurna/wifidebug/code/espurna/ws.ino:76: undefined reference to `crashDump()'
.pioenvs/nodemcu-lolin/src/espurna.ino.cpp.o: In function `function<_telnetNewClient(AsyncClient*)::__lambda41, void>':
/mnt/c/Users/Maxim/github/espurna/wifidebug/code/espurna/ws.ino:76: undefined reference to `crashClear()'
collect2: error: ld returned 1 exit status
*** [.pioenvs/nodemcu-lolin/firmware.elf] Error 1
```

Given that DEBUG_ defines are for debug text (what is wifiDebug shows) and TERMINAL_ is for interactive input (not really dependent on output) you initial change is more in line with configuration logic. Maybe be best to make define check inside the functions, keeping the signature, so it is just empty when not needed.

debarko:
Making the define check inside the signature makes sense. I'll update the PR by EOD today.

debarko:
> Given that DEBUG_ defines are for debug text (what is wifiDebug shows) and TERMINAL_ is for interactive input (not really dependent on output) your initial change is more in line with configuration logic.

If I don't put that TERMINAL check, `settingsRegisterCommand(F("WIFI").` in the terminal will stop working. It is a standard information command which can be used for finding the details by any application connecting to this. So even though this function is called wifiDebug but it is called from an information context as well. So I made the second commit to retain support for that. Otherwise nothing will get printed if I give command WIFI in terminal.

One option is to hide the settingsRegisterCommand for WIFI when in non-debug mode if you guys are considering it as a debug only command.

I genuinely feel that it can have other purpose in life to handle a lot of cases like triggering things on connecting to a specific WiFi.

Let me know your thoughts and I'll update accordingly.

mcspr:
Which is why I was asking what exactly does not work / what config causes it. Or `WiFi.getMode()` returns unexpected thing for some reason.

If DEBUG_SUPPORT is empty, then every DEBUG_MSG_P / DEBUG_MSG will become empty. No debug strings compiled in => less space used.
When you just checked for DEBUG_SUPPORT and it compiled at all, `wifi` should've still called `wifiDebug(WIFI_AP_STA)` and produced output somewhere (DEBUG_SERIAL_SUPPORT=1 then it is serial, DEBUG_TELNET_SUPPORT=1 for telnet etc.)

debarko:
Updated.

mcspr:
Thanks!
*:oops: merged as-is by "refreshing" github page :/ i swear i clicked squash*

As for DEBUG_SUPPORT in general:
DEBUG_TELNET_SUPPORT comment in `general.h` describes what I encountered. Error does make sense, only slightly unexpected. Maybe dependency should just disable TELNET_SUPPORT in that case.
Preprocessor checks are a bit unwieldy to navigate, maybe more functions like this should be added (keep empty function when something is disabled)

Issue #291, opened 2017-11-26
Magic Home LED Controller Domoticz


xoseperez:
Originally reported by: **piotereczek (Bitbucket: [piotereczek](https://bitbucket.org/piotereczek), GitHub: [piotereczek](https://github.com/piotereczek))**

----------------------------------------

Not Controllable with Domoticz…
Not changing Color but only can switch on or off via RGS Switch.

tammycutzet:
Hi, 

integration of this Controllers, and also the AL-LC0X (Which are using the same protocol) would be awesome. There is no cheaper LED Controller available. Maybe its worth to stick on the flux_led python script.

sq5gvm:
You can check and build source on my fork of espurna (see https://github.com/sq5gvm/espurna/).
I've added (ugly) support for RGB(WW?) MQTT messages and I've been using it successfully for last two weeks on my AL-LC02 with RGB strip. I think communication from Domoticz to Espurna works OK. For now I haven't tried adding Espurna to Domoticz communication (if I find some time this week I'll look into it).

Only issue I have at this moment is that (at least on my unit) pins for RGB are defined incorrectly, I'll have three more units of AL-LC02 next week from different sources (and of different ages) and if all of those have the same pin connections I'll make changes to the definitions of espurna pinout for this boards and create pull request.

sq5gvm:
See referenced pull request.
Regarding one issue I had. I can confirm that all AL-LC02 I was able to get a hold of have different pin-to-color connections than one defined for ARILUX_AL_LC02 in hardware.h.
Checked using 14 devices bought from multiple sources at different dates (some mine, some borrowed). 
In all cases units had to be connected with "crossover" cable as such:
Controller | RGB Strip
-------------|---------------
GREEN | GREEN
RED | Not connected (white?)
BLUE | RED
WHITE | BLUE

I'll check pinout directly on board and add pin numbers for my LC02.

sq5gvm:
I was able to acquire one OLD AL-LC02 and it works without any problems. 
Other LC02 devices I have are newer PCB versions and have problems with displaying correct colours due to lack of support of those in current espurna version. 
I've sent pull request that adds support for AL-LC02 v1.4 so if you have one of those you might like to wait for it to be approved and hopefully merged.

Tankyspanky:
I bought a "Home-Magic RGBW controller from Amazon and i can configure it as a AL-LC02 RGBW switch,
[...]

PR #1459, opened 2018-12-28
Support for RGB(WW) controlled using Domoticz MQTT messages

sq5gvm:
Support for control of RGB(WW) LED strips from Domoticz over MQTT.
Tested using AL-LC02 for the last ~3 weeks.
All comments highly appreciated.

References #291


xoseperez:
Thank you very much!

PR #1469, opened 2019-01-04
Support for newer AL-LC02 boards with different pinout

sq5gvm:
Newer AL-LC02 boards (or at least devices sold as such) **have different LED pins used** than original LC02 boards. 
From outside the device looks almost exactly as AL-LC11 (including voltage rating on input side) with exception of LED strip connector (RGBW so only 5 pins and label with description of 5 pins). Unfortunately pinout of new board is different from existing LC11.
Above mentioned differences unfortunately are not bullet proof way to distinguish devices as I have few LC-02 units with "old" label and "new" boards.
So only sure way to distinguish device with new board from one with old board is to disassemble the device and see markings on the PCB which will read "ESP-IR-A V1.4".

This pull request contains changes that add new firmware for such devices (arilux-al-lc02v14) by adding new platform and new entries in hardware configuration.
Appropriate changes should also be made in Wiki mentioning existence of such boards and need to use alternative firmware is user experiences problems with mismatched colours (only problem I found).

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -1637,6 +1637,26 @@
     #define LIGHT_CH3_INVERSE   0
     #define LIGHT_CH4_INVERSE   0
 
+#elif defined(ARILUX_AL_LC02V14)
+
+    // Info
+    #define MANUFACTURER        "ARILUX"
+    #define DEVICE              "AL_LC02V14"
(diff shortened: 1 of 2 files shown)


sq5gvm:
Just to be clear I'm willing to send appropriate changes to wiki after support for device has been merged.

xoseperez:
Thanks, I changed a bit the board id to ARILUX_AL_LC02_V14. This way it is consistent with other devices.

PR #1474, opened 2019-01-06
Configure Heartbeat from WebUI & option HEARTBEAT_REPEAT_STATUS

martiera:
- Heartbeat mode and interval configuration added to WebUI ADMIN page
- New option HEARTBEAT_REPEAT_STATUS: on device startup full heartbeat information and after HEARTBEAT_INTERVAL only STATUS report.
- Enhanced first heartbeat message - on device startup it will send heartbeat after ntpSynced will be true to get device time.

P.S. Code is tested. Web part is not compiled,and not tested, but I hope it will work :)

### DIFF
--- a/code/espurna/system.ino
+++ b/code/espurna/system.ino
@@ -82,8 +82,6 @@ unsigned long systemLoadAverage() {
 }
 
 void _systemSetupHeartbeat() {
-    if (getSetting("hbMode").length() == 0)  setSetting("hbMode", HEARTBEAT_MODE);
-    if (getSetting("hbInterval").length() == 0)  setSetting("hbInterval", HEARTBEAT_INTERVAL);
     _heartbeat_mode = getSetting("hbMode", HEARTBEAT_MODE).toInt();
     _heartbeat_interval = getSetting("hbInterval", HEARTBEAT_INTERVAL).toInt();
 }
@@ -116,9 +114,9 @@ void systemLoop() {
     } else if (_heartbeat_mode == HEARTBEAT_REPEAT || _heartbeat_mode == HEARTBEAT_REPEAT_STATUS) {
         static unsigned long last_hbeat = 0;
         #if NTP_SUPPORT
-            if ((_system_send_heartbeat && ntpSynced()) || (millis() - last_hbeat > _heartbeat_interval)) {
+            if ((_system_send_heartbeat && ntpSynced()) || (millis() - last_hbeat > _heartbeat_interval * 1000)) {
         #else
-            if (_system_send_heartbeat || (millis() - last_hbeat > _heartbeat_interval)) {
+            if (_system_send_heartbeat || (millis() - last_hbeat > _heartbeat_interval * 1000)) {
         #endif
             last_hbeat = millis();
             heartbeat();
--- a/code/espurna/system.ino
+++ b/code/espurna/system.ino
@@ -12,6 +12,8 @@ Copyright (C) 2018 by Xose Pérez <xose dot perez at gmail dot com>
 
 unsigned long _loop_delay = 0;
 bool _system_send_heartbeat = false;
+unsigned char _heartbeat_mode = HEARTBEAT_MODE;
+unsigned long _heartbeat_interval = HEARTBEAT_INTERVAL;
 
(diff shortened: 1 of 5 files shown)


code/espurna/system.ino
```diff
@@ -97,19 +110,21 @@ void systemLoop() {
...
-    #elif HEARTBEAT_MODE == HEARTBEAT_REPEAT
+    if (_system_send_heartbeat && _heartbeat_mode == HEARTBEAT_ONCE) {
+        heartbeat();
+        _system_send_heartbeat = false;
+    } else if (_heartbeat_mode == HEARTBEAT_REPEAT || _heartbeat_mode == HEARTBEAT_REPEAT_STATUS) {
         static unsigned long last_hbeat = 0;
-        if (_system_send_heartbeat || (last_hbeat == 0) || (millis() - last_hbeat > HEARTBEAT_INTERVAL)) {
-            _system_send_heartbeat = false;
+        #if NTP_SUPPORT
+            if ((_system_send_heartbeat && ntpSynced()) || (millis() - last_hbeat > _heartbeat_interval)) {
```
xoseperez: Why checking for `ntpSynced` here?

code/html/index.html
```diff
@@ -620,6 +620,34 @@ <h2>Device administration and security settings</h2>
                                     <div class="pure-u-1 pure-u-lg-3-4 hint">Request password when starting telnet session</div>
                                 </div>
 
+                                <div class="pure-g">
+                                    <label class="pure-u-1 pure-u-lg-1-4">Heartbeat mode</label>
+                                    <select class="pure-u-1 pure-u-lg-1-4" name="hbMode" tabindex="15" >
+                                        <option value="0">HEARTBEAT_NONE</option>
```
xoseperez: Use the hints here as option text, the HEARTBEAT_NONE,... keys here mean nothing for the end user. And we save some space removing the hints below.

code/html/index.html
```diff
@@ -620,6 +620,34 @@ <h2>Device administration and security settings</h2>
...
+                                        <strong>HEARTBEAT_NONE:</strong> no heartbeat at all, never.<br />
+                                        <strong>HEARTBEAT_ONCE:</strong> only on device startup.<br />
+                                        <strong>HEARTBEAT_REPEAT:</strong> on device startup and after HEARTBEAT_INTERVAL.<br />
+                                        <strong>HEARTBEAT_REPEAT_STATUS:</strong> on device startup full heartbeat information and after HEARTBEAT_INTERVAL only STATUS report.
+                                    </div>
+                                </div>
+
+                                <div class="pure-g">
+                                    <label class="pure-u-1 pure-u-lg-1-4">Heartbeat interval</label>
+                                    <input name="hbInterval" class="pure-u-1 pure-u-lg-1-4" type="text" tabindex="16" />
```
xoseperez: This should be defined as a number, and probably in seconds, not millis.

code/espurna/system.ino
```diff
@@ -75,6 +81,13 @@ unsigned long systemLoadAverage() {
     return _load_average;
 }
 
+void _systemSetupHeartbeat() {
+    if (getSetting("hbMode").length() == 0)  setSetting("hbMode", HEARTBEAT_MODE);
```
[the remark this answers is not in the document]
martiera: Ok. That's I understood. But initially when first time device has been flashed we need to save those settings, otherwise always will be default value, not saved one. I'm wrong?

code/espurna/system.ino
martiera: I have added ntpSynced because if we have report interval ONCE, I have noticed that on first report sometimes datatime value is not published because of not synced ntp. I was thinking that it would be nice to get the datatime on device startup to check if everything is ok.

code/html/index.html
martiera: I agree. Will do a fix.

code/html/index.html
martiera: Yes, I was thinking about sec as well. Will do a fix. How would you like to store that setting in flash as sec or millis?

martiera:
Requested changes done.

xoseperez:
Merged!
Thank you

PR #1477, opened 2019-01-07
Build RF_SUPPORT in travis

mcspr:


mcspr:
@Niek it seems #1475 did not actually solve all errors with rf_support=1

Niek:
Fixed in https://github.com/xoseperez/espurna/pull/1479

PR #1478, opened 2019-01-07
Correct Teckin 1.4-1.6 LED logic

gmag11:
LED was signalled with inverse logic


xoseperez:
Not in my units. Notice the second LED (LED2) is set in LED_MODE_FINDME, that means it will be ON if the relay if OFF and vice-versa. The first one should blink on for a short brief every 5 seconds if connected to your home router.

gmag11:
Ok, I understand. Delete the PR then. I'll keep it in my fork as I like it to be red if relay is off and blue if on.

xoseperez:
You will be able to do it with v2, since you will be able to change the configuration from the terminal or the web interface.

PR #1489, opened 2019-01-14
ONLY processes Domoticz RGB MQTT Messages for the current idx

soif:
Fix for #1459

MQTT messages were processed regardless of the current idx.

Now it  only decodes messages with the right domoticz idx.


xoseperez:
Merged, thank you!

soif:
you're welcome !

PR #1526, opened 2019-02-05
Python cleanup

Cabalist:
Hey there,

I was just reading through the code and cleaned up the Python code.   Nothing functionally should change.   I've put some comments in the PR to explain the motivation behind some changes.  I feel they make the code more obvious

### DIFF
--- a/code/memanalyzer.py
+++ b/code/memanalyzer.py
@@ -19,15 +19,16 @@ import argparse
 import os
 import re
 import shlex
+import subprocess
 import sys
 from collections import OrderedDict
-from sortedcontainers import SortedDict
-import subprocess
 
-if (sys.version_info > (3, 0)):
-    from subprocess import getstatusoutput as getstatusoutput
+from sortedcontainers import SortedDict
+
+if sys.version_info > (3, 0):
+    from subprocess import getstatusoutput
 else:
-    from commands import getstatusoutput as getstatusoutput
+    from commands import getstatusoutput
 
 # -------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


code/memanalyzer.py
```diff
@@ -92,24 +93,26 @@ def analyse_memory(elf_file):
...
     return ret
 
 
 def run(env_, modules_):
     flags = ""
-    for item in modules_.items():
-        flags += "-D%s_SUPPORT=%d " % item
-    command = "ESPURNA_BOARD=\"WEMOS_D1_MINI_RELAYSHIELD\" ESPURNA_FLAGS=\"%s\" platformio run --silent --environment %s 2>/dev/null" % (flags, env_)
+    for k, v in modules_.items():
+        flags += "-D{}_SUPPORT={:d} ".format(k, v)
```
Cabalist: This was doing auto tuple unpacking.   I made it more explicit.    This took some time to decipher.  :)

code/memanalyzer.py
```diff
@@ -120,7 +123,8 @@ def modules_get():
     del modules_['NETBIOS']
     return modules_
 
-try:
+
```
Cabalist: `try: except: raise` is the same as the default behaviour.   The code will behave the same without it.   I removed it and replaced it with a `if __name__ == '__main__':` in the case any of the functions would end up getting reused.

code/ota.py
```diff
@@ -120,31 +121,35 @@ def get_boards():
...
     return board
 
+
 def get_board_by_index(index):
     """
     Returns the required data to flash a given board
     """
     board = {}
-    if 1 <= index and index <= len(devices):
+    if 1 <= index <= len(devices):
```
Cabalist: This is equivalent and more concise

Cabalist:
Looks like I am failing the Codacy review over existing code.   `subprocess` is necessary to functionality here.

code/ota.py
```diff
@@ -272,13 +283,13 @@ def run(device, env):
...
 
-    # Look for sevices
+    # Look for services
     zeroconf = Zeroconf()
     browser = ServiceBrowser(zeroconf, "_arduino._tcp.local.", handlers=[on_service_state_change])
     discover_last = time.time()
     while time.time() < discover_last + DISCOVER_TIMEOUT:
-        None
-    #zeroconf.close()
+        pass
```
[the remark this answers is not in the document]
Cabalist: I like the idea of breaking these into functions as well. 

I also like that that library has dropped Python2 support.   😉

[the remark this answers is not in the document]
xoseperez: Yes, but this whole PR looks like a good clean-up. I'm going to merge it.  @Cabalist maybe you'd like to test the pychromecast lib approach and submit a different PR with it?

PR #1533, opened 2019-02-07
Added Lyasi light bulb

Eichhoernchen:
This PR adds support for a Lyasi light bulb.
It is similar to the AiLight but uses different GPIO pin assignments.


xoseperez:
Thank you!

PR #1535, opened 2019-02-08
Print each HA config entry separately

### LINKED ISSUE #1233: Sonoff RF Bridge HASS config
The Sonff RF Bridge HASS configuration is not showing anything when I click 'Show' button. I can able to see on other esp8266 switches.
I made my own firmware with 16 switches enabled, compiled and uploaded. The RF switch configuration page for switch 10 to 16 contains 'Switch OFF' default value as 36EC01CC054610C10E. That was the 'Switch OFF' entry for my 2nd switch

mcspr:
Fixes #1233 
Does involve sort-of a hack by knowing that DEBUG_MSG will not print timestamp if the last symbol is not new-line/cr-return.

### DIFF
--- a/code/espurna/homeassistant.ino
+++ b/code/espurna/homeassistant.ino
@@ -9,9 +9,12 @@ Copyright (C) 2017-2018 by Xose Pérez <xose dot perez at gmail dot com>
 #if HOMEASSISTANT_SUPPORT
 
 #include <ArduinoJson.h>
+#include <queue>
 
 bool _haEnabled = false;
 bool _haSendFlag = false;
+
+std::queue<uint32_t> _ha_send_config;
 
 // -----------------------------------------------------------------------------
 // UTILS
@@ -147,9 +150,7 @@ void _haSendSwitches() {
 
 // -----------------------------------------------------------------------------
 
-String _haGetConfig() {
-
-    String output;
+void _haDumpConfig(std::function<void(String&)> printer) {
 
     #if (LIGHT_PROVIDER != LIGHT_PROVIDER_NONE) || (defined(ITEAD_SLAMPHER))
         String type = String("light");
@@ -163,8 +164,12 @@ String _haGetConfig() {
         JsonObject& config = jsonBuffer.createObject();
         _haSendSwitch(i, config);
 
-        output += "\n" + type + ":\n";
+        String output;
+        output.reserve(config.measureLength() + 32);
+
+        output += "\n\n" + type + ":\n";
         bool first = true;
+
         for (auto kv : config) {
             if (first) {
                 output += "  - ";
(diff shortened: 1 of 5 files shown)


code/espurna/homeassistant.ino
```diff
@@ -296,6 +313,21 @@ void _haInitCommands() {
...
 
+void _haLoop() {
+    if (_ha_send_config.empty()) return;
+
+    uint32_t client_id = _ha_send_config.front();
+    _ha_send_config.pop();
+
+    _haDumpConfig([client_id](String& output) {
+        output.replace(" ", "&nbsp;");
+        output.replace("\n", "<br />");
```
mcspr: Maybe this also can be done on the client.
And, after #1387 wsSend change, can even directly use JsonObject to create the data structure

code/espurna/homeassistant.ino
```diff
@@ -317,6 +349,7 @@ void haSetup() {
 
     // Main callbacks
     espurnaRegisterReload(_haConfigure);
+    espurnaRegisterLoop(_haLoop);
```
xoseperez: Is it really worth it to have it being checked in the main loop all the time? Wouldn't it be better t use a ticker?

code/html/custom.js
```diff
@@ -1366,7 +1367,9 @@ function processData(data) {
         // -----------------------------------------------------------------------------
 
         if ("haConfig" === key) {
-            $("#haConfig").show();
+            websock.send("{}");
```
xoseperez: Why this?

code/espurna/homeassistant.ino
mcspr: ...definitely must be in the client bc `Text(...)` can auto convert newlines to the proper html (like the logging does)

code/espurna/homeassistant.ino
mcspr: One option is to make additional API for such one-off functions to not overuse Ticker. Somehing accepting std::function with bound args.
Also see #1370

And I think I should also check if client is connected before starting the process.

code/html/custom.js
mcspr: Speeding up Windows <-> ESP tcp exchange. Quite noticeable.
https://github.com/xoseperez/espurna/pull/1387#issuecomment-446388717
https://github.com/xoseperez/espurna/issues/1300#issuecomment-432781707

mcspr:
...which reminds me, yield() in sending func should be delay(10...20ms) instead. while this speeds things up, it is nowhere near instant. 
will double check with default Windows install + asyncwebserver master that queue is used fast enough

code/html/custom.js
mcspr: NOTE: in case DEBUG_SUPPORT is enabled, client must send something back too in response to the "weblog". Otherwise, haConfig messages are stuck behind. Gosh darnit...

PR #1547, opened 2019-02-13
Add unique id and device for better HA UI integration

abmantis:
The unique id allows the device name and entity_id in HA to be edited from the UI. The "device" info allows the entities to be associated with a single device and appear on the MQTT Integrations UI.

### DIFF
--- a/code/espurna/homeassistant.ino
+++ b/code/espurna/homeassistant.ino
@@ -38,9 +38,11 @@ void _haSendMagnitude(unsigned char i, const JsonObject& deviceConfig, JsonObjec
     config.set("platform", "mqtt");
     config["state_topic"] = mqttTopic(magnitudeTopicIndex(i).c_str(), false);
     config["unit_of_measurement"] = magnitudeUnits(type);
-    config["uniq_id"] = getIdentifier() + "_" + type + "_" + String(i);
-    config["device"] = deviceConfig;
 
+    if (deviceConfig.size() > 0) {
+        config["uniq_id"] = getIdentifier() + "_" + type + "_" + String(i);
+        config["device"] = deviceConfig;
+    }
 }
 
 void _haSendMagnitudes(const JsonObject& deviceConfig) {
@@ -84,8 +86,10 @@ void _haSendSwitch(unsigned char i, const JsonObject& deviceConfig, JsonObject& 
     config.set("name", _haFixName(name));
     config.set("platform", "mqtt");
 
-    config["uniq_id"] = getIdentifier() + "_switch_" + String(i);
-    config["device"] = deviceConfig;
+    if (deviceConfig.size() > 0) {
+        config["uniq_id"] = getIdentifier() + "_switch_" + String(i);
+        config["device"] = deviceConfig;
+    }
 
     if (relayCount()) {
         config["state_topic"] = mqttTopic(MQTT_TOPIC_RELAY, i, false);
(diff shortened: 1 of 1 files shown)


mcspr:
Hey.

I don't think it should add anything into the configuration yaml generator, see
https://www.home-assistant.io/components/switch.mqtt/#device
https://www.home-assistant.io/components/sensor.mqtt/#device

abmantis:
@mcspr sorry, I don't understand what you mean.

mcspr:
@abmantis try running `ha.config` in terminal:
```
sensor:
  - name: ESPURNA_3601AF_event
    platform: mqtt
    state_topic: test/ESPURNA-3601AF/event/0
    unit_of_measurement:
    uniq_id: ESPURNA-3601AF_14_1
    device: {"identifiers":["ESPURNA-3601AF"],"name":"ESPURNA_3601AF","manufacturer":"Espurna","model":"NODEMCU_LOLIN","sw_version":"1.13.4-dev (2f147ec0)"}
```

abmantis:
Oh ok. So you think this should only be applied to the discovery message and not the generator. Is that it?

mcspr:
Yes, it does not apply to anything other than discovery.

code/espurna/homeassistant.ino
```diff
@@ -31,17 +31,21 @@ String _haFixName(String name) {
...
+void _haSendMagnitude(unsigned char i, const JsonObject& deviceConfig, JsonObject& config) {
 
     unsigned char type = magnitudeType(i);
     config["name"] = _haFixName(getSetting("hostname") + String(" ") + magnitudeTopic(type));
     config.set("platform", "mqtt");
     config["state_topic"] = mqttTopic(magnitudeTopicIndex(i).c_str(), false);
     config["unit_of_measurement"] = magnitudeUnits(type);
 
+    if (deviceConfig.size() > 0) {
+        config["uniq_id"] = getIdentifier() + "_" + type + "_" + String(i);
```
mcspr: Note that this uses internal type id. I think this should be `magnitudeTopic(type)`

code/espurna/homeassistant.ino
```diff
@@ -54,7 +58,7 @@ void _haSendMagnitudes() {
         if (_haEnabled) {
             DynamicJsonBuffer jsonBuffer;
             JsonObject& config = jsonBuffer.createObject();
-            _haSendMagnitude(i, config);
+            _haSendMagnitude(i, deviceConfig, config);
```
mcspr: maybe avoid checking anything inside _haSend... and just set things right here?
```js
config["device"] = deviceConfig;
config["unique_id"] = getIdentifier() + "_" + magnitudeTopic(magnitudeType(i)) + "_" + String(i);
```

code/espurna/homeassistant.ino
abmantis: the unique_id is an ID that is not supposed to be user friendly, so I think the magnitudeType char is ok for this.

mcspr: I mean, it is internal to the espurna, and might be changed to be something else. String is unlikely to change though.

abmantis: Oh ok. I thought that the string was more likely to change. I'll changed that then.

code/espurna/homeassistant.ino
```diff
@@ -135,6 +137,9 @@ void _haSendSwitches() {
             DynamicJsonBuffer jsonBuffer;
             JsonObject& config = jsonBuffer.createObject();
             _haSendSwitch(i, config);
+            config["uniq_id"] = getIdentifier() + "_switch_" + String(i);
```
mcspr: btw, why not use `type` here?

abmantis: no need. switch_**i** is unique already

abmantis: but I have no problem changing it

mcspr: huh, i think it only matters per platform. just tried changing topic from switch to light and it works ok.
i see that you already changed it though :) so let's keep it with `type`

abmantis:
Anything missing for this to be merged?

mcspr:
...note that it pretty much does not work when switches are >3, on initial connection.
manual `ha.send` works (default build, where debug messages do yielding)

'll instigate fix from #1370

PR #1551, opened 2019-02-14
added support for gosund ws1/KS-602S

nsvrana:
Added support for the low-cost Gosund WS1 that can be found on Amazon. Built using PlatformIO and tested on my device. Back of the device lists the model number as the KS-602S but the original firmware and board identify it as a WS1.

Upon acceptance of PR will upload pictures and steps to flash. Switch is connected to GPIO0 for easy flash mode configuration.

This is my first time working with an embedded platform like this, so feel free to correct me if I have done it wrong in any way. Thanks!


code/platformio.ini
```diff
@@ -2802,16 +2802,18 @@ upload_port = ${common.upload_port}
 upload_flags = ${common.upload_flags}
 extra_scripts = ${common.extra_scripts}
 
-[env:gosund-sp1-v23]
+[env:gosund-ws1]
```
mcspr: Please do not delete existing gosund-sp1-v23 env :)

nsvrana: Oops, fixed. Thanks!

mcspr:
I took the liberty of fixing platformio.ini specifics with upload_port / upload_flags when uploading via serial.

BTW I'd suggest moving hardware.h definition to avoid confusion with comment above + maybe add amazon / manufacturer link in the comment?

PR #1553, opened 2019-02-17
ADDED SUPPORT FOR OUKITEL P1 SMART SWITCH

quinnsam:
Added initial support for Oukitel Smart Switch. This is a 2 outlet plug
that has the same ESP8266 pinout as the YiDian XS-SSA05.


mcspr:
Thanks!
*...and thanks for noticing migrate.ino :)*

PR #1559, opened 2019-02-19
.ld scripts refactoring

### LINKED ISSUE #1260: Arduino Core from git and ldscripts
https://github.com/esp8266/Arduino/commit/38779149d07aecefc0e23d1452d38ee01d2437f8#diff-a57fd98a29a009fc8cfd4b56cd019b53
added **local.** prefix to ldscript, so this will no longer work:
https://github.com/xoseperez/espurna/blob/2de92f87206422f764e7689203d5c8214dd4b3f0/code/eagle.flash.1m0m2s.ld#L19

using pio 'feature/stage' and trying to build anything:
```
c:/users/maxim/.platformio/packages/toolchain-xtensa/bin/../lib/gcc/xtensa-lx106-elf/4.8.2/../../../../xtensa-lx106-elf/bin/ld.exe: cannot open linker script file ../ld/eagle.app.v6.common.ld: No such file or directory
collect2.exe: error: ld returned 1 exit status
```

mcspr:
Fixes #1260 

- distribute two sets of ldscripts
- automatically pick proper ldscript via platformio python extra script
- additional flash sizes for Generic board
- float support ON/OFF menu option for IDE
- ldscript generator (sector calculation included, based on boards.txt.py script from Core)

#### Arduino IDE (only tested with 2.5.0 release):
copy proper ldscripts into the `<esp8266-package-dir>/tools/sdk/ld`
copy boards.local.txt into the `<esp8266-package-dir>`
run / restart IDE, select `ESPurna board` in the menu

#### PlatformIO
same as above, but happens automatically
added `platform = ${common.arduino_core_git}` and `platform = ${common.arduino_core_2_5_0}` to allow testing with the newest Core

#### TODO
<s>boards.local.txt generator script</s> Done
<s>ldscript generator</s> Done
<s>maybe try to find out different way to detect core version (platform.txt? parse framework package version?)</s> Done (Checking 'local....' presence)


mcspr:
Also trying out draft PRs.
I do not see travis build, so here it is: https://travis-ci.org/mcspr/espurna/builds/495245870

mcspr:
When building from IDE, it generates special string that describes settings. Like this:
```
$ grep fqbn $TMP/arduino_build_446496/build.options.json
"fqbn": "esp8266com:esp8266:espurna:xtal=160,vt=flash,exceptions=disabled,ResetMethod=ck,CrystalFreq=26,FlashFreq=40,FlashMode=dout,ip=lm2f,float_support=disabled,variant=esp8285,eesz=1M,wipe=none,baud=115200"
```
</br>
If board definition has some defaults (not under `.menu.*`), it is possible to launch cli tools with fqbn just being `esp8266com:esp8266:espurna` to build "default" flavour.
https://github.com/arduino/arduino-builder, arguments for which can be copied from IDE build log
https://github.com/arduino/arduino-cli, which is merciful and only wants fqbn setting

Ofc it is not compatible between versions (search SUBSTITUTIONS in boards script)

mcspr:
But binutils DEFINED and preprocessor defined() are different things. That is a linker script level expression, checking if symbol exists. Example given uses ternary expression to assign symbol some default value. And even then, are there global level if/else expressions to use it?

But we can do something like this:
```
$ cat check.sh.in
#include <core_version.h>
#if defined(ARDUINO_ESP8266_RELEASE_2_3_0) || defined(ARDUINO_ESP8266_RELEASE_2_4_2)
echo old
#else
echo new
#endif
$ cc -x c -P -E -I~/.platformio/packages/framework-arduinoespressif8266@2.20402.2/cores/esp8266 check.sh.in -o check.sh
$ bash check.sh
old
```

[BTW, Arduino Core does use preprocessor to generate that file](https://github.com/esp8266/Arduino/blob/f950d53d821ed1f65eebdbefd632956a99a29562/tools/platformio-build.py#L214)

Niek:
OK, I haven't tested too much with the ``DEFINED()`` in binutils.

Another option would be to use regular C ``#define`` statements in the ld files and preprocess them with something like ``gcc -E -x c -C eagle.flash.1m0m1s.ld | grep -v "^# "`` - but the bash script probably makes more sense.

dist/boards_local_txt.py
```diff
@@ -0,0 +1,172 @@
...
+        return len(self._store)
+
+
+BOARDS_LOCAL = {
+    "global": collections.OrderedDict(
+        [("menu.float_support", "scanf and printf float support")]
+    ),
+    "flash_size": collections.OrderedDict(
+        [
+            (".menu.{eesz}.1M", "1M (1 EEPROM Sector, no SPIFFS)"),
```
mcspr: This collides with Core's 1M (no SPIFFS) option (2.4.0 and up, because 2.3.0 does not have it)
Names need to be unique.

sadly, no way to hide other ones. https://github.com/arduino/Arduino/issues/6034 is undocumented way to hide boards, but not menu options :(

dist/ld/README.txt
```diff
@@ -0,0 +1,15 @@
...
+
+## PlatformIO
+
+LIBPATH automatically picked based on platform.json:version
+
+## Arduino IDE
+
+### Pre 2.5.0
+
+cp pre_2.5.0/*.ld <esp8266-package>/tools/sdk/ld
```
mcspr: Since boards.local.txt no longer depends on core version, it can go here too

dist/ld/latest/eagle.flash.1m0m1s.ld
```diff
@@ -11,9 +12,9 @@ MEMORY
   irom0_0_seg :                         org = 0x40201010, len = 0xf9ff0
 }
 
-PROVIDE ( _SPIFFS_start = 0x402FB000 );
-PROVIDE ( _SPIFFS_end = 0x402FB000 );
+PROVIDE ( _SPIFFS_start = 0x402fb000 );
+PROVIDE ( _SPIFFS_end = 0x402fb000 );
```
Niek: Before you merge this, can you add the _FS_* aliases here as well so it also builds with core git?

mcspr:
```diff
-INCLUDE "../ld/eagle.app.v6.common.ld"
+PROVIDE ( _FS_start = _SPIFFS_start );
+PROVIDE ( _FS_end = _SPIFFS_end );
+PROVIDE ( _FS_page = _SPIFFS_page );
+PROVIDE ( _FS_block = _SPIFFS_block );
+
+INCLUDE "../ld/local.eagle.app.v6.common.ld"
```
?

Maybe Core needs a ESP::getFlashLayout() or something? I tried looking at other projects, ESPEasy for example defines global flag at runtime to switch between _SPIFFS_... and _FS.... Tasmota hides them under ARDUINO_RELEASE... flags. 2.6.0 is due in september, maybe we can figure something out before that.

mcspr: Also, see https://github.com/xoseperez/eeprom_rotate/issues/6 (cc @reaper7)
I will fix pio builds, but not IDE. I don't think IDE builder will search in the project for ldscripts, so it kind of annoying thing to manage already, both INCLUDE and this. And until IDF (partitions.csv) switch it probably will be :/

Niek:
That diff is exactly what I have now in my linker script, and it works fine (with PIO). Either that or utils.ino and eeprom need to be ifdef-ed with the correct macro name.
By the way is anybody seriously building Espurna with IDE? I tried it once but it takes an hour to gather all dependencies with the correct versions etc. PIO is just so much better IMHO.

mcspr:
Yeah, and an interesting note from the Tasmota issue tracker:
https://github.com/arendst/Sonoff-Tasmota/issues/6073#issuecomment-511109803

We do reference include inside tools/sdk/ld, so it does (?) work ok, or at least how it always worked before. But, we can't exactly not use them with 2.3.0 and rely on default scripts. We still need those aliases for some time, so I'll add this as a temporary solution (and to speed up -git builds setup time)

I'll try to come up with a reasoning for Core issue tracker and open an issue there regarding _FS_... symbols usage (or some other way to use them through helper methods)
Rotate lib can be forked or imported inside / be submodule / lib header
boards.local kind of solves this too, since our scripts have different names

Niek:
FYI the SPIFFS symbols have been added to core as well now: https://github.com/esp8266/Arduino/commit/418b00f7c05d8658e265ff3246b508998423a14c

PR #1588, opened 2019-02-27
Domoticz lights refactoring

mcspr:
- refactor domoticz lights handler into separate function
- less debugging strings. show that domoticz had parsed values in a single line. -480 bytes off the firmware. (travis03 dev:631248 domoticz/update:630768)
- fix relay getter to return signed value, as intended

cc @karmacoma92
please check if it fixes the problem you are experiencing (from gitter https://gitter.im/tinkerman-cat/espurna?at=5c7651cd8a7def0752e40812)

### DIFF
--- a/code/espurna/domoticz.ino
+++ b/code/espurna/domoticz.ino
@@ -17,7 +17,7 @@ std::vector<bool> _dcz_relay_state;
 // Private methods
 //------------------------------------------------------------------------------
 
-unsigned char _domoticzRelay(unsigned int idx) {
+int _domoticzRelay(unsigned int idx) {
     for (unsigned char relayID=0; relayID<relayCount(); relayID++) {
         if (domoticzIdx(relayID) == idx) {
             return relayID;
@@ -135,7 +135,7 @@ void _domoticzMqtt(unsigned int type, const char * topic, const char * payload) 
                 }
             #endif
 
-            unsigned char relayID = _domoticzRelay(idx);
+            int relayID = _domoticzRelay(idx);
             if (relayID >= 0) {
                 unsigned char value = root["nvalue"];
                 DEBUG_MSG_P(PSTR("[DOMOTICZ] Received value %u for IDX %u\n"), value, idx);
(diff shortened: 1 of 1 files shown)


karmacoma92:
Hi Max, 

Tested and working, something weird appeared after first reboot, in the general tab instead of appearing Temperature#0
Humidity#0
appeared something like NaN#0 repeated a number of times, ... 

it is working now Thanks a lot!
Oscar

mcspr:
@karmacoma92 thank you.
about the second one - no idea right now. [this function generates the string](https://github.com/xoseperez/espurna/blob/5c0abca719f4151c294c7277da4baf9d8629a9c4/code/html/custom.js#L59), but i cannot get it to output NaN with strings / numbers / nulls :/
if you'll have time, do you mind tracing websocket sensor data (`{magnitudes: type:[...], ...}` ) that is sent to the webui? if you are using chrome or safari, see `/ws` in the network tab of web developer tools (ctrl / cmd + shift + i)

karmacoma92:
Hi Max, 
It happened only once, the first time I finally upgraded it, I have tried to reproduce it upgrading another esp01s-dht11 but it didn't happened so I don't know if makes any sense to investigate further. 
Here you have a screenshot of the requested trace, (I do not see anything weird though,...)
espurna_details

I'm experiencing some other problem with a basic Sonoff which I cannot swtich on, it keeps on going back to off, but I do not know if I should open a different case....

Thanks once more
Oscar

mcspr:
ok, thanks anyways.
and btw, web inspector would only show the requests after they are made, meaning you need to have it active and then reload the page. it'll show /auth, hostname, and /ws, where /ws will have "frames" tab with websockets data. which may or may not have been broken, somehow.

> I'm experiencing some other problem with a basic Sonoff which I cannot swtich on, it keeps on going back to off, but I do not know if I should open a different case...
---
if that is reproducible, then yes.

karmacoma92:
I'll try reproducing it with the latest code from today but with the previous one, yes I can reproduce it....

karmacoma92:
I'll try reproducing it with the latest code from today but with the previous one, yes I can reproduce it....
Regarding the other issue, I was trying to reload with F5, but it was giving nothing as shown in the picture....

karmacoma92:
I tried with a compilation with the changes form yesterday, but still the same, please tell me where to report it... 
Thanks
Oscar

PR #1591, opened 2019-02-28
Add reactive power calculation and bring output to in line with HLW80…

AlbertWeterings:
CSE7766 Sensor based units

### DIFF
--- a/code/espurna/sensors/CSE7766Sensor.h
+++ b/code/espurna/sensors/CSE7766Sensor.h
@@ -280,7 +280,7 @@ class CSE7766Sensor : public BaseSensor {
 			unsigned int active = _active;
 			unsigned int apparent = _voltage * _current;
             if (apparent > active) {
-                _reactive = (apparent * apparent - active * active);
+                _reactive = sqrt(apparent * apparent - active * active);
             } else {
                 _reactive = 0;
             }
--- a/code/espurna/sensors/CSE7766Sensor.h
+++ b/code/espurna/sensors/CSE7766Sensor.h
@@ -161,9 +161,10 @@ class CSE7766Sensor : public BaseSensor {
             if (index == 0) return MAGNITUDE_CURRENT;
             if (index == 1) return MAGNITUDE_VOLTAGE;
             if (index == 2) return MAGNITUDE_POWER_ACTIVE;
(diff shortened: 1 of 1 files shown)


code/espurna/sensors/CSE7766Sensor.h
```diff
@@ -273,6 +275,16 @@ class CSE7766Sensor : public BaseSensor {
                 }
             }
 
+			// Calculate reactive power
+			_reactive = 0;
+			unsigned int active = _active;
+			unsigned int apparent = _voltage * _current;
+            if (apparent > active) {
+                _reactive = (apparent * apparent - active * active);
```
mcspr: _reactive = sqrt(...) ?

based on:
[hlw8012 lib code](https://github.com/xoseperez/hlw8012/blob/2dace562c339e77a2889446d461e8e56048fe634/src/HLW8012.cpp#L120)
[power triangle (c) google](https://en.wikipedia.org/wiki/AC_power#Active,_reactive,_and_apparent_power)

AlbertWeterings: See line 116-124 of HLW8012.cpp I used the exact same way of calculating reactive power as the HLW lib.

mcspr: hlw version takes a square root though
`_reactive = sqrt(apparent * apparent - active * active);`

AlbertWeterings: reactive power is the power where the current and voltage are not in phase. Calculating the reactive power is done in hlw lib in a way the phase difference is 90 degrees. In reality this is not always the case as it can be a different angle. so the calculation is more or less a approach I think. As long as we use the same approach for all power meters I can live with it.

By the way with coils, capacitors and shunt resistance you can play with phase difference in a circuit. That is why you often see in fluorescent lights with coils also capacitors to make sure blind power is reduced. To much blind power can heat up wiring and course fire but is not measured by power meters.

xoseperez: Apparent and active power are provided by the HLW chip itself. Assuming their values are correct, reactive power is their relation depends on the phase, as you say. The reactive power is the missing component, and it's module is the square root of the difference of the squares.

mcspr:
🔲 merged 🔲

Issue #648, opened 2018-03-09
RFBridge buzzer


myounges:
Hello first of all would like to thank you for your great work, my hat goes off for you.
I have an enhancement suggestion unless I'm not aware of possible ways to do it.

It is regarding the buzzer the RFBridge has included, is there any way to enable that an mqtt message can enable the buzzer for a second or so? Lets say that the buzzer is connected to GPIOX we can assign the mqtttopic/relay/4/set -->1/0 or similar. I believe it is not that easy as it doesn't seem to be connected to a GPIO directly, but it is worth asking.

I use the RFbridge to enable or disable an RF alarm and would be interesting to beep once the alarm is enabled.

thanks!
Michael

myounges:
Me again...

Doing some research I see the buzzer is connected to the EFM8BB1 chip. 
![2018-03-09 11_37_08-433rt sch 1 1](https://user-images.githubusercontent.com/13675206/37203578-68f2e352-238e-11e8-97ff-47510d61faaf.png)

Maybe through the UART pins, this can be done?

https://www.itead.cc/wiki/images/3/3e/Sonoff_RF_Bridge_433_Schematic.SCH.pdf


Cheers,

xoseperez:
Nom the buzzer is controlled by the EFM8BB1 and there is no command in the protocol to drive it from the ESP8266.

myounges:
Ok, thank you very much!

Portisch:
Beep will be supported with the new EFM8 firmware:
https://github.com/Portisch/RF-Bridge-EFM8BB1/commit/fc5774fb956d7b7610b2573d324393bdb2b5b1d2

AlbertWeterings:
As @Portisch has included the beep feature in his firmware I will later create a pull request to add the feature to ESPURNA. My idea is to add a MQTT topic where you can send the number of beeps you want to hear to. If anyone has a better or other idea to implement this then please share your thoughts.

myounges:
That would be great! Mqtt topic would be perfect!
thanks guys!

PR #1600, opened 2019-03-02
Added #define RF_DO_BEEP              0xC0

kiralikbeyin:
https://github.com/Portisch/RF-Bridge-EFM8BB1/blob/8f5de50c546053e560d1f35537560645af8fefcd/inc/uart.h


mcspr:
???

There's also version code and extended protocols... but some code should probably do something with them

kiralikbeyin:
https://github.com/xoseperez/espurna/issues/648

This is only a one raw code which may help someone in future

```
RF_DO_BEEP = 0xC0
```

xoseperez:
My opinion: do not add it until we are using it, otherwise it's a YANGNI

PR #1603, opened 2019-03-03
Add thermostat module

ElderJoy:


Misiu:
@ElderJoy this looks awesome. Do I understand that I can use a single relay and DS18B20 sensor connected to the same device or I can use two devices - one with relay and second with a sensor?
Description of changes and the functionality would be more than welcome 🙂

ElderJoy:
Hi @Misiu! Thank you. Sorry for lack of description. I promise to create it soon. For now briefly:
- I used this module in gas boiler controller. But I wanted to keep up remote temperature, that measured by another sonoff with DS18B20 in other room.
- Operation priority:
1. Remote temperature, that obtained by mqtt from remote sensor. You can setup remote sensor name in web page. Used standard tick data. Required Json message packaging.
2. If there is no remote sensor or it's data not actual - local sensor used. For case if something happened with remote sensor.
3. If there is no local sensor or it's data not stable yet - used timings for on\off intervals. For worst case to avoid freeze in heat system.

So you can use local or remote sensor for your choice. Just skip remote sensor name setup.

thermostat_web
img_20190203_153255

It also support I2C display where indicating status, temp. interval, local and remote temp.
On screenshot display 4 icons that mean (inverse in opposite case):
1. WiFi connection established successfully
2. Mqtt connection established successfully
3. openHab connection (responded on interval update request)
4. actual remote temperature present. 

Kamin - remote temp. source
Kitchen - local temp. 

I can add more pictures if needed.

xoseperez:
It looks like the `thermostat` and `display` modules are missing in the PR. Also, I would not change the `MQTT_USE_JSON` value generally since it will impact all devices. Instead, create a new device in the hardware module (`itead-sonoff-th-thermostat`?) and set there any specific flags it should have.

ElderJoy:
Hi Xose,
I going to fix conflicts, you've mentioned. There are several connected questions:
1. Missing in PR... Do you mean add thermostat and display to espurna_modules in progmem.h? Will add.
2. As I understand, in the index.html parts of code, that belong to correspondent module should be marked as removeIf(!thermostat) to exclude it if module does not used? Will do this.
3. You marked espurna.ino as conflicting around thermostat definition. What should be fixed there?
4. I agree with you that general change of MQTT_USE_JSON is not a good idea. But creating separate hardware module will limit module usage for particular device (TH?) of forced to create multiple devices with thermostat configuration. Probably it would be better to set MQTT_USE_JSON depending on THERMOSTAT_SUPPORT? It would not affect MQTT_USE_JSON in default case as THERMOSTAT_SUPPORT switched off by default but enable JSON for thermostat usage if one select it. Also THERMOSTAT_SUPPORT definition could be commented to notify this dependency.

ElderJoy:
Hi Xose,
I resolved mentioned conflicts:
1. Add thermostat to the espurna_modules in progmem.h
2. Add thermostat to webui, gulp, html, js.
3. In general.h MQTT_USE_JSON returned to 0 by default and 1 if thermostat enabled.

ElectricImpossible:
Looks very impressive, do you have a bin file that I can flash to my TH16?

ElderJoy:
> Looks very impressive, @ElderJoy do you have a bin file that I can flash to my TH16?

 My binaries outdated. I built them year ago. Espurna project developing quite fast, so it gone far since. So, I recommend you build it from hoseperez/espurna sources.

PR #1607, opened 2019-03-04
Allow to set relative brightness and relative color in mireds through…

tsymbaliuk:
… command using +N and -N notation

It allows not to ask for current value of brightness and color temperature, but increase/decrease it relatively. So controller like arduino board with buttons and infinite knobs might be more dummy and more efficient (since communication is not bidirectional).

Backward compatibility is almost untouched (except rare cases when people used commands like "brightness +50" instead of "brightness 50").

Example of commands:
brightness +50 // increase brightness for 50 points
mired -10 // decrease color temperature for 10 mireds
mired 100 // set color temperature to 100 mireds

### DIFF
--- a/code/espurna/light.ino
+++ b/code/espurna/light.ino
@@ -989,8 +989,15 @@ void _lightInitCommands() {
 
     terminalRegisterCommand(F("BRIGHTNESS"), [](Embedis* e) {
         if (e->argc > 1) {
-            lightBrightness(String(e->argv[1]).toInt());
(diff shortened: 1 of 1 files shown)


xoseperez:
Thank you!

mcspr:
+ update https://github.com/xoseperez/espurna/pull/1938/commits/919ba260a88ca8d8d28ff23e797a7b0f31bd7b7b includes the same change for MQTT and HTTP as well

Issue #1062, opened 2018-07-19
New Sync mode special for motors


glarfs:
Hi,
I'm using a 4ch board with espurna firmware, activating cover motors. The configuration I use:

1- motor A up
2- motor A down
3- motor B up
4- motor B down

At hardware level is configured to be only one motor up or down ( by default up will be used if both 1 and 2 relays are active (11) ). So on relay1=1 and relay2=1 will cause motor A up.

I checked the code as I wanted to add a new sync method "Zero or One switches  active on Pair relays" that allows only one on every 2 relays on but is not as clear/clean as the current options.

The problem is that If I use Zero or One  switches active I can't open both motors up at the same time.
I've also tried group sync but I couldn't find a good way to do it.

I also have another board with a single motor (relay1 and relay2) and a light on the relay3 and the relay4. So this new functionality won't apply on this other case, but that can be addressed on another issue.

I've also though to use a mask or something on each relay to specify the group.

AlbertWeterings:
Hi glarfs,

Maybe something to consider, I use one relay to set the motor direction 'nc' (off) is up the second relay is starting the motor. This way it is physical not possible to have UP and DOWN activated at the same time. I also found the self-induction of the motor can cause a reasonable spark inside a normal relay so I replaced the relay that activate the motor with a solid state version. The direction relay is a normal one as it will never switch when the motor is running it is not stressed.

hyteoo:
@glarfs from an electrical engineering standpoint, I would recommend you use a H-bridge controller which would have several benefits over relays, like:
- quiet
- control speed with PWM
- integrated protection against ShortCircuit, Overload, etc
- cheaper
- practically no way of accidentally driving a motor up&down at the same time

I could give an example that would fit your dual motor setup: http://www.ti.com/product/DRV8848
Of course everything depends on the motor power so if you could give us more details, that would be great ;)

xoseperez:
I think hardware-based control is a must when using motors, but you said you already have that covered.
From the software point of view, I like the relay groups idea. By default, all relays will belong to the same group and it will be the current behavior, but you could specify groups of 2 (or 2, 4) relays together. Then the sync options will apply per-group.

biot:
[...]

PR #1609, opened 2019-03-06
New relay sync mode where all additional switches follow the first one 

foxel:
The new sync mode (named "Switch #0 controls other switches" in UI) means that if switch #0 changes other switches will sync to it but not vice-versa 

This scheme come to my mind as a solution to a case when I want to have a switch turning all the relays ON and OFF in one click, but be able to control the secondary switches independently.

My UCs:
1. I have a set of high-power and low-power lights for my desk, the low-power one is controlled by motion detection thru home-assistant, and i have a button which controls the high-power one. With this sync mode when i press this button I get all the lights on or off, and motion detection controls only one light.
1. Over the dinner table I have a chandelier with two groups of bulbs, the chendelier is controlled thru 433 MHz RF remote on the wall and most common use of this is to turn the whole chendelier on/off, I could connect all the bulbs into one group but i still want to be able to set the lights to some dimmer state from my phone. This is also acheived with the synck mode suggested.

Do you think this is a valuable addition to the firmware?


mcspr:
I think this matches nicely with other logic here.

And maybe one could describe several of such groups on one device?
e.g. 8 switch board, switch#1 controls first 3 switches, switch#4 controls last 3 in a similar manner.
idk though how this can be described in the config yet. only rough idea:
`relayMode` reuse relay mode from this option
`relaySyncMap<relayID>` bitmap of switches that are synchronized with the main one when it changes, but independent otherwise. e.g. `relaySyncMap0 => 01110000`, `relaySyncMap4 => 00000111` for my example.

xoseperez:
Nice one.

This could be the answer for #1062. Still, I'd prefer to merge this as is and the look for a more general solution.

Issue #1279, opened 2018-10-16
board config files


biot:
The current board configuration system requires two entries in `platformio.ini`, and a lengthy entry in `config/hardware.h`. Since the config is based solely on C definitions, it's also not very flexible. For example, declaring multiple values would run into a syntax problem, and necessitate a clumsy workaround to get it configured.

Instead, I propose doing per-board configuration files, in `.ini` style like platformio for convenience. This would include all configuration currently in `hardware.h`, but opens the way for better future expansion: it's a much more flexible way to declare config.

Adding a new board is just a matter of adding a single, small config file.

The device config file gets read during the `pio run` stage, which is called with the board name in the environment variable `ESPURNA_BOARD`, and converted to  entries in `BUILD_FLAGS`. The board name has an entry in the top-level `devices/` directory, with `.ini` extension.

I'm adding a first patch (on top of dev branch) that does the above, and adds config files for all known devices. These were auto-generated from their `hardware.h` entries so as to avoid pin number mistakes when transcribing. Some comments were preserved where possible. This saves tons of entries in platformio.ini, and removes the `hardware.h` file entirely.

Not everything is finished, but I wanted to get some comments in before I take this further. 
I test with the Sonoff Dual R2, and certainly it works there, but that means I've only really tested config entries for manufacturer/device, buttons, relays and LEDs. Other definitions from `hardware.h` get translated just the same to `.ini` entries, but may not generate the right `BUILD_FLAGS` definition yet. That's the following `.ini` sections:
  - cse7766
  - dallas
  - debug
  - dht
  - dummy
  - hlw8012
  - led1
  - light
  - my92xx
  - relay

...and lots of keywords in those sections, everything but pin/pin_inverse/relay/type. Lots of these will do the right thing, but are just unchecked/untested.

Code is here: https://github.com/biot/espurna

mcspr:
Isn't it just trading one format for the other? And Arduino IDE compatibility is still expected, so you might want to generate hardware.h back from the device configs instead of removing it :) (see code/espurna/config/arduino.h) One can also include custom platform.txt, but idk how it would look for this many boards.

As for numbered tokens, there are actually more options:
1. Name them instead. Some section naming trickery?
```ini
[button:green_one]
pin = 0
...

[relay:main]
button = green_one
pin = 3
...
```

2. Have it like you propose, but use .ini list syntax and have some substitution in place (examples from biot/espurna still use button1, button2, [light]/ch1_... etc.):
```ini
[...]

PR #1611, opened 2019-03-06
Added Teckin SP20 Power Plug.

brmo:


mcspr:
@brmo thanks!
@xoseperez there are some boards that got through pr recently without migrate.ino changes. both parties forgot it exists :> i think i got a shortcut to migration without something drastic like https://github.com/xoseperez/espurna/issues/1279 or v2 config files, will try to test it soon

edsai:
I can't tell which release to use to flash the SP20. I see a binary for the SP22 but not sure if I'm supposed to use that or espurna-1.13.5-espurna-core-1MB.bin

PR #1618, opened 2019-03-08
Added LITESUN LA-WF3 support.

Zebble:
Support for LITESUN LA-WF3 / Costco Charging Essentials wall socket.


mcspr:
I've updated hw define to have underscore instead of hyphen, so MANUFACTURER + _ + DEVICE produce the board string
(sorry, that should probably be mentioned in the hw header top comment...)

Thank you! 🌻

Issue #1621, opened 2019-03-09
Xiaomi Desk Lamp - Encoder Read Correctly, But Doesn't turn on


thehellodonkey:
Thanks for the great firmware, look forward to your thoughts on the issue below.

**Bug description**
Light doesn't turn on regardless of encoder inputs or switch/brightness/color settings in web interface.

**Steps to reproduce**
Load firmware on Xiaomi Smart Desk Lamp, attempt to turn on.

**Expected behavior**
Light on/off status reflects switch position in web interface

**Screenshots**
This shows web interface responding as expected to all encoder actions, but light does not turn on.
![image](https://media.giphy.com/media/3JNNxQRNbAZ13z8uEh/giphy.gif)

**Device information**
```
[753816] [MAIN] ESPURNA 1.13.5 (83195fd)
[753817] [MAIN] xose.perez@gmail.com
[753818] [MAIN] http://tinkerman.cat

[753819] [MAIN] CPU chip ID: 0xA30B3B60
[753821] [MAIN] CPU frequency: 80 MHz
[753825] [MAIN] SDK version: 1.5.3(aec24ac9)
[753829] [MAIN] Core version: 2.3.0
[753832] [MAIN] Core revision: 159542381
[753835] 
[753836] [MAIN] Flash chip ID: 0x1540C8
[753841] [MAIN] Flash speed: 40000000 Hz
[753843] [MAIN] Flash mode: DOUT
[753846] 
[753848] [MAIN] Flash size (CHIP)   :  2097152 bytes /  512 sectors (   0 to  511)
[753860] [MAIN] Flash size (SDK)    :  2097152 bytes /  512 sectors (   0 to  511)
[753862] [MAIN] Reserved            :     4096 bytes /    1 sectors (   0 to    0)
[753871] [MAIN] Firmware size       :   501840 bytes /  123 sectors (   1 to  123)
[753879] [MAIN] Max OTA size        :   520192 bytes /  127 sectors ( 124 to  250)
[753887] [MAIN] EEPROM size         :     4096 bytes /    1 sectors ( 251 to  251)
[753893] [MAIN] Reserved            :    16384 bytes /    4 sectors ( 252 to  255)
[753900] 
[753902] [MAIN] EEPROM sectors: 507, 506
[753905] [MAIN] EEPROM current: 506
[753907] 
[753909] [MAIN] EEPROM:  4096 bytes initially |  1315 bytes used (32%) |  2781 bytes free (67%)
[...]

PR #1627, opened 2019-03-10
Update hardware.h

Ctrl-F4:
fix for Xiaomi Desk Lamp (#1621)


mcspr:
@Ctrl-F4 Thank you!
@1043717432 this should fix the on / off problem

PR #1639, opened 2019-03-17
Specific Sonoff iFan02 changes

fquinto:
Improvement only for specific Sonoff iFan02:
- Add new MQTT topic for light (relay 0, first relay)
- Adding GPIO2 to low
- Adjust configuration relays
- According coding style guide, removed some unused spaces

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -123,7 +123,7 @@
     #define RELAY1_PIN          5
     #define RELAY1_TYPE         RELAY_TYPE_NORMAL
 
-    // Light RGBW 
+    // Light RGBW
     #define LED1_PIN            2
     #define LED1_PIN_INVERSE    1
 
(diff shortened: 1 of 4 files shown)


code/espurna/relay.ino
```diff
@@ -241,6 +241,19 @@ void setSpeed(unsigned char speed) {
     }
 }
 
+unsigned char getLight() {
+    unsigned char light = (_relays[0].target_status ? 1 : 0);
+    return light;
+}
+
+void setLight(unsigned char light) {
+    if ((0 <= light) & (light <= 1)) {
```
mcspr: Isn't this essentially renaming `relay/0` topic to simply `light`? We can use relayParsePayload to parse toggle and some strings too, payload becoming either 0 (off), 1 (on) or 2 (toggle). relayStatusWrap directly accepts it's result.

And unless I am missing something, did you mean && instead of &?

fquinto: Yes. Is specific for the **Sonoff iFan02**: the relay 0 is `light`.
I don't know that we can use `relayParsePayload`.
This function is a _copy and paste_ from `setSpeed` and is just the same at this line: https://github.com/xoseperez/espurna/blob/dev/code/espurna/relay.ino#L234

code/espurna/config/hardware.h
```diff
@@ -3147,7 +3166,6 @@
     #define BUTTON2_LNGLNGCLICK     BUTTON_MODE_RESET
 
     // Light
-    #define RELAY_PROVIDER      RELAY_PROVIDER_LIGHT
```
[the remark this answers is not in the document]
fquinto: I don't remember that. I think I don't remove this line (3150). Sorry for that.

code/espurna/relay.ino
mcspr: But the first one is real though, I think we totally can.

About speeds, see the speed bitmap above. Each speed is one combination of relays, so we are using AND to set desired speed setting.
We should probably introduce Rotary Switch type for this (i.e. more that 0 or 1 for state)

fquinto: What is your proposal to do now?

mcspr: Second part is suggestion in general, not for this pr.

The way that setSpeed works is not really applicable here. We can just use relayParsePayload when parsing MQTT / API string and pass it to helper function that I mentioned: https://github.com/xoseperez/espurna/blob/0987e0195041b39109f775865ac16bd14b2801a6/code/espurna/relay.ino#L861

code/espurna/config/hardware.h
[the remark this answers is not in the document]
mcspr: No worries btw. I think just merging dev will fix this

mcspr:
Closing via https://github.com/xoseperez/espurna/compare/8ceeebdb24aa9bd863fa28676a0364497ec193ae...df0f1a2e19e90295905bc7d35dec61090029f6a6, since it is no longer mergable
Just got around to it 😴t.

PR #1645, opened 2019-03-21
Add option to disable relay reporting to MQTT

Niek:
This feature is especially useful when using custom MQTT groups for relays. This adds the option to disable the reporting of relay statuses to MQTT, and report only to the custom MQTT groups (if defined).

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -356,6 +356,10 @@
 #define RELAY_SAVE_DELAY            1000
 #endif
 
+#ifndef RELAY_REPORT_STATUS
+#define RELAY_REPORT_STATUS         1
+#endif
+
 // Configure the MQTT payload for ON/OFF
 #ifndef RELAY_MQTT_ON
 #define RELAY_MQTT_ON               "1"
--- a/code/espurna/mqtt.ino
+++ b/code/espurna/mqtt.ino
@@ -237,7 +237,7 @@ void _mqttConfigure() {
     // Getters and setters
     _mqtt_setter = getSetting("mqttSetter", MQTT_SETTER);
     _mqtt_getter = getSetting("mqttGetter", MQTT_GETTER);
-    _mqtt_forward = !_mqtt_getter.equals(_mqtt_setter);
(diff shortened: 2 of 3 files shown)


Niek:
It might make sense to change all ``mqttForward()`` calls in the code as well (which is determined by a check if ``MQTT_GETTER != MQTT_SETTER`` - I don't really understand the logic there).

mcspr:
> which is determined by a check if MQTT_GETTER != MQTT_SETTER - I don't really understand the logic there

For example, if both are empty (and therefore equal), we are subscribed to the `<root>/relay/0` and will also use it to send our state.

Niek:
> > which is determined by a check if MQTT_GETTER != MQTT_SETTER - I don't really understand the logic there
> 
> For example, if both are empty (and therefore equal), we are subscribed to the `<root>/relay/0` and will also use it to send our state.

OK that makes sense - I changed the code to use ``mqttForward()`` instead.

mcspr:
@Niek `mqttForward()` is not always available. 300KiB builds break because of that:
https://travis-ci.org/mcspr/espurna-nightly-builder/jobs/512070680#L984-L990

Niek:
Ah I didn't think of that, what do you think would be the cleaned approach? Make a dummy ``mqttForward()`` function in relay.ino in case ``MQTT_SUPPORT`` is false? Or duplicate those calls with ``if #MQTT_SUPPORT`` blocks around it?

Niek:
I submitted a fix in PR https://github.com/xoseperez/espurna/pull/1662

mcspr:
IDK though why it must depend on it. Forwarding setting is only ever checked when relay status (or light, or domoticz) function is called from within MQTT callback, so that we do not report back to the same topic we just received data from. Now, it is always called.
Heartbeat mechanism also can send all relays status, which forwarding setting && new define does not stop, but it can be disabled at runtime or via define flag default (see `HEARTBEAT_REPORT_RELAY` in utils)
So maybe it should do something similar and just wrap `relayMqtt(id)` with optional setting / define and do nothing? Or there can be a more elaborate reporting mechanism, following the idea of broker module.

PR #1647, opened 2019-03-22
sensors - allow a second BMX280 sensor

CraigMarkwardt:
Up to two BMX280 sensors are allowed on a single I2C bus, with addresses 0x76 and 0x77.  Currently espurna only supports one of them.

This change adds a second sensor object to deal with two sensors, if BMX280_ADDRESS2 is defined.


CraigMarkwardt:
Don't know how to resolve the Codacy quality review, either BMX280_ADDRESS2 is defined or not, and the firmware builds either way.

xoseperez:
Hack-ish but it works, of course. You don't have to specify the individual addresses. Setting both to 0x00 will work since the auto-discover procedure will take care of the already used ones.

Maybe it will be better to add a BMX280_NUMBER to specify the number of sensors connected (1 or 2).

CraigMarkwardt:
OK I can re-work it, but... non-auto addressing is important.  Order matters.  If I have sensor 0x76 on the "incoming air" and sensor 0x77 on the "outgoing air" I don't ever want those mixed up.

xoseperez:
It will always load sensor at 0x76 first. But yes, if you want to be define the order yourself then you are right.

PR #1648, opened 2019-03-22
Sensors can determine precision of their magnitudes

CraigMarkwardt:
Currently espurna magnitude precisions are hard-coded in sensor.ino. For example the default temperature precision is whole integer degrees Celsius.  This causes sensors such as the Dallas DS18B20 to miss 3 digits of extra precision.

This change modifies BaseSensor so that each sensor class can specify the number of decimals to represent its magnitude types. If the sensor does not override this virtual method, the default is to use the sensors.ino hard-coded values (i.e. fall back on previous method).

I also updated sensors BMX280 and Dallas OneWire are with the specified precision knowledge; tested with BME280 and Dallas DS18B20 sensors

### DIFF
--- a/code/espurna/sensor.ino
+++ b/code/espurna/sensor.ino
@@ -20,6 +20,7 @@ typedef struct {
     BaseFilter * filter;        // Filter object
     unsigned char local;        // Local index in its provider
     unsigned char type;         // Type of measurement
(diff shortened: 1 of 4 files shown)


xoseperez:
LGTM. Thank you very much!

PR #1650, opened 2019-03-23
Adding support for Generic GU10 bulb

### LINKED ISSUE #1549: GU10 Lamp - Similar to AI-Thinker
I have got this GU10 led lamp from [Ebay ](https://www.ebay.com/itm/1-10PC-GU10-RGB-Smart-Bulb-Wireless-WiFi-App-Remote-Ctrl-Light-for-Alexa-Google/173724116351).

I assumed that it will be similar to AI-Thinker lamps as it uses same mobile app etc.
Board is in fact very similar, well half of it...

I tried to flash espurna-1.13.3-aithinker-ai-light.bin but no matter what I do in the web interface light is still at maximum brightness.

20190213_201155 - kopie

Chip at the top left is SLM211A (LED PWM driver) which is controlling white channel on GPIO4.
~~It is using GPIO0, GPIO2 & GPIO15 for RGB, I cannot say exactly which one is which but I'm going to figure it out.~~

I will add more info.

ColinShorts:
This might also support more than just GU10 devices, but resolves #1549.


ColinShorts:
Thanks Xose!

PR #1664, opened 2019-03-29
PlatformIO: silence postmortem stackdump for boards using serial

mcspr:
Dummify hardcoded `ets_printf` / `ets_printf_P` from postmortem.
It uses serial to print stackdump whenever board reboots after an error, and we do not want that when using serial connection with some external <s>device</s> chip.

ref: https://github.com/esp8266/Arduino/blob/f139519061bdb81dfc6d41e9cd704fd455f1647a/cores/esp8266/core_esp8266_postmortem.cpp#L85
https://github.com/SuperHouse/esp-open-rtos/issues/188


mcspr:
*tested locally with espurna-core bin and `Serial.begin(115200)` in setup(). `erase.config` in telnet session prints small stack-trace to the serial in normal builds, and no longer does with this one*

After I did this, I wonder why I have not thought about just using `Serial.end()` 🤔

PR #1679, opened 2019-04-04
Send RF MQTT message with retain off

Niek:
Small fix - I think it makes sense to always send RF MQTT messages with retain off. I can't really think of a use case where you always want to retain the last rfin payload.


Niek:
The build error is not related to these changes

Issue #1465, opened 2019-01-02
Current status of MQTT SSL


vogelfreiheit:
Running latest master branch as of today, I'm not sure anymore if the SSL support still conflicts with the web UI regardless of the platform used (I'm building with 180).

It would be nice to have a current document on SSL on its own page, as well as tweaks to enable HTTP access for the UI but MQTT over SSL, to minimize the performance requirements.

mcspr:
Some notes after testing:
- Core 2.4.2 / async-mqtt-client
is ok'ish with both mqtt and web. can save settings, can view debug log. once crashed on page reload
- Core 2.4.2 / lwip2 (low/high) / pubsubclient / modified mqtt module to support bearssl wificlient
works ok with basic payloads, but crashes with mere 260 bytes of json. upping pubsubclient mqtt packet buffer size does not help, neither is tuning in/out fragment length of the wificlient. web ok
- Core 2.5.0 beta (platform staging) / lwip2 low / same pubsubclient - works ok for both basic and json. could not get web to crash.

All tests done with local server, because I could not get cloudmqtt.com to work. It just crashes the board (or fails to connect at the best of times). Tried async/axtls builds, new sync/bearssl, even tried different firmware from mongoose os - nothing works with them :/

Pubsubclient + bearssl combo can also use CA cert instead of fingerprinting, either as hardcoded string or spiffs file. Also, fingerprinting api is different and there is no need to manually do it (client.connect, check fp, mqtt.setClient, mqtt.connect repeat)

Niek:
My experience with core 2.4.2:
* aync-mqtt-client works, but disconnects randomly when sending or receiving messages. I thought it was related to https://github.com/marvinroger/async-mqtt-client/issues/107, but disabling ``setWill()`` doesn't help
* pubsubclient doesn't work at all for me, it keeps disconnecting at connection stage - maybe interfering with web?
* pubsubclient + bearssl is no option because the binary size grows with 100KB+ - maybe using ``BearSSL::WiFiClientSecure`` results in both SSL variants to be compiled in?

I tried testing with core 2.5.0 as well, but unfortunately the binary grows too large to do OTA on a 1MB board - no matter which MQTT library I use.

Niek:
MFLN is interesting - if you compile your broker against OpenSSL 1.1 it should work out of the box. It should be quite easy to do a couple of quick ``probeMaxFragmentLength()`` calls in the MQTT module before connecting.

[...]

PR #1682, opened 2019-04-05
Allow empty fingerprints of MQTT SSL server

Niek:
This change allows empty fingerprints of MQTT SSL servers - so bypass key pinning.

Of course this is not really secure, as it allows MITM attacks with fake signatures. But in my opinion this is worth supporting, this way it's much easier to get started with a MQTT/SSL broker setup. In addition, LetsEncrypt certificates change signature every renewal (90 days), so it makes it possible to work with LE-issued certificated without changing the fingerprint at every renewal.

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -706,9 +706,11 @@
 // you will have to disable all the modules that use ESPAsyncTCP, that is:
 // ALEXA_SUPPORT=0, INFLUXDB_SUPPORT=0, TELNET_SUPPORT=0, THINGSPEAK_SUPPORT=0 and WEB_SUPPORT=0
 //
-// You will need the fingerprint for your MQTT server, example for CloudMQTT:
-// $ echo -n | openssl s_client -connect m11.cloudmqtt.com:24055 > cloudmqtt.pem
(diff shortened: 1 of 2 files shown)


mcspr:
Maybe additional settings flag instead? something like `mqttInsecure`

Another thing, from #1465 : we can use PubSubClient with real certificates [starting with Core 2.4.2](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/bearssl-client-secure-class.html) (although, as I've noticed, it might choke on some specific certificate types. cloudmqtt was one of them)
It is a separate thing from fingerprinting and might also use such option.

Niek:
Ah, I was not even aware of the other issue.

In case support for real certificates is added in the future, it might make sense to use something like `MQTT_SSL_VALIDATION`` which can be none/insecure, fingerprinting, and in future known key (even though it has the same issue as fingerprinting), or trust anchor (maybe in combination with SNI check).

I also have some issues with MQTT/SSL timing of messages, but I guess the other topic is a better place for that.

Niek:
Will be updated once #1751 is merged

Niek:
Closed in favor of https://github.com/xoseperez/espurna/pull/1829

PR #1690, opened 2019-04-10
Update BMX280Sensor

### LINKED ISSUE #1672: BME280/BMP280 sensor shows 'Wrong ID' in values section.
I have setup a Sonoff basic to use Pins 1 and 3 (Tx/Rx) for a BME280, 4 pin sensor.  I have the following in custom.h
```
#define BMX280_SUPPORT 1
#define BMX280_ADDRESS 0x00
#define I2C_SDA_PIN 1  //TX PIN
#define I2C_SCL_PIN 3  //RX PIN
```
Once built I get the following:

status
and at Debug I first get: 

```
[SENSOR] Error reading data from BMP280 @ I2C (0x00) (error: 4)`
[SENSOR] Initializing BMP280 @ I2C (0x00)
[010015] [I2C] Address 0x76 locked
[010031] [SENSOR]  -> temperature:0
[010033] [SENSOR]  -> pressure:0
[010037] [SENSOR]  -> humidity:0
[030015] [SENSOR] Error reading data from BME280 @ I2C (0x76) (error: 6)

```
After that I then get:

`[SENSOR] Error reading data from BMP280 @ I2C (0x00) (error: 4)`

I have also tried setting 0x58, 0x77 and 0x76 but get the same status page as above. I guess espurna is not setting the address, or do I need to install pull up resistors on the data lines? The BME280 is connected using twisted pair twisted with positive and ground and the wires are about 1 metre long.

mcspr:
resolves #1672
- default to one sensor
- dynamic number and address setting
- do not return i2c-specific error for wrong values
(unless it is some value returned by Wire.read() and not sensor?)

### DIFF
--- a/code/espurna/config/sensors.h
+++ b/code/espurna/config/sensors.h
@@ -159,7 +159,7 @@
 #endif
 
 #ifndef BMX280_NUMBER
-#define BMX280_NUMBER                   2       // Number of sensors present. Either 1 or 2 allowed
+#define BMX280_NUMBER                   1       // Number of sensors present. Either 1 or 2 allowed
 #endif
 #ifndef BMX280_ADDRESS
 #define BMX280_ADDRESS                  0x00    // 0x00 means auto (0x76 or 0x77 allowed) for sensor #0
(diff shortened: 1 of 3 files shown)


mcspr:
And another thing - maybe address setting should use HEX format?
Maybe optional, with 0x prefix check

Issue #1597, opened 2019-03-01
Sonoff RF Bridge cant change "Boot Mode" of switch


kiralikbeyin:
Espurna 1.13.5
rc2.5.0 and rc2.4.2

When i  want to change "Always OFF" to "Same as before" it doesnt save and restarts... 
So my switches turns off everytime SRFB restarts.

reaper7:
looks like common problem with "save".
My S20 and RF Bridge do not save settings.
Only Sonoff T1 save settings but I do not know how I forced him to do it...
It seems to me that I had to enable the following options:
TELNET_SUPPORT & DEBUG_TELNET_SUPPORT & DEBUG_WEB_SUPPORT
from this moment pressing the SAVE key on the web interaface began to save the settings.

Espurna current git / LWIP 1.4

mcspr:
Can't reproduce with rfbridge built with 2.4.2, with or without DEBUG_SUPPORT flags active.
Does using lwip2 make any difference?
Does setting `LOOP_DELAY_TIME` to `10` change anything? (or without rebuilding - `loopDelay` setting via terminal / settings upload)

kiralikbeyin:
@reaper7 yes, save problem. TELNET_SUPPORT & DEBUG_TELNET_SUPPORT & DEBUG_WEB_SUPPORT didnt work Espurna current git / LWIP 1.4 AND Espurna current git / V2 Low mem didnt work.

@mcspr loopDelay 10, 20, 55, 200 didnt work

I have a IR RECEIVE problem maybe these problems are related. https://github.com/xoseperez/espurna/issues/1596

I will fallback to 1.13.3 v

kiralikbeyin:
I uploaded espurna-1.13.5-itead-sonoff-rfbridge.bin with esptool.py and it saves no reboot.

I think main problem is we are downloading external libraries like https://github.com/markszabo/IRremoteESP8266 and other 10-15 libraries and they are crushing something. (i downloaded all latest external libs and i cant save which is the most basic thing)

**Please add external libaries with exact relase in source.zip of nice Espurna.**
Also please give info which "rc version" and "lwIP Variant" is used for these bin files.
@xoseperez Thanks alot!

Is there any way to enable IR receive-send from terminal?

reaper7:
my observations:

---

custom config:
```
  #define WIFI_FALLBACK_APMODE        0                                         // Fallback to AP mode if no STA connection
  #define USE_PASSWORD                0

[...]

Issue #1115, opened 2018-08-07
MQTT on -> off freezes the device


jsponz:
With the last version, when switching off the MQTT, device freezes.  I have to plug it off and plug it on again. 

When switching on the MQTT, there is no problem. 

Any hint?

xoseperez:
Cannot reproduce the issue... can you check the debug log to see if there is any useful info there?

jsponz:
The problem is the library async-mqtt-client.

ESP just restarts and the web loses connection. You need to refresh the explorer.

xoseperez:
How did you get to that conclusion? What version/revision of the library are you using?

pniaps:
Hi @jsponz ,
When you hit `Save` button, if MQTT is enabled, it disconnects and connects again, and the problem is in the library async-mqtt-client. When it disconects, it crashes. The ESP reboots, but the web remains frozen until you reload it.

The problem is

https://github.com/marvinroger/async-mqtt-client/blob/33cdd5038b8b411db47322908ca4eb503f630909/src/AsyncMqttClient.cpp#L707

function **_sendDisconnect** closes connection on line 661, so **_client.send()** on line 707 crashes.

https://github.com/marvinroger/async-mqtt-client/blob/33cdd5038b8b411db47322908ca4eb503f630909/src/AsyncMqttClient.cpp#L661

I will send a pull request to https://github.com/marvinroger/async-mqtt-client

xoseperez:
Please, let me know when you send the PR.

Misiu:
@pniaps did You send PR for this?

Misiu:
@pniaps any news?

pniaps:
Hi.
pull request sent https://github.com/marvinroger/async-mqtt-client/pull/117

Sorry for the delay

mcspr:
@pniaps I'll wait for someone to answer the pr you proposed, but isn't that the case of having network request inside system callback? Inside WiFi event callback that is. If you change the callbacks to set boolean flags and do the connection inside loop() instead everything works ok. And, in espurna case, that is exactly what it does.

I did track this for some time but could not reproduce the original issue :/ 2.3.0 ... 2.4.2 did ok switching mqtt on / off / on again.

pniaps:
Will try, but callback or not, '_client.send();' is after '_client.close(true);'

mcspr:
Some notes, since PR was never merged and this is still a possible bug:
[...]

PR #1701, opened 2019-04-17
Safer settings reload and MQTT change detection

mcspr:
- remove ws module change detection specific to `MQTT_SUPPORT`
- wait until loop to reload settings (to avoid issues when espurnaReload is called from async context:  https://github.com/xoseperez/espurna/issues/1115, https://github.com/xoseperez/espurna/issues/1597)
- move const chars into String. should be slightly more heap, but avoids some of the copying and allows to use String convenience methods. mqtt-client APIs are given const pointer to the internal buffer.
small optimization is to also set them to default inplace (`String _mqtt_server(MQTT_SERVER);`) 
- apply placeholders in config()
- keep host (aka mqttServer) in memory for secure client, because it takes much more time to connect


mcspr:
The best effect for settings is running Core 2.5.1 and up, with SSO optimization for String. Negligible effect when strings are changing size (but still present when changing topic).

To keep memory in check with current setup - `_mqtt_setting_string = String()` invalidates internal buffer -> getSetting creates new one. If i underestimate "negligable" part, we can just invalidate **all** settings, like it was before with connection part, and read them in order again (Added benefit is that String object uses aligned sizes for strings)

In general, something can be done to getSetting / Embedis / some other storage abstraction. Embedis itself already uses 4k of eeprom, in theory some String-like object can point directly to the correct position in that buffer. However, if something changes everything needs to be invalidated.

Issue #1510, opened 2019-01-27
Interlock by software. Controling blinds, curtains or motors


oscarsan1:
Going on with issue:
Interlock by software on Sonoff T1 2 gang, or Sonoff Dual
https://github.com/xoseperez/espurna/issues/491

When control curtains or motors with T1 2 gang I'm afraid of this ms that two realys are on. Is safer to take some time to switch off before switch on the new relay. This time is also good for the motor to stop moving before start in the other direction. Also relays offen takes more time to switch off (15ms) than to switch on (5 ms).
I propose to modify  "void relaySync(unsigned char id)" in the case of RELAY_SYNC_ANY

        if (relaySync != RELAY_SYNC_ANY)
        {
            for (unsigned short i = 0; i < _relays.size(); i++)
            {
                if (i != id)
                {
                    if   ( relayStatus(i) ){  // if relay i is on
                               relayStatus(i, false);
                                               //delay 500 ms before on for wait to turn off the other relay
                               _relays[id].change_time = millis() + 500; 
                   }
                }
            }
        }

This will work?

oscarsan1:
Finaly I can test it, fixed a bug in last proposal. I test this one and works well. Only one relay on and at lest 500 ms between relay "x" off an relay "y" on:


void relaySync(unsigned char id){
...
       // If NONE_OR_ONE or ONE and setting ON we should set OFF all the others
    }
    else if (status)
    {
        if (relaySync != RELAY_SYNC_ANY)
        {
            for (unsigned short i = 0; i < _relays.size(); i++)
            {
                if (i != id){
                    if   ( relayStatus(i) ){  // if current status relay i is on
                        _relays[id].change_time = millis() + 500; //delay 500 ms before on for wait to turn off the other relay
                    }
                    relayStatus(i, false);  
               }
            }
        }

        // If ONLY_ONE and setting OFF we should set ON the other one
    }

xoseperez:
`relayStatus` does not change the relay status right away, it only enqueues the change. It the `_relayProcess` method that processes the queue and calls `_relayProviderStatus` that executes the changes. The `_relayProcess` is first called to turn whatever has to be turned off and the whatever has to be turned on. I think it should be in between this two call where a safety delay should be added, probably in the `_relayLoop` method that does both calls.

oscarsan1:
Yes I see now.
[...]

Issue #1505, opened 2019-01-23
set configuration options and settings via MQTT (Telnet-MQTT-proxy?) ?


schweini:
I have a couple of espurna-powered Sonoff POWs deployed 'in the wild'. They are behind different types of NATs and firewalls, and usually impossible to access 'from the outside'. When I set them up, I just make sure that they can connect to my MQTT server, and let them be.

Is there some way to set (and save) configuration parameters (secondary wifi network, sensor configuration, etc.) via MQTT messages? This would be a very useful feature, since MQTT bypasses the whole NAT issue.

Maybe a MQTT-telnet-proxy topic I could send telnet commands to, and get telnet responses back? IIRC, everything can be done over telnet, so that would allow everything to be set via MQTT, too?

schweini:
I'd just like to re-iterate that this functionality would be extremely
useful for remote Espurna management, and IMHO relatively easy to implement.

(And I'm responding so that the issue doesn't become "stale")

mcspr:
Also depends on #1234, because we may need to buffer incoming messages (e.g. we send whole config) 
Basic idea is still the same - subscribe to a specific topic and apply all kv pairs as settings?

schweini:
For me, any kind of NAT-traversal-able way to change config things would be
very very useful.
Passwords, wifi settings, sensor calibration, etc.
Pushing whole firmware images wouldn't be necessary because of nofuss, I
guess.

schweini:
Huh. 
I just searched around the code, and two interesting things popped up:
uartmqtt.ino seems to be a MQTT-UART-gateway. Maybe this could be used to send "telnet" commands to the serial port?

mqtt.ino
subscribes to the MQTT topic 'actions', and if it received 'reboot', it does just that.
It shouldn't be that hard to find the espurna call for "set configuration key X to Y" and the call for "save config" and add it there.

I'll tryo too find that part some other day, but if someone knows the relevant calls, we could implement this relatively quickly via the /action topic, i think?

schweini:
Sorry to use this issue as a notepad of sorts, but:
in `telnet.ino`, all that seems to be done is to call `terminalInject(data, len);` with the received line.
So maybe, just maybe, adding `terminalInject(data, len);` in mqtt.ino could to the whole terminal-MQTT thing. hmm.

mcspr:
Sorry, I was a bit carried away with this.

[...]

PR #1705, opened 2019-04-26
Lock relay status on boot

### LINKED ISSUE #1697: Disable the switch option
We use the Sonoff POWs to measure power on devices **that we never want to turn off**.  

Can we have an option in the Switches menu that is "I really mean Always ON!" instead of just being on at boot mode.  I don't want it ever to go off if someone taps the switch or sends an MQTT relay/off message to this device.  

We already have other ways of turning our machines off, like kill switches and a separate power unit that turns a whole lot of other machines off at the same time.  

Having this easy off switch in the Sonoff is a liability.  But I want it to be easy to read the power and do the other config features.  

Adding this lockin feature into the Boot mode option would be the minimum to achieve this.  But you could add a new thing on the API that locks or unlocks the switch remotely.  

But the really main thing is we'd like to easily disable that button as a switch.

mcspr:
resolves #1697 

- `relay_t::lock` flag to control which target status we want. after relay changes to that status, we can no longer change it
- boot mode setting to lock relay in either OFF or ON state

cc @goatchurchprime @xoseperez 
this *might* interfere with planned motor changes from #1510 ?


mcspr:
cc @lancelon
this will lock the relay instead of disabling the button. meaning, you can still receive button events if you want that

and plugging #1680 again, where you can modify hardware preset and just remove button gpio in config. also, see #1505 (instead of #1322 proposal).

xoseperez:
I would add some visual notification to the web UI when the switch linked to the button is locked. Like disabling the toggle in the UI...

mcspr:
"input[disabled]" does just that. 
However, if lock becomes togglable, message should go with relayStatus instead. Maybe even nested within new relayState: {status: [...], lock:[...]}, since processData can encounter lock before any UI is initialized (can't depend on proper key ordering)

PR #1716, opened 2019-05-13
[WIP] Remove ESPAsyncTCP dependency of OTA code, SSL fixes, better parsing

Niek:
This PR changes the following:
- Update GitHub SSL fingerprint
- Remove ESPAsyncTCP dependency in OTA module
- OTA uses SSL with AxTLS on core < 2.5.0 and BearSSL on 2.5.0
- Better OTA response header parsing

Please test carefully, but I had no issues on core 2.4.2 and 2.5.0 on my setup.


code/espurna/ota.ino
```diff
@@ -55,107 +69,123 @@ void _otaFrom(const char * host, unsigned int port, const char * url) {
...
+        if (port == 443) {
+            WiFiClientSecure _ota_client = _ota_client_secure;
+            connected = _ota_client.connect(host, port);
+            if (connected) {
+                char fp[60] = {0};
+                if (sslFingerPrintChar(getSetting("otafp", OTA_GITHUB_FP).c_str(), fp)) {
+                    #ifdef USING_AXTLS
+                    if (!_ota_client.verify(fp, host)) {
+                    #else
+                    if (!_ota_client.setFingerprint(fp)) {
```
mcspr: That means it is bearssl, so we need to sed fingerprint before connecting.
Can't check rn, but doesn't it also allow to pass char'p directly?

code/espurna/ota.ino
```diff
@@ -55,107 +69,123 @@ void _otaFrom(const char * host, unsigned int port, const char * url) {
...
+    // Disabling EEPROM rotation to prevent writing to EEPROM after the upgrade
+    eepromRotate(false);
+
+    DEBUG_MSG_P(PSTR("[OTA] Downloading %s\n"), _ota_url);
+    char buffer[strlen_P(OTA_REQUEST_TEMPLATE) + strlen(_ota_url) + strlen(_ota_host)];
+    snprintf_P(buffer, sizeof(buffer), OTA_REQUEST_TEMPLATE, _ota_url, _ota_host);
+
+    _ota_client.print(buffer);
+
+    // Skip the response headers
```
mcspr: Any reason not to use HTTP client class? We can also check response code here.
Updater will fail verifying the data though... So maybe not needed?

code/espurna/ota.ino
```diff
@@ -30,8 +30,22 @@ void _otaLoop() {
...
 
-#include <ESPAsyncTCP.h>
-AsyncClient * _ota_client;
+#ifndef ARDUINO_ESP8266_RELEASE_2_5_0
+#define USING_AXTLS // do not use BearSSL
+#endif
+
+WiFiClient _ota_client;
+
+#if ASYNC_TCP_SSL_ENABLED
```
mcspr: OTA_SECURE_ENABLED?

code/espurna/ota.ino
```diff
@@ -30,8 +30,22 @@ void _otaLoop() {
...
+    #ifdef USING_AXTLS
+        #include "WiFiClientSecureAxTLS.h"
+        axTLS::WiFiClientSecure _ota_client_secure;
```
mcspr: using namespace axTLS / BearSSL?

code/espurna/ota.ino
```diff
@@ -30,8 +30,22 @@ void _otaLoop() {
 
 #if TERMINAL_SUPPORT || OTA_MQTT_SUPPORT
 
-#include <ESPAsyncTCP.h>
-AsyncClient * _ota_client;
+#ifndef ARDUINO_ESP8266_RELEASE_2_5_0
+#define USING_AXTLS // do not use BearSSL
+#endif
+
+WiFiClient _ota_client;
```
mcspr: Using single _ota_client object instead of separate non-secure and secure ?
Minor question - does this create _ota_client when secure one is used? (i.e. is it in memory at this point?)

code/espurna/ota.ino
Niek: Oops, missed that one. I have fixed it now.
We can also pass in the fingerprint char* directly but it wouldn't add anything.

code/espurna/ota.ino
Niek: I looked into the ``HTTPClient`` class, it has some nice extra checks and allows to verify response codes etc - but I think it will only increase the build size and in this case the ``Update`` class will verify the response anyway. But if you think it's better to use it I'll change the code.

code/espurna/ota.ino
Niek: Maybe it's even better to use a generic ``SUPPORT_SSL`` flag (or similar)? We can use that in MQTT and other modules as well.

code/espurna/ota.ino
Niek: Changed now!

code/espurna/ota.ino
Niek: I thought about it but it clutters the code a lot if you're using a separate _secure object: all subsequent ``_ota_client`` calls need to be surrounded with if-checks. But maybe this is not the nicest solution. Now that I think about it - not sure what happens when connecting to a HTTPS host first and to a HTTP one later. Do you have any suggestions?

mcspr: My bad not looking at this closer. Secure client will not fallback to basic one :(

Essentially, what my thinking is
- use separate methods for secure and non-secure update flow
- remove global client objects, create them inplace. We are not async, we expect them to live up until the function ends.

code/espurna/ota.ino
mcspr: Per pr - we no longer use async client, so why depend on it?

> We can use that in MQTT and other modules as well.

Yep, no reason not to enable all secure things at the same time (since they use the same library and we bundle it anyways).

code/espurna/ota.ino
mcspr: Then, maybe later. Keep current state of things, when parsing is done by us not by some HTTP lib.
But I'd double-check that updater does not needlessly write things to flash

code/espurna/ota.ino
mcspr: And mqtt /ota handler needs to be aware of that too. All async-mqtt-client callbacks are inside SYS context - we cannot use Core's sync clients inside of it

code/espurna/ota.ino
Niek: I have added SSL_ENABLED now - if it looks good I'll make the changes across the whole project in a separate PR

code/espurna/ota.ino
Niek: I just pushed some changes, the client objects are created inside the function now. I have a strange crash with SSL requests though (alloc issues in SSL) - can you check on your end as well?

code/espurna/config/dependencies.h
```diff
@@ -5,6 +5,10 @@
 // Configuration settings are in the general.h file
 //------------------------------------------------------------------------------
 
+#if !defined(ARDUINO_ESP8266_RELEASE_2_5_0) && !defined(ARDUINO_ESP8266_RELEASE_2_5_1) && !defined(ARDUINO_ESP8266_RELEASE_2_5_2)
+#define USING_AXTLS // do not use BearSSL on core < 2.5.0
+#endif
+
```
Niek: We need a way to check this in a better way - maybe ``ESP.getCoreVersion()``?

mcspr: That way bearssl is also off with git versions of the Core.
2.5.0 and up introduced https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_version.h
tag with proper comparison, but we again are hit by multi-Core support :/

In theory, one can depend on some behaviour / implementation instead. Kind of dangerous though, if that changes. For example, see:
https://github.com/xoseperez/espurna/compare/dev...mcspr:utils/fragmentation

mcspr: That way bearssl is also off with git versions of the Core.
2.5.0 and up introduced https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_version.h
tag with proper comparison, but we again are hit by multi-Core support :/

In theory, one can depend on some behaviour / implementation instead. Kind of dangerous though, if it changes. For example, see:
https://github.com/xoseperez/espurna/compare/dev...mcspr:utils/fragmentation

code/espurna/ota.ino
mcspr: Created dummy server from dockered nginx:1.15-alpine and letsencrypt cert chain. Fingerprinted or not, it either gets stuck on connection or errors out. Not even http request logged, unlike opening same .bin with browser.
```
ota https://<server>/firmware.bin
[014623] +OK
[014624] [OTA] Connecting to <server>:443 to download /firmware.bin
[014625] [OTA] Using fingerprint: <bla bla bla>
ERROR[0]: No Error
[015533] [OTA] Disconnected
```

mcspr: Something that I tried yesterday with HTTPClient & httpUpdate was requiring ~20seconds for connection to happen.

In general, I think we would want to avoid doing connection twice. HTTPClient, for example, wraps around that internally (which btw can be adapted here or straight up used as-is)

And  I did rewrite bunch of things to stop axtls thing from happening. I think one problematic part is shadowing _ota_client instead of having separate methods. Reviewing HTTPClient I'd noticed that it cleverly wraps around that too (see Traits classes)

mcspr: Remedy for 20 second wait was turning off active webserver (that was not ok...)
```
if (getSetting("webEnable", 1).toInt() == 0) return;
```
In webSetup(), wsSetup() and debugWebSetup()

At least now I have a reproducible "conflict" mentioned in the comments (which I did not remember having earlier :/). I wonder if async webserver polling method is broken somehow

mcspr:
For some of long-stack-trace-crashes: https://github.com/esp8266/Arduino/commit/d83eabe5b3078b95bddd50f3840234b116aa3063, esp8266/Arduino#6153, esp8266/Arduino#6156
Turns out bearssl client silently corrupted memory.
Also interesting advice from the linked issue - put WiFiClient object into heap (see unique_ptr example in HTTPClient https example). Haven't checked the exact size, but it may be important when we are using multiple levels of functions.

Still no clue what exactly webserver does differently from telnet server :/
Timeout problem (15000ms is the default) would also be explained by that, since we may never receive esp_schedule() in time to get back to the client. *Something* is blocking the process.

code/espurna/config/dependencies.h
Niek: Maybe we can use ``ARDUINO_ESP8266_GIT_VER`` instead? The type_traits stuff looks pretty rough (not to mention badly readable).

code/espurna/ota.ino
Niek: In my opinion it would be better to remove ESPAsyncTCP/ESPAsyncWebServer in the whole project - it's not maintained anymore and you'll see more and more compatibility issues with future core versions. Rewriting the telnet server is not too much work (I'm testing it now). For web support: the standard ESP8266WebServer is pretty good, you'll only need something like arduinoWebSockets for WS(S) support.

code/espurna/config/dependencies.h
mcspr: ARDUINO_ESP8266_GIT_VER has the same problem as RELEASE defines. BTW I think we are carried away. Does this work? Shouldn't it be global flag instead? IMO documentation would solve this better. If we sure this breaks horribly, static_assert(condition, "message") can also do this.

Regarding other branch, you'd say that this is a clearer intent?
https://github.com/xoseperez/espurna/blob/9ceca44770cc7772afc020ce7bad2d4cf321eadb/code/espurna/config/prototypes.h#L55-L66

https://en.cppreference.com/w/cpp/language/decltype
https://en.cppreference.com/w/cpp/types/enable_if
It is just an overload detection. We force compiler to pick specific 'detect()' method and store bool based on the one it found (either internal or external return type)
From my limited understanding, to make this shorter 'getHeapStats' function would need different return types.

code/espurna/ota.ino
mcspr: Likely thing :( Web server part is problematic for some parts though - alexa triggers header parsing bugs, some credentials stuff works differently... Which again puts the Core version as a dependency on some bug or feature. *(OT: we also can switch to espressif IDF 🙄)*
Websocket is not a huge problem, since it is both an [external lib and pretty widely used](https://github.com/Links2004/arduinoWebSockets)

mcspr:
Running latest GIT Core (esp8266/Arduino@dae64c4dae293172e469962a5bae8b66b204adc5) I no longer see the client hanging even when ESPAsyncWebServer is listening for connections.
Must've been the latest scheduled_functions changes (yield / no yield after loop etc.), not related to wifi client itself

code/espurna/ota.ino
mcspr: Another heads up regarding slightly getting distracted and dependencies in general.
What we can do instead of removing the module straight away, is implement another one side by side (just another file) and make them incompatible with each other. That way we can already use new one, and don't need to thing about dependencies in a strange way.

Niek:
Closing in favor of https://github.com/xoseperez/espurna/pull/1751

PR #1725, opened 2019-05-19
Configurable MQTT JSON topic

### LINKED ISSUE #1717: espurna MQTT use with thingsboard
hi dear;
i would like use espurna with thingsboard, but not get telemetry date from espurna. 
connection is ok. but cat't send telemetry data to thingsboard.
**settings:**
MQTT Root Topic : v1/devices/me/telemetry

**DEBUG**
[557691] [MQTT] Sending v1/devices/me/telemetry/data => {"temperature":26,"humidity":51,"time":"2019-05-15 23:31:30","mac":"18:FE:34:9D:59:75","host":"ESPURNA-9D5975","ip":"192.168.0.134","id":8758} (PID 1)

what is problem.? 
thanks.

mcspr:
resolve #1717 

- `MQTT_TOPIC_JSON` controls full topic
- `mqttTopicJson` setting
- update Thermometer to use new JSON topic rules

@ElderJoy something like this, then? instead of specifying name we specify full mqtt topic. will convert current name + '/data' to the topic setting + check if topic exists.


code/espurna/thermostat.ino
```diff
@@ -304,7 +305,7 @@ void _thermostatWebSocketOnSend(JsonObject& root) {
   root["thermostatVisible"] = 1;
   root[NAME_TEMP_RANGE_MIN] = _temp_range.min;
   root[NAME_TEMP_RANGE_MAX] = _temp_range.max;
-  root[NAME_REMOTE_SENSOR_NAME] = _thermostat.remote_sensor_name;
```
ElderJoy: The idea is that user in web interface set remote sensor name and this is enough to compose topic. And user can see remote sensor name i web interface. Removing this string woud affect web interface - user would not able to see current remote sensor name.

mcspr:
Right, but using combination of name (hostname of remote device?) and topic ("data")  we rely that remote device is using mqtt root topic such as `{hostname}`
Root can be multiple subtopics though, it is just a prefix e.g. if we are using something like `floor/room/DEVICE_NAME` instead of default `DEVICE_NAME`, we are leaking part of MQTT topic anyways.

ElderJoy:
Make sense. I agree with your point. It is much more flexible. It just need to add comment to this field to the web page that describes to user what does it mean and how to compose topic.

PR #1727, opened 2019-05-19
Separate device name for alexa integration

m-kozlowski:
Allow defining custom device name for alexa integration, other than hostname.
Still uses hostname as a fallback.


xoseperez:
This is OK. 
At some point we should be able to name each defined switch and that name will be the one to use also for alexa... But I guess we can merge this now.

Issue #1575, opened 2019-02-24
Magic Home LED Controller: white channel not independent anymore


hyteoo:
Using a RGBW controller
As of <s>1.13.4</s> 1.13.3, the white channel (channel #3) seems to be connected to the brightness setting.
Therefore the RGB and white channel can't be controlled independently anymore.

mcspr:
Hey,

I think I am missing something, but how exactly can you reproduce it? Yesterday I built a test image (i do not have magichome controller hw), but did not see anything unusual when doing `brightness N` or `channel 3 N` via terminal. I will also try with some leds attached asap, to actually see the change

*sorry for the delay 🙉*

hyteoo:
You can't see anything unusual in the interface, you have to connect a controller.
What happens is now the white channel is changed both by brightness and channel3.
Before brightness controlled RGB and channel3...the white.

mcspr:
OK. So does that mean it happens when "Use white channel" setting is off? I see one change that removed check for channels >= 3 when applying brightness:
https://github.com/xoseperez/espurna/commit/01802741f28933b23302508b1513290a363043bd#diff-1e2d5ef62603513c39da08ec25f60755L150
But that was already in 1.13.3

Only other change to lights after that was https://github.com/xoseperez/espurna/commit/19d0c5d17a576a22e8b7d38c1f09e3a14e41d6a5, but that was merely renaming 'shadow' into 'target' for transitions

hyteoo:
Yes, "Use white channel" setting is off.
The issue is old, since 1.13.3

mcspr:
Encoder thing was for the Xiaomi lamp with only WW channels without the RGB part, so to check both color and RGB channel number did not make sense there. 

IDK what is the right thing to do here. Change useWhite meaning to support ignoring white channel value (aka brightness) when controlling global brightness? Kind-of complicated.

mcspr:
Also, maybe related to #1292. MQTT brightness is not exposed unless we have RGB channels

copyrights:
@mcspr wrote in #1730
> Thanks! Brightness should be there.
> Next question is what to do in cases like #1575

Good question. I think in most chases the current behaviour will fit for most users need. But I also understand the need @hyteoo has.

The change would be quit simple. All you need to do is to reduce the loop in `_generateBrightness()`.
[...]

PR #1730, opened 2019-05-20
MQTT and HA brightness

### LINKED ISSUE #1292: Quinled brightness vs MQTT vs HA
First, thank you for making this great firmware. I want to use it with Quinled. I am happy to see it's supported.

There are small issue what I found:

- Brightness control is not added when I use auto discovery in Home Assistant, this part:

```
    brightness: true
    brightness_scale: 255
    brightness_command_topic: 'QUINLED-BED/brightness/set'
    brightness_state_topic: 'QUINLED-BED/brightness'
```

And I have one more question: what's the difference between setting brightness trough Brightness command vs. Channel #? Or I should leave both channels at maximum 255, then adjust brightness only with Brightness command?

screenshot 2018-10-20 at 14 46 46

Thank you,
Adrian

copyrights:
MQTT and HA brightness shall only be depended on LIGHT_PROVIDER != LIGHT_PROVIDER_NONE

This should fix #1292. I did a quick test with MQTT and web interface on `intermittech-quinled` target without observations. The homeassistant part is untested.

### DIFF
--- a/code/espurna/homeassistant.ino
+++ b/code/espurna/homeassistant.ino
@@ -56,7 +56,7 @@ void _haSendMagnitudes(const JsonObject& deviceConfig) {
             _haSendMagnitude(i, config);
             config["uniq_id"] = getIdentifier() + "_" + magnitudeTopic(magnitudeType(i)) + "_" + String(i);
             config["device"] = deviceConfig;
-            
+
             config.printTo(output);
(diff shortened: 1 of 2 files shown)


mcspr:
Thanks! Brightness should be there.
Next question is what to do in cases like #1575

Issue #1289, opened 2018-10-19
CR: Make WEB_REMOTE_DOMAIN configurable from admin console


cracautanu:
HTTP Api/Restful Api are not usable by custom web clients because of the CORS error. I tried ASP MVC 5 and Angular 6 but couldn't pass it. I also tried jsonp ajax get requests - they work to set a switch status, but it's really hard to get a response since they expect to set a callback and always throw an error. 
From my point of view this is a big limitation especially for those who don't want to get their hands dirty and rebuild/flash everything when taking a custom app online.

Playing with a restful extension for Chrome will work because the browser doesn't set the origin in the header...

mcspr:
\-> https://github.com/xoseperez/espurna/pull/1789

`webDomain` setting now handles this in addition to the WEB_REMOTE_DOMAIN define

Issue #1027, opened 2018-07-05
WEB_REMOTE_DOMAIN How to use


kiralikbeyin:
in general.h 

// Defining a WEB_REMOTE_DOMAIN will enable Cross-Origin Resource Sharing (CORS)
// so you will be able to login to this device from another domain. This will allow
// you to manage all ESPurna devices in your local network from a unique installation
// of the web UI. This installation could be in a local server (a Raspberry Pi, for instance)
// or in the Internet. Since the WebUI is just one compressed file with HTML, CSS and JS
// there are no special requirements. Any static web server will do (NGinx, Apache, Lighttpd,...).
// The only requirement is that the resource must be available under this domain.
#define WEB_REMOTE_DOMAIN           "http://192.168.0.43:4000"

https://github.com/xoseperez/espurna/wiki/WebInterface nothing found about WEB_REMOTE_DOMAIN

I am serving index.html from http://192.168.0.43:4000 -> Failed to load resource: the server responded with a status of 404 (Not Found) ->  http://192.168.0.43:4000/auth

What to do now?

mcspr:
I think you want to change `WEB_PORT` from 80 to 4000

`WEB_REMOTE_DOMAIN` is used for `Access-Control-Allow-Origin` header value

mcspr:
Hm. You'll need to proxy /auth and /ws paths then (404 error)
html assumes you are using it from the device and does js `connect("192.168.0.43")` which triggers other things based on that ip.

xoseperez:
This is a feature in beta-stage. It will let you host the web interface outside the device, saving some previous space and hopefully allowing to manage different devices from a unique location. For CORS to work a non-empty (or "*") domain name must be provided, this will be where the web interface will be hosted. 

You will then load the page from that domain and use the `connect` method to connect to the local device. If the domain matches that in WEB_REMOTE_DOMAIN the CORS rules will let you access and manage the device. There is a work-in-progress to add a discovery feature so you can scan your local network for ESPuna devices.

Please note that this is by no means a threat since all communications are done locally between your computer and your device, the remote domain only serves a static page with all the required contents.

kiralikbeyin:
WEB_REMOTE_DOMAIN sounds good but i didin like beta stage :)


> You will then load the page from that domain and use the connect method to connect to the local device. 

So we nede a static ip? Open port or dns like duckdns?

+I think current index.html need changes if there will be more than 1 device - show all devices?
[...]

PR #1733, opened 2019-05-22
Allow wildcard to enable discovery from browser using CORS

KimNyholm:
I believe wildcard as origin is a more useful default for CORS, enabling to discover espurna devices from a browser, using the /discover endpoint. Wildcard works for this endpoint, since it does not require credentials.

In the future [Make WEB_REMOTE_DOMAIN configurable from admin console](https://github.com/xoseperez/espurna/issues/1289) could be used to override the default.

The issue has also been discussed in [WEB_REMOTE_DOMAIN How to use](https://github.com/xoseperez/espurna/issues/1027)

An example of a browser discory applikation is [ESPurna Device Locator](https://github.com/KimNyholm/espurna-device-locator/)

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -531,7 +531,7 @@
 // there are no special requirements. Any static web server will do (NGinx, Apache, Lighttpd,...).
 // The only requirement is that the resource must be available under this domain.
 #ifndef WEB_REMOTE_DOMAIN
-#define WEB_REMOTE_DOMAIN           "http://espurna.io"
+#define WEB_REMOTE_DOMAIN           "*"
 #endif
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


AllWize:
This is working fine and it's an interesting solution. I would even say it will be great to have the whole webpage hosted externally (this was the reason for the original tests with CORS). But I wouldn't hardcode the WEB_REMOTE_DOMAIN to "*", instead I would allow this to be changed in config, maybe.

xoseperez:
This is working fine and it's an interesting solution. I would even say it will be great to have the whole webpage hosted externally (this was the reason for the original tests with CORS). But I wouldn't hardcode the WEB_REMOTE_DOMAIN to "*", instead I would allow this to be changed in config, maybe.

KimNyholm:
Agree, hardcoding is not the best solution. But it is a quick start, and a better value than the current.
Making it configurable is already described in [Make WEB_REMOTE_DOMAIN configurable from admin console](https://github.com/xoseperez/espurna/issues/1289).

Issue #1721, opened 2019-05-18
NoFUSS: X-ESP8266-DEVICE has 11 usable bytes, which is way too low for current hardware.h


m-kozlowski:
**Bug description**
Device identification header is set as follows:
```
        char buffer[20];
        snprintf_P(buffer, sizeof(buffer), PSTR("%s-%s"), APP_NAME, DEVICE);
```
APP_NAME is always set to ESPURNA (7 chars), add dash and null terminator and we have 11 chars left of 20.
There are 40 devices with longer IDs in current version of hardware.h, that will have header value truncated.
It doesn't break anything yet (closest to being ambiguous are SONOFF_T1_[1|2|3]CH, but names are truncated right after last digit) but it makes defining rules in versions.json harder.

**Expected behavior**
IMHO it's unnecessary to include "ESPURNA-" prefix in this case, but extending buffer length a little is also fine, I guess.

m-kozlowski:
fixed in 2cd7a8717aff1d277b4777d7f90a3e086ed9e619

PR #1737, opened 2019-05-25
Build date nofuss

m-kozlowski:
I counldn't find a proper way to upgrade between different builds of the same version using NoFuss
(defining rules to upgrade from 1.13.6-dev to 1.13.6-dev caused infinite upgrade loop).
Therfore I decided to include build date in X-ESP8266-VERSION header to allow more gradual upgrades.
While it's probably not the most elegant solution, it does the job and preserves backwards compatibility (by simply ommiting build date str in versions.json).
And since i had to increase buffer length anyway, this PR also solves xoseperez/espurna#1721

### DIFF
--- a/code/espurna/nofuss.ino
+++ b/code/espurna/nofuss.ino
@@ -54,7 +54,7 @@ void _nofussConfigure() {
 
     } else {
 
-        char buffer[20];
+        char buffer[32];
         snprintf_P(buffer, sizeof(buffer), PSTR("%s-%s"), APP_NAME, DEVICE);
 
         NoFUSSClient.setServer(nofussServer);
--- a/code/espurna/nofuss.ino
+++ b/code/espurna/nofuss.ino
@@ -58,12 +58,15 @@ void _nofussConfigure() {
         snprintf_P(buffer, sizeof(buffer), PSTR("%s-%s"), APP_NAME, DEVICE);
 
         NoFUSSClient.setServer(nofussServer);
-        NoFUSSClient.setDevice(buffer);
-        NoFUSSClient.setVersion(APP_VERSION);
-
         DEBUG_MSG_P(PSTR("[NOFUSS] Server : %s\n"), nofussServer.c_str());
+
+        NoFUSSClient.setDevice(buffer);
         DEBUG_MSG_P(PSTR("[NOFUSS] Dervice: %s\n"), buffer);
-        DEBUG_MSG_P(PSTR("[NOFUSS] Version: %s\n"), APP_VERSION);
+
+        snprintf_P(buffer, sizeof(buffer), PSTR("%s/%s"), APP_VERSION, buildTime().c_str());
+        NoFUSSClient.setVersion(buffer);
+        DEBUG_MSG_P(PSTR("[NOFUSS] Version: %s\n"), buffer);
+
         DEBUG_MSG_P(PSTR("[NOFUSS] Enabled\n"));
 
     }


mcspr:
The interesting part is that we partly have additional version information via APP_REVISION (from `git rev-parse --short HEAD`)
But it is only available when build.sh script is used - current release builds, nightly builder or when user runs it (and even knows about it). And it does not work with arduino-builder (arduino-cli / arduino ide)

m-kozlowski:
Yeah, i started with adding APP_REVISION in ota.py builds, but git revision numbers doesn't sort too well;)

mcspr:
Not directly, no. Using the same approach as `git-describe --tags` and referencing number of commits 
```
$ git describe --tags
1.13.5-121-g27474db9-dirty
$ git describe --tags --abbrev=0 origin/master
1.13.5
$ git rev-list --count 1.13.5..27474db9
121
```
I do agree the date requires less moving parts.

xoseperez:
There is yet another option. I just committed a change to the NoFUSS library to accept another filter, "build". This filter works slightly different from the others, since it will only filter out if it's defined both in the filter and the request (in the X-ESP8266-BUILD header) and their value is not the same.

This way you can also use revisions (if you want, dates will work well too) but there is no sorting, either you have the same version (and then it's not updated) or a different one.

mcspr:
On second thought, this exploits [`version_compare()`](https://www.php.net/manual/ru/function.version-compare.php) behaviour in dropping any text and treating things as a big number. And since we are using fixed-size ISO date string, it works.

Plus, there is no formalized way to recreate versions.json (?), so maybe locking by build and expecting just one entry per version per @xoseperez proposal

Format consideration though:
We are not required to parse __TIME__ and __DATE__, there is __TIMESTAMP__ string (`/` search `$ info cpp`)
In a similar fashion, nothing stops local user to modify APP_VERSION instead (like nightly builder, dropping git hash though)

m-kozlowski:
> On second thought, this exploits [`version_compare()`](https://www.php.net/manual/ru/function.version-compare.php) behaviour in dropping any text and treating things as a big number. And since we are using fixed-size ISO date string, it works.

Such behaviour is quite common (sort -v, python natsort...), so I'd say "exploits" is a little too harsh word;)
It does require builtTime() to return timestamp in non-retarded format (not the way __TIME__, __TIMESTAMP__ or any other preprocessor macro does), but I hope this format won't change as it's most readable of all options, both for automated scripts and human beings.

As for using __TIMESTAMP__:
- it's non mandatory preprocessor macro by current standards 
- output format is just as rubbish as __DATE__ and __TIME__ combination
- it doesn't contain build time, but time of last modification of *current* file
I think it would be best to convert __DATE__ and __TIME__ to unix timestamp during build stage using dark preprocessor voodoo like [this](https://stackoverflow.com/questions/11697820/how-to-use-date-and-time-predefined-macros-in-as-two-integers-then-stri/16472369#16472369), so we have universal value that requires minimum handling at runtime and occupy only 4 bytes of flash.

As for X-ESP8266-BUILD header: I like it. It gives little less flexibility than buildTime() solution, but in my usecase it does the job just as good and requires less tweaking in versions.json.

mcspr:
Not saying it is not common, just pointing out how it is parsed.
Timestamp macro is out then, haven't noticed that it uses file for reference.  Format is problematic too. But, it is always unique because we are using .ino format and preprocessor uses new temporary .cpp file each time.

Any reason why overriding APP_VERSION wouldn't work? If gcc fails us (and to avoid preprocessor magic), we can just use external tool to pass it into the build. Either via `date -u --iso-8601=seconds`, some other format or something else completely.
For example, creating version.h.in -> version.h script or wrapping version.h definitions in ifndefs

m-kozlowski:
> Any reason why overriding APP_VERSION wouldn't work?

It would. And personally I like it.  Either passing through `PLATFORMIO_BUILD_FLAGS` or version.h regeneration is simple and effective. But doesn't it binds us to specific environment? It's not that I'm against such approach. I genuinely don't know if there are anyone who cares about Windows builds for example (if that's even possible) and what to expect from such build environment. But since PlatformIO is multiplatform, I tend to stick to universal solutions. Maybe needlessly;)

mcspr:
Each environment can run things, but platformio's automation is much more straightforward.

Little external python script can keep current version info data inside of it (MAJOR,MINOR,PATCH,SUFFIX,AUTHOR,WEB,EMAIL). Depending on some parameters (environment, dotfile), generate either git-describe version, add user-provided string or put default version data like now.
Pre-commit script can keep things up-to-date outside builds. Platformio can run it each build. Arduino IDE needs some hacking of local platform files to make things work.
Should be pretty straightforward to add.

Anyways, PR, revert 0e0ca16b7e1fab336df320a78fae1cbe08c38675 and keep buffer update?
To properly fit name we can init charbuffer like this: `buffer[strlen(APP_NAME) + strlen(DEVICE) + 1]`

m-kozlowski:
commit 2cd7a8717aff1d277b4777d7f90a3e086ed9e619 made this PR obsolete

PR #1740, opened 2019-05-27
Terminal: heap fragmentation stat

mcspr:
An example of using new Core functionality, without preprocessor magic (using template magic!)
*edit: need verification if that would work with templated classes, inheritance etc.* 

However, it is more of a discussion of how to present it.
- should there be additional topic like `<root>/heapfrag`?
- should `heap` value be replaced with the real one, minus fragmentation? mqtt /heap, terminal, influxdb etc.


mcspr:
Small example of "how things can be"
Following C++ stdlib headers (specifically, is_destructible -> https://github.com/gcc-mirror/gcc/blob/9bcca88e24e64d4e23636aafa3404088b13bcb0e/libstdc%2B%2B-v3/include/std/type_traits#L608), to space out nested arguments:
```c++
struct trait_has_getHeapStats {
    template<typename T, typename = decltype(declval<T&>().getHeapStats())>
      static std::true_type detect(int);

    template<typename>
      static std::false_type detect(...);
};

template <typename T>
struct type_has_getHeapStats : public trait_has_getHeapStats {
    using type = decltype(detect<T>(0));
};

template <typename T>
struct has_getHeapStats : public type_has_getHeapStats<T>::type
{ };
```

PR #1746, opened 2019-05-31
update wrong pinout on ag-l4

zerog2k:


mcspr:
tnx! file size grew +140% though, but merging as-is to test if imgbot is still working.

PR #1751, opened 2019-06-02
Experimental support of HTTPUpdate for OTA

mcspr:
- select OTA_CLIENT (ESPAsyncTCP or HTTPUpdate)
- select SSL_CLIENT (WiFiClientSecure variant)
- partial support for CA certs with SSL_CLIENT_BEARSSL (example letsencrypt isrg x1 root as default. might not work 50% of the time)

Configuration might need to be OTA specific though, `otaSslCheck`?

@Niek This is what I meant by separate implementations. I think this builds 🌊 
ArduinoOTA and OTA client are split into separate modules, Core HTTP updater added in a similar manner. Both axtls and bearssl should be supported. Only fingerprint support with 2.4.2, because travis envs do not use 2.5.. yet (and at this point, idk what to do with it. Core git broke spiffs symbols.... agrh)

CA support is trickier. The most robust way is to allow multiple of them (as common CA storage works on general-purpose oses), but rn it is only a single one. Either some personal CA entity and that surely will work, or we hope that LE cert is properly signed (because I do see that some of my certs are signed with DST root instead).

*edit: This is more of a reference implementation and a dumping ground for some nice things found on the way, which will be merged asap (URL class, ArduinoOTA support flag...) before this*


mcspr:
For example, X1 root from LE is ~1.4KB when stored as .der
With spiffs it could be pretty straightforward thing to manage - just use [CertStore](https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/bearssl-client-secure-class.html#certificate-stores)
One simple option here is to just reserve a single flash sector for it and give some API to overwrite it
(4KB, up to 2 certs! aka root+intermediate. plus some storage left over)

Niek:
Very nice, great work @mcspr! I will be leaving some comments on the code, but this looks much nicer split up this my PR - I'll be closing that one now in favor of this.

code/espurna/config/general.h
```diff
@@ -626,14 +626,48 @@
...
+                                                         // const char _ota_client_http_update_ca[] PROGMEM = "...";
+                                                         // By default, use Letsencrypt ISRG Root X1
+#endif
+
+// -----------------------------------------------------------------------------
+// SSL Client
+// -----------------------------------------------------------------------------
+
+#ifndef SSL_CLIENT
+#define SSL_CLIENT                          SSL_CLIENT_NONE     // What variant of WiFiClient to use (no SSL support by default)
```
Niek: I think it would be good to mention the option here, so: ``SSL_CLIENT_NONE``, ``SSL_CLIENT_AXTLS`` and ``SSL_CLIENT_BEARSSL``. Also maybe do some auto detection based on the core version?

code/espurna/config/prototypes.h
```diff
@@ -128,7 +129,33 @@ void i2c_read_buffer(uint8_t address, uint8_t * buffer, size_t len);
...
+#endif
+
+#if SSL_CLIENT != SSL_CLIENT_NONE
+
+    #include <WiFiClientSecure.h>
+
+    // TODO: add DigiCert CA for github?
+    #if SSL_CLIENT_BEARSSL && OTA_SSL_CLIENT_INCLUDE_CA
+    #include "static/ota_ssl_client_ca.h"
+    #else
```
Niek: ``#else if SSL_CLIENT_BEARSSL``?

code/espurna/ota_asynctcp.ino
```diff
@@ -0,0 +1,222 @@
...
+    delay(0);
+
+}
+
+void _otaClientOnConnect(void *arg, AsyncClient *client) {
+
+    #if ASYNC_TCP_SSL_ENABLED
+        if (443 == _ota_url->port) {
+            uint8_t fp[20] = {0};
+            sslFingerPrintArray(getSetting("otafp", OTA_GITHUB_FP).c_str(), fp);
```
Niek: ``OTA_FINGERPRINT``?

code/espurna/ota_httpupdate.ino
```diff
@@ -0,0 +1,201 @@
...
+        new BearSSL::WiFiClientSecure
+    );
+    
+    if (check == SSL_CLIENT_CHECK_NONE) {
+        DEBUG_MSG_P(PSTR("[OTA] !!! Connection will not be validated !!!\n"));
+        client->setInsecure();
+    }
+
+    if (check == SSL_CLIENT_CHECK_FINGERPRINT) {
+        String fp_string = getSetting("otafp", OTA_GITHUB_FP);
```
Niek: ``OTA_FINGERPRINT``?

code/espurna/ota_httpupdate.ino
```diff
@@ -0,0 +1,201 @@
...
+
+#if SSL_CLIENT == SSL_CLIENT_AXTLS
+
+void _otaClientFromHttps(const String& url) {
+
+    int check = getSetting("sslClientCheck", SSL_CLIENT_CHECK_FINGERPRINT).toInt();
+
+    String fp_string = "";
+    if (check == SSL_CLIENT_CHECK_FINGERPRINT) {
+        fp_string = getSetting("otafp", OTA_GITHUB_FP);
```
Niek: ``OTA_FINGERPRINT``?

code/espurna/ota_httpupdate.ino
```diff
@@ -0,0 +1,201 @@
...
+#endif // TERMINAL_SUPPORT
+
+// -----------------------------------------------------------------------------
+
+void otaClientSetup() {
+
+    #if TERMINAL_SUPPORT
+        _otaClientInitCommands();
+    #endif
+
```
Niek: I'm missing the MQTT OTA support here, is that on purpose? Maybe it's better to make the terminal+MQTT code separate of the used OTA module?

code/espurna/config/general.h
mcspr: Autodetection? Cipher problem that was kind of recent one (stack thunk size increase) and can be mitigated locally by using hand-made cipher list. And that affects 2.5.0-2.5.2. It kind of tricky to define what works, so maybe it is better to leave this alone.
I will add a comment about possible clients options + Core from which they work / become deprecated etc

code/espurna/config/prototypes.h
mcspr: I think I wanted to add something here using .cpp files and weaken the progmem char symbol, but that did not quite work. This probably can even be just a check for include_ca flag, progmem'ed array will be discarded anyways as unused variable.

code/espurna/ota_asynctcp.ino
mcspr: Yep. Thanks! Will fix in the morning

code/espurna/ota_httpupdate.ino
mcspr: I did forget that.
Code can be common, but it would need to be a loop() callback so that wificlient works.

Niek:
Are you planning to merge this soon @mcspr? I will base the MQTT changes I'm working on on this PR (e.g. ``SSL_CLIENT``).

mcspr:
I thought to clean up 1.13.6 first.
Still some things waiting (#1705 , #1769 and others)

@xoseperez did specify that we use semantic versioning, but I think we need to release more versions for that to apply. However, since we don't break existing functionality and only add optional features, maybe it still works for this PR.

Niek:
Maybe it's an idea to make a "SSL improvements" branch? That makes it at least a bit easier to base other PRs on.

mcspr:
It can. I'd sleep on it, but I think just adding it to dev marked as experimental would be less of a hassle. There could be a separate module/feature line in `info` banner to denote that. Maybe even in the same line, with some marker.

And some minor things:
- `ssl...` settings should be namespaced per module? `mqttSsl...`, `otaSsl...` ...
- isn't it TLS now? SECURE_...?
- something bugging me about URL class, need to check how memory is allocated when chunks are substring'ed

mcspr:
Can merge this finally.

So the one thing I have forgot is to support basic urls when ssl client is built. Turns out https://github.com/esp8266/Arduino/pull/4980 marked all but update(WiFiClient, ...) methods as deprecated.

Both HTTPCLIENT_1_1_COMPATIBLE and HTTPUPDATE_1_2_COMPATIBLE look more suitable for internal Core dependency, not something supporting both sides of deprecation (when there were none of those methods present). That and different update() signatures per fingerprinting, it is kind of a pain to support all of the methods. Hence combining deprecated axTLS and update(url, fp) methods together
Bearssl is upped to >=2.5.0, since we don't want to support basic update(...) and want to directly use update(WiFiClient, ...)

Niek:
Looks good! Once it's merged I will change the MQTT code to look at the ``SECURE_CLIENT_CHECK`` defines.

code/espurna/config/general.h
```diff
@@ -621,19 +621,74 @@
...
+                                                                      // SECURE_CLIENT_BEARSSL - BearSSL client secure support (starting with 2.5.0, TLS 1.2)
+                                                                      //
+                                                                      // axTLS marked for derecation since 2.4.2 and **will** be removed in the future
+#endif
+
+// Support Maximum Fragment Length Negotiation TLS extension
+// "...negotiate a smaller maximum fragment length due to memory limitations or bandwidth limitations."
+// - https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/bearssl-client-secure-class.html#mfln-or-maximum-fragment-length-negotiation-saving-ram
+// - https://tools.ietf.org/html/rfc6066#section-4
+#ifndef SECURE_CLIENT_MFLN
```
Niek: Just one thing I thought about: this should mention that it's only working with ``SECURE_CLIENT`` set to ``SECURE_CLIENT_BEARSSL``

Niek: Also, the value should be between 512 and 16384:
https://github.com/esp8266/Arduino/blob/c18b402c31dc27ffc4b8817035ad8a44670c1b2e/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L186-L187

code/espurna/ota_httpupdate.ino
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    // TODO: provide terminal command for probing?
+    // TODO: RX and TX buffer sizes must be equal?
+    uint16_t requested_mfln = getSetting("otaScMFLN", OTA_SECURE_CLIENT_MFLN).toInt();
+    if (requested_mfln) {
+        URL _url(url);
+        bool supported = client->probeMaxFragmentLength(_url.host.c_str(), _url.port, requested_mfln);
+        DEBUG_MSG_P(PSTR("[OTA] MFLN buffer size %u supported: %s\n"), requested_mfln, supported ? "YES" : "NO");
+        if (!supported) return;
```
Niek: Instead of returning, shouldn't this code be:
```c++
if (supported) {
  client->setBufferSizes(requested_mfln, requested_mfln);
}
```
Because now it never calls ``_otaClientRunUpdater()`` if MFLN is not supported.

mcspr: But why ignore `requested_mfln`? Default is off, so it either works or not.

>  RX and TX buffer sizes must be equal?

Looking at this now, does probing check how much server can send or receive? Do buffers need symmetry?

code/espurna/config/general.h
mcspr: Come to think of it, recv and send values can be different. Default sets it as high as possible for receiving end (since server can just dump a lot of data and expect it to be parsed):
https://github.com/esp8266/Arduino/blob/c18b402c31dc27ffc4b8817035ad8a44670c1b2e/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L83

But, MFLN *only* supported values are 512, 1024, 2048 and 4096. 16k is maximum record size for TLS in general.
https://tools.ietf.org/html/rfc6066#page-8
https://tools.ietf.org/html/rfc8449#page-3

code/espurna/ota_httpupdate.ino
Niek: The question is if the function should fail if MFLN is not supported by the server. I think it's more of a "nice to have" thing, and the function should just proceed if MFLN is not supported?

code/espurna/config/general.h
Niek: Ah, good to know - I see the same is documented here:
https://github.com/esp8266/Arduino/blob/c18b402c31dc27ffc4b8817035ad8a44670c1b2e/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L1359-L1361

code/espurna/ota_httpupdate.ino
mcspr: Maybe I am overthinking it. It would be nice to not blow up memory though?
Or client can have a "nice" mode, probing all of the possible buffer sizes and settling on some and saving it as a setting.

code/espurna/config/general.h
mcspr: And down below the exact protocol byte sequences:
https://github.com/esp8266/Arduino/blob/c18b402c31dc27ffc4b8817035ad8a44670c1b2e/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L1399-L1405

It is actually mentioned in the documentation proper! Missed that (since I looked as setBufferSizes)
> probeMaxFragmentLenght() ... (len must be a power of two from 512 to 4096, per the specification)

code/espurna/ota_httpupdate.ino
Niek: Actually probing by default might be a good idea - but each probe is of course some extra network traffic. For long-lived connections like MQTT I think it's definitely worth it to do by default.

The question is if TLS implementations (server-side) have limitations on the probe length. I'd think it MFLN is supported it will always support 512 and up, and not e.g. 4k only? So maybe 1 probe is enough? But hard to say without diving in OpenSSL/gnuTLS/OTP/etc code.

code/espurna/ota_asynctcp.ino
```diff
@@ -0,0 +1,220 @@
...
+    delay(0);
+
+}
+
+void _otaClientOnConnect(void *arg, AsyncClient *client) {
+
+    #if ASYNC_TCP_SSL_ENABLED
+        if (443 == _ota_url->port) {
+            uint8_t fp[20] = {0};
+            sslFingerPrintArray(getSetting("otafp", OTA_FINGERPRINT).c_str(), fp);
```
Niek: Should have a check for ``otaScCheck`` here as well, and skip the fingerprinting if it's ``SECURE_CLIENT_CHECK_NONE``.

code/espurna/terminal.ino
```diff
@@ -204,7 +204,28 @@ void _terminalInitCommand() {
...
+    #endif
+
+    #if SECURE_CLIENT == SECURE_CLIENT_BEARSSL
+        terminalRegisterCommand(F("MFLN.PROBE"), [](Embedis* e) {
+            if (e->argc != 3) {
+                terminalError(F("[url] [value]"));
+                return;
+            }
+
+            URL _url(e->argv[1]);
```
mcspr: Should it be `<host> <port> <value>` instead? URL needs to be a bit more clever (or something like external lib instead of bare-minimum parser) to parse things like `example.com:8443`

code/espurna/ota_httpupdate.ino
```diff
@@ -0,0 +1,251 @@
...
+}
+
+#endif // SECURE_CLIENT_BEARSSL
+
+
+#if SECURE_CLIENT == SECURE_CLIENT_AXTLS
+
+void _otaClientFromHttps(const String& url) {
+
+    const int check = getSetting("otaScCheck", SECURE_CLIENT_CHECK_FINGERPRINT).toInt();
```
mcspr: ref #1829 comment about the defaults... How about
general.h
```
#define SECURE_CLIENT_CHECK SECURE_CLIENT_CHECK_FINGERPRINT
...
#define OTA_SECURE_CLIENT_CHECK SECURE_CLIENT_CHECK // chained global default
```
so this uses the `SECURE_CLIENT_CHECK_FINGERPRINT` instead of trying to think about possible default values.

Niek: Yes, I like this approach - will update my PR as well

Issue #1739, opened 2019-05-26
Missing xtensa-lx106-elf-g++


wsw70:
I am trying to build a `Dockerfile` to better automatize the build of ESPurna binaries and I have a problem with `xtensa-lx106-elf-g++`.

The `Dockerfile` is simple: it downloads `bash`, `curl`, `git` and  `g++`, and `platformio` ; then clones the repository and informs that when teh container is ran, `pio run -e ...` will be launched.

```
FROM python:2-alpine
RUN apk add --no-cache bash curl git g++
RUN python -c "$(curl -fsSL https://raw.githubusercontent.com/platformio/platformio/develop/scripts/get-platformio.py)"
RUN git clone https://github.com/xoseperez/espurna.git
WORKDIR /espurna/code
CMD pio run -e itead-sonoff-basic -e itead-sonoff-rfbridge -e itead-sonoff-touch -e aithinker-ai-light
```

The build is successful, but when running the container I get:

```
root@srv /e/d/espurna# docker run -e PLATFORMIO_BUILD_FLAGS="'-DUSE_CUSTOM_H'" -v /etc/docker/bin:/espurna/code/.pioenv -v /etc/docker/custom.h:/espurna/code/espurna/config/custom.h espurna
********************************************************************************
If you like PlatformIO, please:
- follow us on Twitter to stay up-to-date on the latest project news > https://twitter.com/PlatformIO_Org
- star it on GitHub > https://github.com/platformio/platformio
- try PlatformIO IDE for IoT development > https://platformio.org/platformio-ide
- support us with PlatformIO Plus > https://pioplus.com
********************************************************************************


Processing itead-sonoff-basic (framework: arduino; platform: espressif8266@1.5.0; board: esp01_1m)
--------------------------------------------------------------------------------
LibraryManager: Installing id=64 @ 5.13.4
Downloading...

(... libraries are being cloned ...) 

Unpacking
PackageManager: Installing framework-arduinoespressif8266 @ ~1.20300.1
Downloading
Downloading
Unpacking
Unpacking
CorePackageManager: Installing tool-scons @ ~2.20501.7
Downloading
Downloading
Unpacking
Unpacking
tool-scons @ 2.20501.7 has been successfully installed!
Verbose mode can be enabled via `-v, --verbose` option
CONFIGURATION: https://docs.platformio.org/page/boards/espressif8266/esp01_1m.html
PLATFORM: Espressif 8266 > Espressif Generic ESP8266 ESP-01 1M
[...]

PR #1754, opened 2019-06-02
Docker build

copyrights:
# Docker build image

This PR contains a docker file for an ESPurna build environment.

Two volumes can be used. 
* `/espurna` with ESPurna source code.
* `/firmware` target directory for complied firmware files.

## Build

```bash
docker build -t espurna-build .
```

## Examples

The simples example will build all firmware files from dev branch to /tmp/firmware.

```bash
docker run --rm -it -v /tmp/firmware/:/firmware espurna-build
```

This example will only build firmware for environment `intermittech-quinled` from local files in `/home/user/espurna`

```bash
docker run --rm -it -v /tmp/firmware/:/firmware -v /home/user/espurna/:/espurna espurna-build intermittech-quinled
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,41 @@
+FROM debian:buster-slim
+
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install -y python2-minimal \
+    python-pip \
+    nodejs \
+    npm \
+    git \
+    libc-bin \
+    bash \
+    && npm install -g npm \
+    && pip install platformio
+
+RUN groupadd --gid 1000 worker && \
+    useradd \
+    --uid 1000 \
+    --gid worker \
+    --shell /bin/bash \
+    --create-home worker
+
+RUN mkdir -p /espurna && \
+    mkdir -p /firmware && \
+    chown -R worker:worker /espurna && \
+    chown -R worker:worker /firmware
+
+USER worker
+
+RUN git clone -b dev https://github.com/xoseperez/espurna.git /espurna && \
+    cd /espurna/code && \
+    npm install --only=dev && \
+    platformio run --target clean
+
+VOLUME ["/espurna", "/firmware"]
+
+WORKDIR /espurna/code
+
+COPY docker-entrypoint.sh /
+ENTRYPOINT ["./build.sh", "-d", "/firmware"]
+CMD []
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,28 @@
+# Docker build image
+
+This directory contains a docker file for an ESPurna build environment.
+
+Two volumes can be used. 
+* `/espurna` with ESPurna source code.
+* `/firmware` target directory for complied firmware files.
+
+## Build
+
+```bash
+docker build -t espruna-build .
+```
+
+## Examples
+
+The simples example will build all firmware files from dev branch to /tmp/firmware.
+
+```bash
+docker run --rm -it -v /tmp/firmware/:/firmware espruna-build
+```
+
+
+This example will only build firmware for environment `intermittech-quinled` from local files in `/home/user/espurna`
+
+```bash
+docker run --rm -it -v /tmp/firmware/:/firmware -v /home/user/espurna/:/espurna espruna-build intermittech-quinled
+```
--- a/docker/README.md
+++ b/docker/README.md
@@ -9,7 +9,7 @@ Two volumes can be used.
 ## Build
 
 ```bash
-docker build -t espruna-build .
+docker build -t espurna-build .
 ```
 
 ## Examples
@@ -17,12 +17,12 @@ docker build -t espruna-build .
 The simples example will build all firmware files from dev branch to /tmp/firmware.
 
 ```bash
-docker run --rm -it -v /tmp/firmware/:/firmware espruna-build
+docker run --rm -it -v /tmp/firmware/:/firmware espurna-build
 ```
 
 
 This example will only build firmware for environment `intermittech-quinled` from local files in `/home/user/espurna`
 
 ```bash
-docker run --rm -it -v /tmp/firmware/:/firmware -v /home/user/espurna/:/espurna espruna-build intermittech-quinled
+docker run --rm -it -v /tmp/firmware/:/firmware -v /home/user/espurna/:/espurna espurna-build intermittech-quinled
 ```


copyrights:
I did version pinning to pass the checks, but I'm not sure if this is a good idea. It will need more maintenance with only little benefits.

mcspr:
Pins can be relaxed imo. We don't depend on the specific version of node, but gulp. PlatformIO I would keep at 3.6.* instead though.

What would the flow be to just update internal sources? docker build --target N-step ?
Maybe just always force to use external directory? (or docker storage volume, for example)

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    && apt-get install -y --no-install-recommends \
+    python2-minimal=2.7.16-1 \
+    python-pip=18.1-5 \
+    python-setuptools=40.8.0-1 \
+    ca-certificates=20190110 \
+    nodejs=10.15.2~dfsg-2 \
+    npm=5.8.0+ds6-4 \
+    git=1:2.20.1-2 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
```
mcspr: Codacy is slacking in regard of updating linter config files (or we should have it here instead)
Per https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
> Official Debian and Ubuntu images [automatically run](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105) `apt-get clean`, so explicit invocation is not required.

copyrights:
> What would the flow be to just update internal sources? docker build --target N-step ?
> Maybe just always force to use external directory? (or docker storage volume, for example)

Main purpose for the internal source is to install dependencies. Secondary to just build the latest `dev`. But maybe it would be better to use `COPY` instead of `git clone`. This should trigger the build process on changes. I will try.

docker/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Build
+
+```bash
+docker build -t espurna-build -f Dockerfile ..
+```
+
+## Examples
+
+The simples example will build all firmware files from dev branch to /tmp/firmware.
```
mcspr: *This simple?

docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    --gid worker \	
+    --shell /bin/bash \
+    --create-home worker
+
+RUN mkdir -p /espurna && \
+    mkdir -p /firmware && \
+    chown -R worker:worker /espurna && \
+    chown -R worker:worker /firmware
+
+COPY --chown=worker:worker . /espurna
```
mcspr: .dockerignore is needed then (in code/ or at the top?)
we need to ignore platformio build directories
code/.pio
code/.piolibdeps
code/.pioenvs

mcspr:
Also, ping @wsw70 from https://github.com/xoseperez/espurna/issues/1739

wsw70:
I took a different way in my automated builds (which are still in dev phase, I need to clean up the `Dockerfile`) - building the `dev` and `master` branches.

-  the `master` is currently disabled because the build crashes due to an incompatibility of a lib recently upgraded which name escapes me right now. 
- I build only the firmware which is interesting for me but this is something I would like to configure - how to do it is to be assessed because there is a loop to code (over the platforms) which cannot be done directly in the `Dockerfile` but via some bash as part of the `RUN`.

My current `Dockerfile` is below.

```
FROM python:2
RUN apt update && apt install bash curl git g++
RUN pip install platformio
RUN git clone https://github.com/xoseperez/espurna.git
WORKDIR /espurna/code

# # build the master version
# RUN git checkout master
# COPY custom.h espurna/config/custom.h
# ENV PLATFORMIO_BUILD_FLAGS '-DUSE_CUSTOM_H'
# RUN pio run -e itead-sonoff-basic -e itead-sonoff-rfbridge -e itead-sonoff-touch -e aithinker-ai-light
# RUN mkdir binaries-master
# RUN cp .pioenv/itead-sonoff-basic/firmware.bin binaries-master/itead-sonoff-basic.bin
# RUN cp .pioenv/itead-sonoff-rfbridge/firmware.bin binaries-master/itead-sonoff-rfbridge.bin
# RUN cp .pioenv/itead-sonoff-touch/firmware.bin binaries-master/itead-sonoff-touch.bin
# RUN cp .pioenv/aithinker-ai-light/firmware.bin binaries-master/aithinker-ai-light.bin

# build the dev version
RUN git checkout dev
COPY custom.h espurna/config/custom.h
ENV PLATFORMIO_BUILD_FLAGS '-DUSE_CUSTOM_H'
RUN pio run -e itead-sonoff-basic -e itead-sonoff-rfbridge -e itead-sonoff-touch -e aithinker-ai-light
RUN mkdir binaries-dev && touch binaries-dev/.touch
RUN cp .pioenvs/itead-sonoff-basic/firmware.bin binaries-dev/itead-sonoff-basic.bin
RUN cp .pioenvs/itead-sonoff-rfbridge/firmware.bin binaries-dev/itead-sonoff-rfbridge.bin
RUN cp .pioenvs/itead-sonoff-touch/firmware.bin binaries-dev/itead-sonoff-touch.bin
RUN cp .pioenvs/aithinker-ai-light/firmware.bin binaries-dev/aithinker-ai-light.bin

#CMD cp -r binaries-master /binaries && cp -r binaries-dev /binaries
CMD cp -r binaries-dev /binaries
```

copyrights:
> There is a small issue when user has different UID than 1000. We need to both remap docker-run uid and append HOME env for the npm to cache things, or it will crash when trying to write /.npm
> i.e. something like this: ... -e "HOME=/tmp" -u $(id -u ${USER}):$(id -g ${USER}) ...

I make UID and GID configurable. The home directory is inside the image on purpose so that platformio is already ready to build. 
The lib_deps and npm modules are more of an issue to me. If you use your local code they get downloaded again. I thought about copying them after initialisation to a location outside the volume. Then check if e.g. .piolibdeps exists and is not empty otherwise copy them from inside the image.

copyrights:
@wsw70 thanks for sharing your current status. I though about using `FROM python:2` too, but as nodejs is needed for complete build I chose debian.

wsw70:
@copyrights 
> as nodejs is needed for complete build I chose debian.

I am not really knowledgeable about platformio compilation: why is nodejs needed for a complete build? I use the platformio from python and the firmware compile fine. Is there something I am missing in my compilation?

copyrights:
@wsw70 If you change some web files gulp is used to shrinks them to gz.h files.

mcspr:
Which is why I wonder if there is any need to use internal source and keep libraries at all. And we also exchange small git history for docker images :/
Either that, or we need some really persistent storage (storage volume, shared between builders or current work dir)

There could be another volume exporting toolchain, located at ~/.platformio
~/.platformio/platforms/... downloads active platform. which can be different per environment
~/.platformio/packages/... keeps Core installation, xtensa toolchain, esptool
~/.platformio/.cache/... keeps library archives and network data caches (note that it might not work properly, because some of the current bugs and tz difference between container and host)

The part with COPY, pio installs and npm can be removed in favour of customizable entrypoint, either building or installing / updating dependencies.

PR #1765, opened 2019-06-08
Disable Serial log for hardware using TX or RX pins

mcspr:


oscarsan1:
I think that serial support must be dissabled in most of devices in order to use serial pins to do something, connecting sensors, ... In my case with Sonoff Basic. 

mcspr:
We define pins as preprocessor tokens, so it would be possible to gather them in one place and compare to both TX and RX. When they match, such it can set `DEBUG_SERIAL_SUPPORT` to 0

Issue #1747, opened 2019-05-31
MQTT -> Publish button status


skyynet:
I publish from my Sonoff Dual R2 to my Mosquitto Broker on my Synology.

When I read /home/relay/0 any time I get the current state (0 or 1). This also applies to relay/1.

When I read /home/button/0 I get no result. I see its status is published but would like to read it any time using phpMQTT.

So generally I'd like to see button/0 and button/1 behave like relay/0 and relay/1.

Is this something I can configure somewhere, a bug or a feature request?

In hardware.h I defined:

```
#elif defined(ITEAD_SONOFF_DUAL_R2)

    #define MANUFACTURER        "ITEAD"
    #define DEVICE              "SONOFF_DUAL_R2"

    // Buttons
    #define BUTTON1_PIN         0       // Button 0 on header
    #define BUTTON2_PIN         9       // Button 1 on header
    #define BUTTON3_PIN         10      // Physical button
    #define BUTTON1_RELAY       0
    #define BUTTON2_RELAY       0
    #define BUTTON3_RELAY       1
    #define BUTTON1_MODE        BUTTON_SWITCH | BUTTON_SET_PULLUP | BUTTON_DEFAULT_HIGH
    #define BUTTON2_MODE        BUTTON_SWITCH | BUTTON_SET_PULLUP | BUTTON_DEFAULT_HIGH
    #define BUTTON3_MODE        BUTTON_PUSHBUTTON | BUTTON_DEFAULT_HIGH

    // Relays
    #define RELAY1_PIN          12
    #define RELAY2_PIN          5
    #define RELAY1_TYPE         RELAY_TYPE_NORMAL
    #define RELAY2_TYPE         RELAY_TYPE_NORMAL

    // LEDs
    #define LED1_PIN            13
    #define LED1_PIN_INVERSE    1
```

mcspr:
per https://github.com/xoseperez/espurna/issues/608 i think buttons need to be switches. only then we can reliably say about the current state and enable retained flag for mqtt messages

(also a reminder about? https://github.com/xoseperez/espurna/issues/619)

skyynet:
Thanks for the hint. Unfortunately this doesn't help. Subscribing to relay/0 or 1 gives me immediately status "0" indicating the relay is off. Subscribing to button/0 or 1 gives me nothing. When connecting the according pin with gnd it gives me 2,3 or 4.

I would really like to get state "0" for open buttons and something else for closed ones.

[...]

PR #1772, opened 2019-06-11
BUTTON_MQTT_MODE setting

mcspr:
ref #1747
 
@skyynet I though a bit about eventsensor for this, but it might be a bad match due to hardware limitations and noise (i think you pretty much required to have hardware filtering for the digital pin signal)

Instead, what about an option to send button pressed state instead of event, bypassing event mapping and directly giving user the button state?

### DIFF
--- a/code/espurna/button.ino
+++ b/code/espurna/button.ino
@@ -25,11 +25,11 @@ std::vector<button_t> _buttons;
 
 #if MQTT_SUPPORT
 
-void buttonMQTT(unsigned char id, uint8_t event) {
+void buttonMQTT(unsigned char id, uint8_t event, bool optional_retain = false) {
     if (id >= _buttons.size()) return;
     char payload[2];
     itoa(event, payload, 10);
-    mqttSend(MQTT_TOPIC_BUTTON, id, payload, false, false); // 1st bool = force, 2nd = retain
+    mqttSend(MQTT_TOPIC_BUTTON, id, payload, false, optional_retain ? mqttRetain() : false); // 1st bool = force, 2nd = retain
 }
 
 #endif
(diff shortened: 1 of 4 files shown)


mcspr:
Per definitions configuration, button limit is 8 (=> uint8_t as mask makes sense)
And some useful things for container values:
- `uint8_t value{DEFINITION};` instead of `uint8_t value = DEFINITION;` will trigger compilation error instead of just a mere warning.
- Based on v2 limits definition, BUTTONS_MAX is added. One wordy solution to this is to use container type as reference: `std::numeric_limits<decltype(_button_mqtt_mask_events)>::digits` would result in 8 for uint8_t

davebuk:
Is there currently any way I can use the mqtt reporting of a button state whose `BUTTON_ACTION` is set to `NONE`? I have a relay bound to switch#0 and use `PRESSED` as TOGGLE and **also** `LONG CLICK` for OFF. All others are set to `NONE` I use the `LONG CLICK` mqtt state `4` with openHAB to put the light into a different state using rules within openHAB.

I'd like to be able to use the `DOUBLE CLICK` state `3` or other states to carry out other functions in openHAB depending on the mqtt state of that same button, without operating the bound relay.

mcspr:
> Is there currently any way I can use the mqtt reporting of a button state whose BUTTON_ACTION is set to NONE?

See BUTTON_MQTT_SEND_ALL_EVENTS flag (default off)?
I have changed it a bit in the #2162 so it also checks `btnMqttSendAll<button#>` key (as boolean 1 or 0) to modify behaviour for a single button. After setting to 1 it should always send button event, even without any action attached.

davebuk:
Thanks. I thought BUTTON_MQTT_SEND_ALL_EVENTS was already default to ON. I'll test.

davebuk:
That all works. I did have one issue as I had set `Click repeat delay` in the webUI general section to `0` so accidental presses stop resets or AP mode. I've had to set it to 500ms to get the double/triple clicks to work.

Can `Click repeat delay` be defined per button at build to override the webUI setting?

mcspr:
I've yet to add this to config list in the Wiki:
https://github.com/xoseperez/espurna/blob/e2e0409c5962e4a90ec6632f8636ad88d7841db3/code/espurna/button.ino#L581-L596
`btnRepDel<#index>` key can be used for specific button. It will cancel out what WebUI sets, which is global default value.

mcspr:
I was mistaken about the order above. To actually do that:
```diff
diff --git a/code/espurna/button.ino b/code/espurna/button.ino
index ff311352..7f5b7796 100644
--- a/code/espurna/button.ino
+++ b/code/espurna/button.ino
@@ -490,7 +490,7 @@ void _buttonConfigure() {
 // TODO: compatibility proxy, fetch global key before indexed
 template<typename T>
 unsigned long _buttonGetSetting(const char* key, unsigned char index, T default_value) {
-    return getSetting(key, getSetting({key, index}, default_value));
+    return getSetting({key, index}, getSetting(key, default_value));
 }
```

Without the patch `btnRepDel` overrides every `btnRepDel0`, `btnRepDel1` etc., `btnRepDel0` read second, then hard-coded value.
With the patch `btnRepDel0` is read first, then `btnRepDel`, then hard-coded value.

Issue #1458, opened 2018-12-28
Tuya devices


germaguire:
Has anyone looked at tuya device support for espurna? All of the wifi devices have an esp8266ex board TYWE1S so I presume they are candidates for reflashing.
https://docs.tuya.com/en/hardware/WiFi-module/wifi-e1s-module.html

Im still looking for a more definitive list of products but it looks interesting.
https://en.tuya.com

Ger.

xoseperez:
These are no different from any other ESP8266 module. TheTeckin SP22 v1.4+ uses a Tuya TYWE2S and is supported in the current `dev` branch.

![20181207_090543s](https://user-images.githubusercontent.com/706487/50514652-e041d780-0a9f-11e9-8cf3-d7b1ea810020.jpg)

Wikibear:
Question will be answered here:
https://github.com/xoseperez/espurna/issues/1431

germaguire:
Some interesting developments re not having to solder Tuya devices in order to flash espurna. Have to wait until this evening to test but I have some new light bulbs, extension socket and wall plugs ready to try out.
https://github.com/ct-Open-Source/tuya-convert

https://www.youtube.com/watch?v=O5GYh470m5k&feature=youtu.be

forast3r:
First of all, my apologies for posting on a closed issue.
Can't thank @xoseperez enough for the great work with espurna firmware.

@End0rphi is the push button on the kloe4 power strip working?
Flashed Espurna 1.13.5 on a kloe4 with tuya-convert but the button only turns off the power strip after keeping it pushed for 10 or more seconds. However haven't been able to switch on any of the relays with the button. Does the boot mode of the individual switches influence in this behaviour? Shouldn't the button toggle (or at least switch all on or off)?

End0rphi:
I can confirm that in the default configuration for this device, the button does not switch the relays. This is expected behavior according to [the device entry in the wiki](https://github.com/xoseperez/espurna/wiki/Hardware-Zhilde-ZLD-44EU-W).

I'm quite sure the button behavior can be changed with the -DBUTTONx_MODE C preprocessor flags documented [here](https://github.com/xoseperez/espurna/wiki/Buttons-and-switches), but I have not tried it

forast3r:
Thank you for the confirmation and the links to the wiki!

PR #1774, opened 2019-06-13
add support for JINVOO VALVE SM-AW713

reaper7:
jinvoo-valve-sm-aw713


reaper7:
Type: Other
Board: Jinvoo SM-AW713
Build flag: JINVOO_VALVE_SM_AW713
Description: motorized valve

mcspr:
Thanks!

Can you please also add platformio.ini entries for the .bin releases?
And I think it can also go into the readme here? https://github.com/xoseperez/espurna#embedded-switches (with images/devices/..pic..)

reaper7:
image added by this commit:
https://github.com/xoseperez/espurna/pull/1774/commits/2bec5e5b7aac6a777f05b0d894376cc3fff9c3ab

about platformio.ini, I do not use myself, it's hard for me to guess what to write.
it seems that the entry should look similar to itead-sonoff-basic, flash size is the same.

mcspr:
No problem, I'll update the pio files then.
And right,, OTHER section is a better fit (not exactly a switch).

also, for #1458 we should probably in have a general info for tuya upgrade method. Sadly, looking at recent issues at https://github.com/ct-Open-Source/tuya-convert/issues, newest devices do not support it :/

reaper7:
about newest devices, yes, there may be a problem, although tuya has forwarded patch implementation to individual brands. But we know how the eastern producers work (sell and forget) :) Most of the less popular devices, already on sale, will never receive this patch...

mcspr:
I added device page to the wiki and some basic tuya description. Idk if it is worth inserting exact steps from tuya-convert readme though, but I added general description of that patch and how to flash specific .bin

PR #1779, opened 2019-06-17
Fix LED Invert for YJZK Switches

CrazyIvan359:
I noticed that the hardware profiles for the YJZK switches have `LED1_PIN_INVERSE` set to `0`, but these devices actually have inverted LED logic. I have corrected all 3 profiles in `hardware.h`.

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -975,7 +975,7 @@
 
     // LEDs
     #define LED1_PIN            13
-    #define LED1_PIN_INVERSE    0
+    #define LED1_PIN_INVERSE    1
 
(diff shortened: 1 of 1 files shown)


ColinShorts:
Whilst I'm not 100% sure on this, I think that was the only real difference between the config on the sonoff t1's and these clones. I've never had a yjzk switch to test myself, but perhaps there are different hardware revisions of the yjzk switches? If not, you could probably drop the separate config and point people to use the sonoff t1 binaries instead?

CrazyIvan359:
@ColinShorts I didn't see that profile. Definitely a possibility that I have a T1 knockoff and not a YJZK device.

PR #1787, opened 2019-06-23
Add support for MagicHome RGBWW LED strip

user890104:
This adds support for MagicHome 5 channel LED strip. It is using ESP8285 board, labelled as ESP-M4, has 5 powerful MOSFETs (M3006D, up to 80A! continuous/160A peak) at 12-24V. They are driven by smaller transistors. Red LED is connected to 3.3V power supply, and can't be controller. Green LED is connected to GPIO. Pin mappings:
BTN IO0
R Q8 Q2 IO14
G Q6 Q3 IO12
B Q9 Q4 IO13
WW Q10 Q5 IO15
CW Q7 Q1 IO5
GREEN_LED IO2
Looks like the same board is used in another product which has support for RF remote, but mine doesn't, so I don't include RF config, if yours has it should be a simple copy/paste job to add it.


user890104:
Some pictures and order link
photo_2019-06-23_09-59-21
photo_2019-06-23_09-59-27
photo_2019-06-23_09-59-24
photo_2019-06-23_09-59-38
https://www.aliexpress.com/item/32827053145.html
Pinout of the ESP8285 module (ESP-M4, compatible with ESP-M2)
photo_2019-06-23_10-10-36
Board markings: ZJ-ESPM-5CH-B V1.3

user890104:
IO4 goes to R25, which is not populated on my board, and in the area of missing RF remote components, so I guess it received RF remote signal after processing by a decoder IC (probably U4).

CrazyIvan359:
I also made a profile for this device because I couldn't find it in the hardware. Turns out it is listed in espurna as an Arilux AL-LC06. Perhaps the wiki page for that device should be updated to include the Magic Home device name as well.

mcspr:
Huh, it does indeed. Only difference is in white channels:
https://github.com/xoseperez/espurna/blob/e445cf8c7003827680d63f33b391bc95faec5de3/code/espurna/config/hardware.h#L1675-L1696

CrazyIvan359:
You sure that @user890104 configuration has the white channels correct? I have one of these devices that I've been using with my custom profile up until a week ago and when I flashed using the LC06 profile all the channels still work as expected.

user890104:
I'm pretty sure my configuration is correct, because I have measured the traces and marked which mosfet/transistor pair goes into which channel. Then I connected the wires according to the markings on the LED strip and it worked fine.
If the board is the same, there's a push button, which can be configured in the LC06 block to toggle the lights. It is not directly accessible (unless you have a thin object to poke it), but can be wired outside for "emergency" control (i.e. when your home automation setup is not accessible, and you need to turn on the lights somehow). Or a 3D-printer hat can be attached to it, so it sticks out of the encolsure.

CrazyIvan359:
Are you using an RGBWW strip?

As for the button, I don't use it so it doesn't affect me. It could be added to the LC06 profile though. 

On a side note, you say those transistors can handle 80A continuous and 160A peak. While is is true, you need to have some massive heatsink and good fans to run that much power through them. You shouldn't be running more than 4A or so without adding cooling

user890104:
I am using this strip: https://www.aliexpress.com/item/32839614476.html type DC12V IP67

I am aware of the power dissipation limitations, just noted that they used a MOSFET which is rated for 20 times the advertised rating, so it should be possible to switch loads close to the specs. Based on my experience with chinese products that use components which are not rated for the purpose, this one looks like it's properly done.

CrazyIvan359:
The MOSFET might be rated for 20x what they say the device can do, but if the board can't handle the current and you don't have proper cooling then you'll just turn your device into slag or worse start a fire.

Don't know about the white pins being swapped then. From memory I have the exact same board as you, but my white channels work as expected with the LC06 profile.

user890104:
> The MOSFET might be rated for 20x what they say the device can do, but if the board can't handle the current and you don't have proper cooling then you'll just turn your device into slag or worse start a fire.
> 
> Don't know about the white pins being swapped then. From memory I have the exact same board as you, but my white channels work as expected with the LC06 profile.

Can you take a picture of the PCB of your board? Specifically the side with components. Also, can you measure which pin is routed to each of the small transistors? I use the bottom pad of the second resistor to the right of each transistor.
image

CrazyIvan359:
Give me some time, it is installed under my kitchen counter presently. I will try to do it tomorrow.

CrazyIvan359:
Well I'm at a loss. Pictures below, my board is the exact same as yours. My only idea is that I have swapped my white channels in my config and been using the LC06 profile longer than I thought. I'm out of time to verify if that is the case right now (I am using it as a 5 channel dimmer device, not driving an RGBCW strip).

IMG_20190625_190411
IMG_20190625_190432

user890104:
I have triple-checked my configuration, and it seems like pins 5 and 15 should be as specified in my first comment. I wasn't sure if the white channels are wired correctly, so I connected red to ww, blue to cw and the results were as expected - red when fully warm white, blue when fully cold white. Pictures attached. Also, a picture of the LED strip I'm using.
IMG_20190626_133833
IMG_20190626_135758
Warm white (channel 4, pin 15):
Screenshot_20190626-135812
IMG_20190626_135910
Cold white (channel 5, pin 5):
Screenshot_20190626-135921
IMG_20190626_135931

CrazyIvan359:
Well I'm confused now. I did some more testing and have discovered two things:

1) You are correct, the white channels are reversed compared to the LC06
2) The channel assignments behave strangely when not using RGBWW in espurna.

When using this device and the LC06 profile with no colour options turned on, the channels map as follows:
```
1/0 > R
2/1 > WW
3/2 > G
4/3 > B
5/4 > CW
```
But then if I turn on colour, rgb, white, and warm white the channels all behave as expected, with WW and CW being swapped as you have confirmed. I don't know what to make of this...

PR #1788, opened 2019-06-23
Update debugSend / debugSend_P

mcspr:
- <s>remove varargs handling from debugSend and debugSend_P</s>
- chain debugSend_P into basic _debugSend instead of duplicating code, pass va_list between them
- when possible, use stack buffer and fallback to heap buffer only when it is more that 64 chars
- check if new[] was successful


mcspr:
Having remembered https://github.com/xoseperez/espurna/compare/dev...mcspr:ws/optimize, what I did not is template size overhead. I'll inspect it more closely, but this does not look good :/

### env:travis01

Size|Description
---|---
579280|templates, both inline
567008|templates, debugSend_P inline
558752|templates, debugSend inline
558288|templates, no inline
551840|old varargs
570832|debugSend varargs, debugSend_P as macro
551920|debugSend varargs, debugSend
551856|debugSend_P, debugSend, _debugSend(format, va_list)

("inline" means `inline __attribute__((always_inline))`

Reverting the template part, but keeping other things

PR #1793, opened 2019-06-25
Update Espressif library, resolve PIO 4 warnings

Niek:


mcspr:
Per https://github.com/platformio/platformio-core/issues/2716, I don't think the old one is needed? I worked out how to share libraries - https://gist.github.com/mcspr/80829350513faf37d3154ab67c40ae8b - so upgrade to 4 is pretty easy. Just need to clean up .ini and I'll push that right when it is released

Niek:
Hmm that's not how I read it. If you have only ``default_envs`` in there, it won't work on PIO 3. But if you keep ``env_default`` in there it works in both versions, but you still get the annoying warning.

mcspr:
What is the reason to support both? The way it worked before, we just updated to the new format

Niek:
Alright, I thought it'd be nice to support both for some amount of time

mcspr:
Platformio IDE will get updated to 4. CLI users can ignore it slightly longer. But we don't actually have something broken with 4, do we? Besides the libs stuff for the travis

Supporting both we get annoying warning on both :)
```
[mcspr@saturn wat]$ pio run
Warning! `env_default` configuration option in section [platformio] is deprecated and will be removed in the next release! Please use `default_envs` instead
...
[mcspr@saturn wat]$ ~/pio2/bin/pio run
Warning! Ignore unknown `default_envs` option in `[platformio]` section
...
```

Niek:
OK agreed, dropping the PIO 3 statement.

Niek:
Updated to dev

code/platformio.ini
```diff
@@ -65,6 +65,7 @@ build_flags_4m3m = ${common.build_flags} -Wl,-Teagle.flash.4m3m4s.ld
 # ------------------------------------------------------------------------------
 upload_port = "${sysenv.ESPURNA_IP}"
 upload_flags = --auth="${sysenv.ESPURNA_AUTH}"
+upload_protocol = espota
```
mcspr: Have not really followed the update_protocol syntax. Do you use this in a custom env: ?

Niek:
I  PIO 4 you get a warning with a OTA env if this is not set. See also https://docs.platformio.org/en/latest/projectconf/section_env_upload.html

PR #1798, opened 2019-06-27
Add ability to print all GPIO values at once

Niek:


mcspr:
Note that PSTR, unlike the normal const char*, is stored twice:
https://github.com/esp8266/Arduino/blob/f9009b8a5e141235d3380cc6c6c06223008a1099/tools/sdk/libc/xtensa-lx106-elf/include/sys/pgmspace.h#L37
Maybe it can always use a loop and stop / break in needed GPIO or just print all of them otherwise?

BTW. Sort of a nice-to-have, but I have not tried it yet (or maybe it is not even needed info)
per https://bbs.espressif.com/viewtopic.php?t=1140#p3806, gpio is marked as enabled when it is in output mode. https://github.com/esp8266/Arduino/blob/5a47cab77d396afeb842c9161327f9f4dca7c833/cores/esp8266/esp8266_peri.h#L57 is arduino-style macro that reads that register, conveniently accepting pin number as an argument.

Niek:
Thanks for the pointers, I optimized the code and added input/output mode at each pin. I think we should support writing to both types, as it's useful for both cases.

Issue #1759, opened 2019-06-05
MQTT sometimes does not notice broker shutdown


mcspr:
**Bug description**

When MQTT broker server is physically disconnected from the network (for example, network cable is removed), device still thinks that connection is OK. Neither onDisconnect or onTimeout firing for MQTT module.
mqtt.info shows state as connected, mqtt.reset finally resets it

While this seems like a bug in async mqtt client / asynctcp, still recording it here as it may be important "feature" to know. And it may be fixed by manual keep-alive, tracking last successful mqttSend() call

**Steps to reproduce**

1. `mqttQoS=1`, `mqttKeep=300`, `hbInterval 5` reset, wait until mqtt is connected on the device
2. remove ethernet cable from the server
3. debug logs still show that mqtt client is sending data, incrementing qos. after some time qos becomes 0 (ref: #1370), meaning that network layer does not accept any more data.

**Additional context**

Using async-mqtt-client master (or v0.8.1, same thing), Core 2.3.0 & 2.5.2 lwip2
Network is Linux server, Router that is directly wired to the server and ESP connected to the Router's AP. Both on the same LAN network bridged together.

I waited for about 10 min and reconnected the cable, after which MQTT reconnection finally happened. 

tcpdump on the router shows that device is still sending data, but it does look like it just tries to retransmit not acked data:
```
08:18:33.534970 IP ESPURNA-35A259.lan.51606 > orion.lan.1883: Flags [P.], seq 1749:1785, ack 181, win 1924, length 36
08:18:36.559839 IP ESPURNA-35A259.lan.51606 > orion.lan.1883: Flags [P.], seq 1749:1785, ack 181, win 1924, length 36
08:18:42.584820 IP ESPURNA-35A259.lan.51606 > orion.lan.1883: Flags [P.], seq 1749:1785, ack 181, win 1924, length 36
08:18:54.784220 IP ESPURNA-35A259.lan.51606 > orion.lan.1883: Flags [P.], seq 1749:1785, ack 181, win 1924, length 36
08:19:19.333148 IP ESPURNA-35A259.lan.51606 > orion.lan.1883: Flags [P.], seq 1749:1785, ack 181, win 1924, length 36
08:20:08.005749 IP ESPURNA-35A259.lan.51606 > orion.lan.1883: Flags [P.], seq 1749:1785, ack 181, win 1924, length 36
...
08:21:45.550983 ARP, Request who-has orion.lan tell ESPURNA-35A259.lan, length 28

```
[...]

PR #1799, opened 2019-06-28
Telnet server: alternative to ESPAsyncTCP

Niek:
This is a rewrite of the telnet module, it's no longer using ESPAsyncTCP but the regular WiFiServer/WiFiClient libraries. Tested it quite a bit, but more testing and feedback is appreciated.

### DIFF
--- a/code/espurna/telnet.ino
+++ b/code/espurna/telnet.ino
@@ -230,7 +230,7 @@ void _telnetNewClient(AsyncClient *cl) {
 
     for (unsigned char i = 0; i < TELNET_MAX_CLIENTS; i++) {
 
-        if (!_telnetClients[i]) {
+        if (!_telnetClients[i] || !_telnetClients[i]->connected()) {
 
             _telnetClients[i] = std::unique_ptr<AsyncClient>(client);
 
--- a/code/espurna/telnet.ino
+++ b/code/espurna/telnet.ino
@@ -8,12 +8,9 @@ Parts of the code have been borrowed from Thomas Sarlandie's NetServer
 
 */
 
-#define TELNET_SERVER_ASYNC        0
-#define TELNET_SERVER_WIFICLIENT   1
-
 #if TELNET_SUPPORT
 
-#if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+#if TELNET_SERVER == TELNET_SERVER_WIFISERVER
     #include <ESP8266WiFi.h>
     WiFiServer _telnetServer = WiFiServer(TELNET_PORT);
     WiFiClient _telnetClients[TELNET_MAX_CLIENTS];
@@ -47,10 +44,10 @@ void _telnetWebSocketOnSend(JsonObject& root) {
 #endif
 
 void _telnetDisconnect(unsigned char clientId) {
-    #if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+    #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
         _telnetClients[clientId].stop();
     #else
-        _telnetClients[clientId].free();
+        _telnetClients[clientId].close(true);
     #endif
     wifiReconnectCheck();
     DEBUG_MSG_P(PSTR("[TELNET] Client #%d disconnected\n"), clientId);
@@ -97,7 +94,7 @@ void _telnetData(unsigned char clientId, void *data, size_t len) {
     // C-d is sent as two bytes (sometimes repeating)
     if (len >= 2) {
         if ((p[0] == 0xFF) && (p[1] == 0xEC)) {
-            #if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+            #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
                 _telnetClients[clientId].stop();
             #else
                 _telnetClients[clientId].close(true);
@@ -107,7 +104,7 @@ void _telnetData(unsigned char clientId, void *data, size_t len) {
     }
 
     if ((strncmp(p, "close", 5) == 0) || (strncmp(p, "quit", 4) == 0)) {
-            #if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+            #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
                 _telnetClients[clientId].stop();
             #else
                 _telnetClients[clientId].close(true);
@@ -140,7 +137,7 @@ void _telnetData(unsigned char clientId, void *data, size_t len) {
     #endif
 }
 
-#if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+#if TELNET_SERVER == TELNET_SERVER_WIFISERVER
 void _telnetLoop() {
     if (_telnetServer.hasClient()) {
         int i;
@@ -180,7 +177,7 @@ void _telnetLoop() {
                 #else
                     _telnetClientsAuth[i] = !_telnetAuth;
                     if (_telnetAuth) {
-                        if (getAdminPass().length() != 0) {
+                        if (getAdminPass().length()) {
                             _telnetWrite(i, "Password: ");
                         } else {
                             _telnetClientsAuth[i] = true;
@@ -226,7 +223,6 @@ void _telnetNewClient(void *cl) {
         if (!telnetSTA) {
             DEBUG_MSG_P(PSTR("[TELNET] Rejecting - Only local connections\n"));
             client->onDisconnect([](void *s, AsyncClient *c) {
-                c->free();
                 delete c;
             });
             client->close(true);
@@ -293,7 +289,6 @@ void _telnetNewClient(void *cl) {
 
     DEBUG_MSG_P(PSTR("[TELNET] Rejecting - Too many connections\n"));
     client->onDisconnect([](void *s, AsyncClient *c) {
-        c->free();
         delete c;
     });
     client->close(true);
@@ -321,7 +316,7 @@ void _telnetConfigure() {
 }
 
 void telnetSetup() {
-    #if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+    #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
         espurnaRegisterLoop(_telnetLoop);
         _telnetServer.setNoDelay(true);
         _telnetServer.begin();
--- a/code/espurna/telnet.ino
+++ b/code/espurna/telnet.ino
@@ -207,10 +207,10 @@ void _telnetLoop() {
             } else {
                 // Read data from clients
                 while (_telnetClients[i] && _telnetClients[i]->available()) {
+                    char data[TERMINAL_BUFFER_SIZE];
                     size_t len = _telnetClients[i]->available();
-                    char data[len];
-                    unsigned int r = _telnetClients[i]->readBytes(data, len);
-
+                    unsigned int r = _telnetClients[i]->readBytes(data, min(sizeof(data), len));
+                    
                     _telnetData(i, data, r);
                 }
             }
--- a/code/espurna/telnet.ino
+++ b/code/espurna/telnet.ino
@@ -220,9 +220,7 @@ void _telnetLoop() {
 
 #else // TELNET_SERVER_ASYNC
 
-void _telnetNewClient(void* client) {
-    AsyncClient *client = (AsyncClient *)cl; // in order to avoid using AsyncClient in function signature
-
+void _telnetNewClient(AsyncClient* client) {
     if (client->localIP() != WiFi.softAPIP()) {
         // Telnet is always available for the ESPurna Core image
         #ifdef ESPURNA_CORE
--- a/code/espurna/telnet.ino
+++ b/code/espurna/telnet.ino
@@ -55,7 +55,7 @@ void _telnetDisconnect(unsigned char clientId) {
 }
 
 bool _telnetWrite(unsigned char clientId, const char *data, size_t len) {
-    if (_telnetClients[clientId]->connected()) {
+    if (_telnetClients[clientId] && _telnetClients[clientId]->connected()) {
         return (_telnetClients[clientId]->write(data, len) > 0);
     }
     return false;
@@ -136,7 +136,7 @@ void _telnetLoop() {
         int i;
 
         for (i = 0; i < TELNET_MAX_CLIENTS; i++) {
-            if (!_telnetClients[i]->connected()) {
+            if (!_telnetClients[i] || !_telnetClients[i]->connected()) {
                 _telnetClients[i] = std::unique_ptr<WiFiClient>(new WiFiClient(_telnetServer.available()));
 
                 if (_telnetClients[i]->localIP() != WiFi.softAPIP()) {
@@ -192,7 +192,7 @@ void _telnetLoop() {
 
     // Read data from clients
     for (int i = 0; i < TELNET_MAX_CLIENTS; i++) {
-        while (_telnetClients[i]->available()) {
+        while (_telnetClients[i] && _telnetClients[i]->available()) {
             char data[512];
             size_t len = _telnetClients[i]->available();
             unsigned int r = _telnetClients[i]->readBytes(data, min(sizeof(data), len));
@@ -203,7 +203,6 @@ void _telnetLoop() {
 }
 #else
 void _telnetNewClient(void *cl) {
-    DEBUG_MSG_P(PSTR("[TELNET] Got new client\n"));
     AsyncClient *client = (AsyncClient *)cl; // in order to avoid using AsyncClient in function signature
 
     if (client->localIP() != WiFi.softAPIP()) {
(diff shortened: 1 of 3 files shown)


mcspr:
I will try this later today, but I think there are two issues:
- the way logging works, we can't issue any writes for the wificlient. For example, try async-mqtt-client active or connect to the webui
- is data[2048] part really needed? stack on esp is pretty limited, unless we are in SYS context (ticker, async callbacks etc.). ref. CONT_STACKSIZE, which is 4k by default

Misiu:
@Niek @mcspr  why this is needed? I'm not an expert and I don't try to disagree with the changes, but I want to understand why this is needed and what benefits it has.
I always used ESPurna as a "best practice" reference project, I learned A LOT from it.
ESPAsyncTCP was used because of `async` support, and I always thought that we should write code in an async way to avoid the need to relay on loop.
I'll be grateful for the explanation :)

Niek:
@Misiu ESPAsyncTCP has not been updated in over a year. It also doesn't play well with Arduino core >= 2.5 especially on the SSL side of things.

Someone made a fork to work with BearSSL (https://github.com/Adam5Wu/ESPAsyncTCP) but it requires a bunch other forked stuff like the Arduino core and it hasn't been updated in a year either.

@mcspr can you elaborate on the first point? I don't seem to run any issues with logging (but in my tests I'm not using any ESPAsyncTCP modules).

The 2048 char array is way overkill indeed for incoming data, maybe we can limit it to 512 or something similar.

mcspr:
Practical reason is to allow to build without it (using pubsubclient for mqtt, for example) and possibly use Core's server + BearSSL. Not yet though, and I kind of missed removal of the existing one. @Niek We might want to guard new telnet server the same way as #1751 , eg TELNET_SERVER == ...

*not to get ot... :accept:*

Async stuff is kind of cool, but sometimes we hit surprising things that were either already handled by WiFiClient abstraction<sup>1</sup> or completely avoided by it. One benefit, however, since we implement it :) - we can fix some quicks between Core versions, while this still applies. We kind of have two ways now - create more robust approach for clients / servers than current one e.g. manually supply all callbacks and streamline buffering ([AsyncPrinter / AsyncBufferred... ?](https://github.com/me-no-dev/ESPAsyncTCP/blob/master/src/AsyncPrinter.h)). One problem is that BearSSL specifically employs yields ([search ref:](https://github.com/earlephilhower/bearssl-esp8266/search?q=optimistic_yield&unscoped_q=optimistic_yield)) inside some heavy calculations, so they need to be offloaded to loop *somehow* :/

Or, try to patch Core WiFiClient::write timeouts to make them behave the same way AsyncClient::write does - just try to write something to the network layer and do nothing otherwise (if there's already too much stuff). I think I saw some issues at the esp8266/Arduino, but not much traction there. Wonder how hard it is.

<sup>1</sup> [ClientContext aka Core's ESPAsyncTCP](https://github.com/esp8266/Arduino/blob/master/libraries/ESP8266WiFi/src/include/ClientContext.h)

Niek:
I have updated the PR, supporting 2 telnet server implementations now. I tested a lot and it seems to work fine on my setup - please review.

code/espurna/telnet.ino
```diff
@@ -143,31 +232,30 @@ void _telnetNewClient(AsyncClient *client) {
...
-
     }
 
     for (unsigned char i = 0; i < TELNET_MAX_CLIENTS; i++) {
 
-        if (!_telnetClients[i] || !_telnetClients[i]->connected()) {
+        if (_telnetClients[i].disconnected()) {
 
-            _telnetClients[i] = client;
+            _telnetClients[i] = *client;
```
mcspr: I thinks this leaks. `client` ptr data is copied to our array but the actual pointer is never destroyed.

Just keep the capture technique that was used previously or, if you want to keep the methods looking common (i.e. dot syntax), you can use std::unique_ptr as AsyncClient pointer holder and [`.reset(new_ptr)`](https://en.cppreference.com/w/cpp/memory/unique_ptr/reset) when needed. ...ptr can be nulled, which deletes pointer. Bool checks still work (if (unique_ptr....)), checking if captured pointer is present.

code/espurna/telnet.ino
```diff
@@ -38,16 +47,18 @@ void _telnetWebSocketOnSend(JsonObject& root) {
 #endif
 
 void _telnetDisconnect(unsigned char clientId) {
-    _telnetClients[clientId]->free();
-    delete _telnetClients[clientId];
-    _telnetClients[clientId] = NULL;
+    #if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
+        _telnetClients[clientId].stop();
+    #else
+        _telnetClients[clientId].free();
```
mcspr: As discovered in #1806, this does nothing for the memory:
https://github.com/me-no-dev/ESPAsyncTCP/blob/7e9ed22ed0078097f895dfee6ebcec90046b81b9/src/ESPAsyncTCP.cpp#L216-L222
Just a status method

See the other comment regarding memory management (& possibly unique_ptr )

code/espurna/telnet.ino
```diff
@@ -186,14 +274,19 @@ void _telnetNewClient(AsyncClient *client) {
                 _telnetClientsAuth[i] = true;
             #else
                 _telnetClientsAuth[i] = !_telnetAuth;
-                if (_telnetAuth) _telnetWrite(i, "Password: ");
+                if (_telnetAuth) {
+                    if (getAdminPass().length() != 0) {
```
mcspr: `if (getAdminPass().length())` ?

code/espurna/telnet.ino
```diff
@@ -8,12 +8,21 @@ Parts of the code have been borrowed from Thomas Sarlandie's NetServer
 
 */
 
+#define TELNET_SERVER_ASYNC        0
+#define TELNET_SERVER_WIFICLIENT   1
+
 #if TELNET_SUPPORT
 
-#include <ESPAsyncTCP.h>
+#if TELNET_SERVER == TELNET_SERVER_WIFICLIENT
```
mcspr: `..._WIFISERVER`?

code/espurna/telnet.ino
```diff
@@ -8,12 +8,21 @@ Parts of the code have been borrowed from Thomas Sarlandie's NetServer
 
 */
 
+#define TELNET_SERVER_ASYNC        0
+#define TELNET_SERVER_WIFICLIENT   1
```
mcspr: Please move to types header.

code/espurna/telnet.ino
mcspr: Similarly, `new WiFiClient(...)` can be used. Since destructor stops it, maybe disconnect part can be simplified to just setting unique_ptr list member to null?

code/espurna/telnet.ino
Niek: That makes more sense indeed - changed

code/espurna/telnet.ino
Niek: OK, will change it to ``close(true)`` in that case

code/espurna/telnet.ino
Niek: 👍🏻

code/espurna/telnet.ino
Niek: 👍🏻

code/espurna/telnet.ino
Niek: Hmm, something like this? ``_telnetClients[i] = std::unique_ptr<AsyncClient>(client);``
We can't use the dot syntax then though, so we'd have to change the WiFiClient array to ``std::unique_ptr`` as well (correct me if I'm wrong).

mcspr: Nah, the std::unique_ptr can always be there. Array declaration like that does object construction asap, there are TELNET_MAX_CLIENTS WiFiClient objects created on boot.
```
WiFiClient _telnetClients[TELNET_MAX_CLIENTS];
...
std::unique_ptr<TELNET_CLIENT_CLASS> _telnetClients[TELNET_MAX_CLIENTS];
_telnetClients[0] = some_pointer_obj;
_telnetClients[0]->write(...);
```

Yes, you right, no dot syntax. ->

Niek: Thanks for the pointers! (sic)

I have updated the PR, it compiles fine but I'm unable to check right now because my dev board seems to have died on me.

Niek: Updated with fixes for the last issues - checked and I can confirm both server implementations are working fine now.

code/espurna/telnet.ino
```diff
@@ -38,9 +44,11 @@ void _telnetWebSocketOnSend(JsonObject& root) {
 #endif
 
 void _telnetDisconnect(unsigned char clientId) {
-    _telnetClients[clientId]->free();
-    delete _telnetClients[clientId];
-    _telnetClients[clientId] = NULL;
+    // ref: we are called from onDisconnect, async is already stopped
```
mcspr: c->close() already calls this, so we print out message twice

code/espurna/telnet.ino
```diff
@@ -120,13 +127,93 @@ void _telnetData(unsigned char clientId, void *data, size_t len) {
...
 
+        //no free/disconnected spot so reject
+        if (i == TELNET_MAX_CLIENTS) {
+            DEBUG_MSG_P(PSTR("[TELNET] Rejecting - Too many connections\n"));
+        }
+    }
+
+    // Read data from clients
+    for (int i = 0; i < TELNET_MAX_CLIENTS; i++) {
+        while (_telnetClients[i] && _telnetClients[i]->available()) {
```
mcspr: One last thing though. What happens when sync client times out? (network connection is cut) Should it check for ->connected() or new client will replace the old one?

Niek: Hmm I didn't think about that one. Best would probably be to check ``connected()`` and free the pointer at a timeout.

code/espurna/telnet.ino
```diff
@@ -120,13 +127,90 @@ void _telnetData(unsigned char clientId, void *data, size_t len) {
...
+            char data[512];
+            size_t len = _telnetClients[i]->available();
+            unsigned int r = _telnetClients[i]->readBytes(data, min(sizeof(data), len));
 
+            _telnetData(i, data, r);
+        }
+    }
+}
+#else
+void _telnetNewClient(AsyncClient *cl) {
```
Niek: Having AsyncClient in the function definition screws up building on environments where ESPAsyncTCP modules are not enabled. That's why I added a void* and manually cast the parameter.

The error looks something like this:
```
code/espurna/espurna.ino:545:23: error: variable or field '_telnetNewClient' declared void
code/espurna/espurna.ino:545:23: error: 'AsyncClient' was not declared in this scope
code/espurna/espurna.ino:545:36: error: 'client' was not declared in this scope
```

code/espurna/telnet.ino
Niek: Fixed in https://github.com/xoseperez/espurna/pull/1799/commits/da065b2b6e00bf9ecec0105327fe3037db6e490e

code/espurna/telnet.ino
Niek: I changed it back in https://github.com/xoseperez/espurna/pull/1799/commits/da065b2b6e00bf9ecec0105327fe3037db6e490e

code/espurna/telnet.ino
mcspr: Think so. But I don't see any ack timeout / keepalive mechanism in ClientContext or WiFiClient, to prevent the thing from #1759. I don't know, does telnet support any protocol-level keepalive? Async variant does ack timeout, it will fail after some writes

code/espurna/telnet.ino
mcspr: What exactly is your config? When I noticed that, I tried turning off web,thingspeak,alexa and switching to pubsub for mqtt - it still compiled.
Both 1.5.0 and 2.2.2 platforms, pio 4 rc (python3, if that ends up being the reason), if it somehow breaks prototype generation.

Niek: That's because the code in dev now has ESPAsyncTCP always included in prototypes, but I'm running your to-be-comitted OTA PR code which has this part removed: https://github.com/xoseperez/espurna/pull/1751/files#diff-52135e36de9de9ff791c1184a170b897

mcspr: Note that da065b2 does not build. 
How come ota_asynctcp.ino works then? It has AsyncClient in arguments.

And you are interested in inspecting the full source, C-z the running pio process and look at espurna/espurna.ino.cpp file

mcspr: ...because ota_asynctcp includes espasync header. ok, so that looks like a bug with #1751 , not with telnet. I'd go back to AsyncClient in argument list and either:
Return ESPAsyncTCP.h to the prototypes regardless of the ifdef conditions
Add `class AsyncClient;` line

Niek: Now that I think about it: I agree, makes sense to just keep it always in prototypes.h instead of making the code uglier - I changed it back. BTW it would make more sense if the generated espurna.ino.cpp would have macros around the forward declarations, but I don't think that's possible.

mcspr:
Regarding client timeout - it does not disconnect and only indication of any problems is write() returning 0. You can try running telnet client on a phone, for example, and disconnect WiFi connection. After running `info` in serial a couple of times, buffer is finally filled and every consecutive write() will timeout. Managing that will require some "write success" counter, but I'm not yet sure where it should go.

Niek:
The docs mention using ``flush(timeout)`` in order to check if the connection has timed out: https://arduino-esp8266.readthedocs.io/en/latest/esp8266wifi/client-class.html?highlight=wificlient#flush-and-stop

I haven't been able to get it to work though (always seem to return true).

mcspr:
IDK why travis does not pick up the latest commit when restarting the PR build.
Since the basics work, this can be added and the timeout thing worked out after that?

Niek:
Seems like a Travis CI issue. Do you have rights to restart jobs?

In any case, yes this can be merged, we'll figure out the timeout later.

Niek:
I pushed a last fix now (allocate only required buffer, not 512 chars] - travis also succeeded this time.

code/espurna/telnet.ino
```diff
@@ -120,13 +127,101 @@ void _telnetData(unsigned char clientId, void *data, size_t len) {
...
+    for (int i = 0; i < TELNET_MAX_CLIENTS; i++) {
+        if (_telnetClients[i]) {
+            // Handle client timeouts
+            if (!_telnetClients[i]->connected()) {
+                _telnetDisconnect(i);
+            } else {
+                // Read data from clients
+                while (_telnetClients[i] && _telnetClients[i]->available()) {
+                    size_t len = _telnetClients[i]->available();
+                    char data[len];
```
mcspr: Returning to my previous comment regarding stack, I don't think it makes sense to have this dynamic.
What I missed was that old 512 might've been too big too. _telnetData(...) will feed the data[] to the terminal buffer and it is limited by 128 bytes (TERMINAL_BUFFER_SIZE) anyways

Niek: OK, changed now

mcspr:
And following up on the https://github.com/xoseperez/espurna/pull/1799#issuecomment-507486225 (since we don't track ESPAsyncTCP issue yet and this is not really worth mentioning in the library issue tracker)... I have a sort-of working bearssl port based on the work from the https://github.com/Adam5Wu/ESPAsyncTCP
Diff would be unreadable though... And the only thing tested was https://canihazip/s with SyncClient example

Turns out, insecure option is a Core-specific thing. But, it can be safely used just by adding 2 declarations so they get linked from the Core's WiFiClient helpers
Thunk / Non-thunk part was not really a problem, and original code works straight away. If needed, thunk macro can remap required functions and builds with WPS will still work fine.
MFLN works too! Because it is as easy as adding a different buffer size for the bearssl API call. Quick solution is to have AsyncClient::setSSLParams(...) method to supply new BearSSL-specific options.
Certificates from SPIFFS are kind of a tricky thing to have, but basic X509List arg should work as it is right now (untested)
Duplication is not really as severe as expected. Maybe some things can be safely moved to be reusable.

Tracking ESPAsyncTCP repo, there are some problems with the classes that I had mentioned (AsyncPrinter and Buffer) regarding lwip2 buffer sizes.

mhightower83:
> Tracking ESPAsyncTCP repo, there are some problems with the classes that I had mentioned (AsyncPrinter and Buffer) regarding lwip2 buffer sizes.

@mcspr - could you point me to more information on this. I recently posted a PR at [me-no-dev/ESPAsyncTCP](https://github.com/me-no-dev/ESPAsyncTCP/pull/115#issue-298136350), I fixed numerous problems in TCPAsyncTCP and SyncClient. Many of the issues in SyncClient were generic in nature and were applied to the other modules; however, there wasn't any test code to verify the results.

mcspr:
My bad! I read the me-no-dev/ESPAsyncTCP#115 patch briefly a couple of days ago and made a mental note about those classes for some reason. I do see now that those changes are for the default values, so disregard my previous comment.

Issue #1808, opened 2019-07-04
Settings overwritten after NoFUSS OTA update


mcspr:
I too faced similar issues and device lost configuration after update via Nofuss.
I enabled debug in eeprom rotate and saw this -

```
Magic value for sector #1019 is 71
Calculated CRC: 0x8ADE
Stored CRC    : 0x8ADE
Magic value for sector #1018 is 12
Calculated CRC: 0x3A78
Stored CRC    : 0x40B9
Sector #1018 has not passed the CRC check
Magic value for sector #1017 is 0
Calculated CRC: 0x2DCD
Stored CRC    : 0x765F
Sector #1017 has not passed the CRC check
Magic value for sector #1016 is 161
Calculated CRC: 0xA924
Stored CRC    : 0xFFA0
Sector #1016 has not passed the CRC check
Current sector is #1019
Current magic value is #71
```
Clearly blocks 1016,1017 and 1018 are overwritten. While 1019 is intact so EEPROM Rotate picked up configuration from 1019, but since the latest data was not present in the sector 1019 . So the device lost some configuration keys.

But once again question is, why are these sectors are being overwritten during OTA. For information - I did a custom build with no SPIFSS enabled(In both previous and new binary).

I also disabled calling of custom_crash_callback as well, via an addition flag
#define SAVE_CRASH_INFO 0.

I think it is not an issue exactly with Espurna, but something with nofuss or esp8266httpUpdater itself(Which is used by Nofuss).
However its note worthy for all espura users as well, that if we build it with latest core, and deliver OTAs to devices, few sectors will be overwritten. Depending on the code size it may overflow all 4 sectors also. 
Anyway, i'll create another issue on arduino esp8266 core as well, but the same could also be true for ArduinoOta library which is used for web based OTA. 

Sectors from 378000 to 3f9000,3f9000 to 3fa000 , 3fa000 to 3fb000, 3fb000 to 3fc000 were used by eeprom rotate. And were clean. In flash dump also as expected. Then i took the dump of these same sectors right after OTA and saw these sectors were filled with something. 

<img width="641" alt="Screen Shot 2019-07-04 at 5 49 00 PM" src="https://user-images.githubusercontent.com/4976460/60666182-12cff100-9e84-11e9-8607-e91c64c1d3b1.png">

_Originally posted by @arihantdaga in https://github.com/xoseperez/espurna/issues/1595#issuecomment-508462424_

mcspr:
[...]

PR #1809, opened 2019-07-04
[Fix] Backing up EEPROM before performing OTA For fixing - [#1808]

arihantdaga:
This is for taking backup of eeprom into last sector (base sector) before updating the firmware. Is helpful for builds with no SPIFFS.


code/espurna/nofuss.ino
```diff
@@ -145,6 +145,7 @@ void nofussSetup() {
             #if WEB_SUPPORT
                 wsSend_P(PSTR("{\"action\": \"reload\"}"));
             #endif
+            eepromBackup(0);
```
mcspr: My original proposal was to insert this into the eepromRotate false condition

arihantdaga:
Yeah and i think i did another mistake also, this isnt the right place to keep eepromBackup, it should be kept in NOFUSS_UPDATING hook. 
I'll do another commit and fix everything.

mcspr:
Just to reiterate, it should do forced backup with nofuss (eepromBackup(0)) right after the rotation is disabled and there should be eepromCommit(false) in the eepromRotate (since it marks eeprom as _dirty, it will force a write next loop()).

You are correct that it is better to write before any flash writes from Updater (while EEPROM buffer is in memory, just to protect from any issues during the update or other accidental failures)

mcspr:
I think I covered all of the options? I have tried some builds without SPIFFS by using ArduinoOTA upgrade method (it does not leave loop(), same as nofuss) and upgrade triggered the rotation backup.
There a small unrelated fixes pending too, so I wanted to merge this right now. Thanks for pointing out the issue!

PR #1810, opened 2019-07-05
Use basic BearSSL ciphers on low memory boards

Niek:
See: https://github.com/esp8266/Arduino/blob/8b1af68e3f996592c4b152e793cde81de4b8957d/libraries/ESP8266WiFi/src/WiFiClientSecureBearSSL.cpp#L725-L854

With this flag enabled, BearSSL is compiled with only these 4 ciphers:
```
TLS_RSA_WITH_AES_128_CBC_SHA256 / AES128-SHA256
TLS_RSA_WITH_AES_256_CBC_SHA256 / AES256-SHA256
TLS_RSA_WITH_AES_128_CBC_SHA / AES128-SHA
TLS_RSA_WITH_AES_256_CBC_SHA / AES256-SHA
```

Setting this build flag saves around 45.5KB in build size, which is very significant on low memory boards. I think it would be nice to enable this by default on the 512KB/1MB boards.

I did some quick checks and most (if not all) SSL server support these ciphers. To check a site, try: ``nmap --script ssl-enum-ciphers -p 443 google.com``


Niek:
Maybe it also makes sense to define ``BEARSSL_BASIC``.

mcspr:
Maybe as a comment block / separate variable in [common]? Not default, but a general recommendation

Niek:
Makes sense - done!

PR #1812, opened 2019-07-07
Web OTA: check authentication result before accepting payload

mcspr:
Using HTTP upgrade method described in the wiki and omitting auth headers:
https://github.com/xoseperez/espurna/wiki/OTA

```
curl -XPOST  \
  -H "Content-Type: multipart/form-data" \
  -F "filename=@.pioenvs/esp8266-1m-ota/firmware.bin" \
  http://192.168.4.1/upgrade
```

Causes file handler to process OTA file and **write it to flash** and only then to go into the request handler, where it will issue auth error.

### DIFF
--- a/code/espurna/web.ino
+++ b/code/espurna/web.ino
@@ -125,6 +125,10 @@ void _onPostConfig(AsyncWebServerRequest *request) {
 
 void _onPostConfigData(AsyncWebServerRequest *request, String filename, size_t index, uint8_t *data, size_t len, bool final) {
 
+    if (!webAuthenticate(request)) {
+        return request->requestAuthentication(getSetting("hostname").c_str());
+    }
+
     // No buffer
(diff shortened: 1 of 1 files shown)


mcspr:
TODO: use AsyncWebHandler::setAuthentication(username, password) instead of doing manual auth?
Per dff81d6795c4b6647463645c2f484be578d10562, /ws is kept with ticker-based auth. Every other handler return value of _server->on(...)) needs to have setAuthentication.
Loosing log message and an easy ability to refresh the password. Password can be updated via change detection in the reload method and applied to local handlers ref storage. Or, handlers can be accessed by subclassing server class and [accessing _handlers attr](https://github.com/me-no-dev/ESPAsyncWebServer/blob/b0c6144886f4e1f88684f708d7b2974143d8601c/src/ESPAsyncWebServer.h#L390)

PR #1814, opened 2019-07-09
Adding Nexte RGBW bulb

konsgn:
Added another version of RGBW bulb with a slightly different pinout from another generic lamp.


mcspr:
Maybe just use NEXETE as manufacturer, since we actually know it? Other generic bulb is kind of noname

BTW pio ini is missing a basic env without -ota, did the serial upload work with empty upload_... settings?

konsgn:
Updated! and non-ota prolly works fine, but flashing initially requires some really tricky soldering to get to the tx/rx test points sandwiched between two pcb's

mcspr:
Thanks! I merged with dev to patch recent PIO changes and updated arduino.h too

PR #1817, opened 2019-07-12
Added some new features to WebUI

foxman69:
Full change-list below:
1) Added Master schedule button
2) Added nicknames for each switch
3) Added an option to change username when logging in using WebUI
4) Fixed the time in scheduling to have a leading zero when having a 1 digit hour or minute

### DIFF
--- a/code/espurna/scheduler.ino
+++ b/code/espurna/scheduler.ino
@@ -138,8 +138,7 @@ void _schCheck() {
     time_t local_time = now();
     time_t utc_time = ntpLocal2UTC(local_time);
     
-    if (getSetting("schAll").toInt() == 0)
-    {
+    if (getSetting("schAll").toInt() == 0) {
         return;
     }
     
--- a/code/espurna/scheduler.ino
+++ b/code/espurna/scheduler.ino
@@ -34,8 +34,7 @@ void _schWebSocketOnSend(JsonObject &root){
     JsonArray& switch_ = schedules.createNestedArray("schSwitch");
     JsonArray& action = schedules.createNestedArray("schAction");
     JsonArray& type = schedules.createNestedArray("schType");
-    JsonArray& hour = schedules.createNestedArray("schHour");
-    JsonArray& minute = schedules.createNestedArray("schMinute");
+    JsonArray& time = schedules.createNestedArray("schtime");
     JsonArray& utc = schedules.createNestedArray("schUTC");
     JsonArray& weekdays = schedules.createNestedArray("schWDs");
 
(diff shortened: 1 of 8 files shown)


code/espurna/scheduler.ino
```diff
@@ -136,7 +137,11 @@ void _schCheck() {
 
     time_t local_time = now();
     time_t utc_time = ntpLocal2UTC(local_time);
-
+    
+    if (getSetting("schAll").toInt() == 0) {
```
mcspr: `schEnabled`?

code/html/index.html
```diff
@@ -1888,6 +1902,10 @@ <h2>
                 <div class="pure-u-1 pure-u-lg-1-4"><label>Pulse time (s)</label></div>
                 <div class="pure-u-1 pure-u-lg-1-4"><input name="relayTime" class="pure-u-1" type="number" min="0" step="0.1" max="3600" /></div>
             </div>
+            <div class="p-g">
+                <div class="p-u-1 p-u-lg-1-4"><label>Nickname</label></div>
+                <div class="p-u-1 p-u-lg-1-4"><input name="relayNickname" class="p-u-1" type="text" original="" /></div>
```
mcspr: original= is a dynamic attr, no need to set it

code/espurna/scheduler.ino
foxman69: schEnabled is already taken (its the toggle of each schedule), i changed to to schMasterEnabled so its more indicative

code/html/index.html
foxman69: deleted it

foxman69:
I have fixed what you have mentioned (deleted the original attribute and changed the schAll name)
I actually see the "nickname" option as more indicative than label or tag, but you change choose what ever you like

I did changed the input type to input=time and deleted all the related js about leading zeros (when using input=time there is no need to this fix)

foxman69:
I did saw that the input=time in android is working well looks a bit cut when the phone is on vertical mode, I will check this weekend how to fix it...

foxman69:
I have fixed the problem in android and now it looks great
also made sure that there would always be a value when adding new schedule (Default value is 00:00)

mcspr:
Note that I have not forgotten about this, but I had a bit of a different idea for the relay naming.
If you followed the latest updates, one problem that we had was excessive payload size with repeating sensor descriptions whenever thingspeak or domoticz were enabled. Instead of sending those each time, sensor callback now sends those just once and caches them in the global 'magnitude' array.
Currently, we send out 'friendly' name via
https://github.com/xoseperez/espurna/blob/43a44cf96447dc2e04f0ef936573950df359889e/code/espurna/relay.ino#L693
That shows GPIO name, CH when it is light etc. What it can do instead is to check if `relayNicknameN` exists and send it instead of generating anything. SWITCHES config already shows that name, STATUS page needs a small adjustment to use it instead of generic 'Switch #'

foxman69:
I will check the latest updates and adjust my code in the next few days

PR #1822, opened 2019-07-13
Add Shelly 1PM

tonilopezmr:
### 💣 Issue

* Add new device [Shelly 1PM](https://shelly.cloud/shelly-1pm-wifi-smart-relay-home-automation/) support following [Shelly 1PM tasmota support](https://github.com/arendst/Sonoff-Tasmota/issues/5716).

### :tophat: What is the goal?

Integrate Shelly 1PM with espurna to control the relay, control built-in led, read wall switch changes, get current and temperature.

### :memo: How is it being implemented?

* Turn relay on/off using `GPIO15`
* Read switch state using `GPIO4`
* Get current and power from `BL0937`, *Shelly1PM doesn't use CF1 and SEL* then I set a free GPIO's for that two values, here is the `BL0937` configuration: 
  -  SEL: `GPIO12` //UNUSED
  -  CF1: `GPIO13` //UNUSED
  -  CF: `GPIO5`
* Use the built-in led in `GPIO0` with a reverse signal.
* Read the built-in temperature sensor using `NTCSensor`
* Read built-in button states for simple reset or factory reset using `GPIO2`

### ⚠️ Fix before merge / help wanted

* Current always is zero but power it's fine, why?
* It seems LED1_PIN is not working if I add `LED_MODE_ON` (Someone can test it too?)
   - Weird behavior: When I set `LED1_PIN 0` espurna doesn't set `pinMode`  to `OUTPUT` then if I try to run the following command `GPIO 0 0` it's not working, then If I set manually `pinMode(0, OUTPUT)` and turn on/off using the wall switch once, then the led it's always ON, even if the `LED_MODE_OFF` is activated.
* When I reload the espurna page twice the device crashes (I don't know if it's related to the current espurna version or shelly 1pm configuration)


tonilopezmr:
```
Exception (29):
epc1=0x4020eb0e epc2=0x00000000 epc3=0x00000000 excvaddr=0x00000000 depc=0x00000000

ctx: sys 
sp: 3ffffb80 end: 3fffffb0 offset: 01a0

>>>stack>>>
3ffffd20:  3fff861c 3fffa6f4 3fff09ac 4020d711  
3ffffd30:  00000b3f 00000b3f 00000b3f 00000000  
3ffffd40:  00000003 3fff7934 3fff09ac 40216a7d  
3ffffd50:  00000000 00000000 00000000 3fff60f4  
3ffffd60:  00000003 3fff60f4 3fff8470 40216b0b  
3ffffd70:  3ffe9558 00000000 3fffa884 00002000  
3ffffd80:  3fff6cbc 00000003 000000c0 000000a8  
3ffffd90:  00000001 00000080 3ffffdd0 00000000  
3ffffda0:  3fff7934 00000001 3fff09ac 402189db  
3ffffdb0:  3fff3b2c 00000980 00000980 4010020c  
3ffffdc0:  00000001 3fff7934 3ffffe10 4010068c  
3ffffdd0:  3ffea6f8 8001a8c0 3ffffe10 4022de94  
3ffffde0:  3fff2524 00000c53 00000c53 4010020c  
3ffffdf0:  00000001 00000000 3ffffe40 3fff8d84  
3ffffe00:  00000001 3fff7934 00000000 4022e7b2  
3ffffe10:  3fff7914 3ffe8611 40222330 4022f255  
3ffffe20:  00000000 00000000 3fff7934 40220a35  
3ffffe30:  3fff2524 00000000 3fff7934 4022131e  
3ffffe40:  3fff7934 3ffe8611 40220788 40221e84  
3ffffe50:  3fff77fc 00000004 3fff8f54 401004d8  
3ffffe60:  3fff2524 00000c39 00000c39 3fff2b10  
3ffffe70:  00000101 00000004 3fff8f54 40221355  
3ffffe80:  3fff8d84 011e3555 00002200 40107660  
3ffffe90:  3fff8edc 3fff2b1c 3fff8d84 4022f2b4  
3ffffea0:  3fff8d84 00000101 3fff77fc 4022248d  
3ffffeb0:  00000001 3fff2d08 3fff77fc 4021b1f4  
3ffffec0:  3fff762c 3fff2b18 3fff2b14 4021b214  
3ffffed0:  3fff762c 3fff2b18 3fff2b14 402495a5  
3ffffee0:  00000000 8401a8c0 00000010 00000000  
3ffffef0:  00000010 00000000 3fff13fc 40107650  
3fffff00:  40240000 00000000 0000007d 3fff2d00  
3fffff10:  3ffed96a 3fff2d08 3fff73a4 40247af1  
3fffff20:  3fff2a0c 3fff6cbc 3fff6cbc 3fff0160  
3fffff30:  00000000 3fff73a4 0000001c 3fff6cbc  
3fffff40:  3ffed95c 00000000 3fff73a4 40246ee9  
3fffff50:  8001a8c0 00000064 00000000 00000031  
3fffff60:  00000002 0000001a 402366e7 3ffeeba0  
3fffff70:  3ffed934 3fffdcc0 3ffeb1b0 3ffeb1b0  
3fffff80:  4023665a 3ffeeba0 00000000 3fff6d04  
3fffff90:  3fffdc80 00000000 3fff73a4 4023eb5b  
3fffffa0:  40000f49 3fffdab0 3fffdab0 40000f49  
<<<stack<<<

 ets Jan  8 2013,rst cause:1, boot mode:(3,7)

load 0x4010f000, len 1384, room 16 
tail 8
chksum 0x2d
csum 0x2d
v09826c6d
~ld
[000172] 

```

When I request twice the espurna page (I think it's related to the GPIO pin 15)

tonilopezmr:
It is a serialization error comming from `ArduinoJson/Deserialization/../Serialization/JsonPirintable.hpp` in this line: https://github.com/xoseperez/espurna/blob/746ad7062dfdaa0a8c58e86d3fc0ad56fe040084/code/espurna/ws.ino#L389

code/espurna/config/hardware.h
```diff
@@ -3103,6 +3103,51 @@
...
+    // Light
+    #define LED1_PIN            0
+    #define LED1_PIN_INVERSE    1
+
+    // HJL01 / BL0937
+    #ifndef HLW8012_SUPPORT
+    #define HLW8012_SUPPORT             1
+    #endif    
+    #define HLW8012_SEL_PIN             12
+    #define HLW8012_CF1_PIN             13
```
mcspr: Comment here too that they are unused?

mcspr:
Based on gitter discussion, we may want to hide unused magnitudes.

HLW8012Sensor does not yet support this, but here is one example of dynamic magnitude list in the BMX280Sensor:
https://github.com/xoseperez/espurna/blob/0b73882fa0e286df96fb43daeb26d3e7f6ea4435/code/espurna/sensors/BMX280Sensor.h#L226-L232
So this can be HLW8012_CURRENT, HLW8012_VOLTAGE... etc.
In case we want to modify this at runtime, appropriate settings can be added (for example, `hlwVoltage` 1 or 0)

This implementation requires index counting  in type() and index() too, besides initial _count setting. Maybe magnitude list can be kept as an array / vector? MAGNITUDE_... are basic ints and you can even pass std::initializer_list as a func argument to set internal array just using a constructor / update method (*if* it is from std::... though)

code/espurna/config/hardware.h
tonilopezmr: yes, they are unused but I didn't know if I comment that then It will take another default pin that we don't want. If espurna will take default values for those SEL and CF1, we need to change it to the default unused pins we want, am I correct?

Here is an example: [SEL](https://github.com/xoseperez/espurna/blob/868d4d5573cabb33b3d50f88f7ff4ac2dd5b56db/code/espurna/config/sensors.h#L440) is by default one of the currently used pins

mcspr: Yeah, defaults are copying the first Sonoff POW, so something must be there. I was thinking of adding a comment, not removing the pin definitions. But, I think you can use `GPIO_NONE` (0x99) instead of the real ones. Arduino layer for digitalWrite, pinMode, attachInterrupt etc. has these two kinds of checks:
```
func(pin) {
  if(pin < 16){
  ...
  } else if(pin == 16){
  ...
  }
}
```
Or just `pin < 16`, so they will do nothing

ref:
https://github.com/esp8266/Arduino/blob/2.3.0/cores/esp8266/core_esp8266_wiring_digital.c
https://github.com/esp8266/Arduino/blob/master/cores/esp8266/core_esp8266_wiring_digital.cpp

tonilopezmr: Wow I didn't know that, I will change it for `GPIO_NONE`

PR #1826, opened 2019-07-14
Light updates

### LINKED ISSUE #1575: Magic Home LED Controller: white channel not independent anymore

mcspr:
Will likely rebase & merge this as-is, so adding all 3 commits
- fixes #1575: limit generic brightness function to only 3 channels when RGBWW color option is ON, but white channels are not used
- implement old TODO: move gamma to progmem
- test: print channel description in the terminal `channel <id>` (i'd also move this as webui description?)

@hyteoo https://github.com/xoseperez/espurna/commit/4702fcfb92726396542791a99300df0cadad8972 restores pre-1.13.3 behaviour when `useWhite` is off 
if `useColor` is also off, it will apply brightness to all the channels (as it does right now)

### DIFF
--- a/code/espurna/light.ino
+++ b/code/espurna/light.ino
@@ -70,8 +70,7 @@ const char _light_channel_desc[5][5] PROGMEM = {
 };
 
 // Gamma Correction lookup table (8 bit)
-// TODO: move to PROGMEM
-const unsigned char _light_gamma_table[] = {
+const unsigned char _light_gamma_table[] PROGMEM = {
     0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,
     1,   1,   1,   1,   1,   1,   1,   1,   1,   1,   2,   2,   2,   2,   2,   2,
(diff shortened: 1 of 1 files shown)


code/espurna/light.ino
```diff
@@ -93,73 +104,87 @@ void _setCCTInputValue(unsigned char warm, unsigned char cold) {
...
 
-void _generateBrightness() {
+void _lightApplyBrightness(unsigned char channels = lightChannels()) {
 
     double brightness = (double) _light_brightness / LIGHT_MAX_BRIGHTNESS;
 
-    // Convert RGB to RGBW(W)
-    if (_light_has_color && _light_use_white) {
+    for (unsigned char i=0; i < channels; i++) {
+        _light_channel[i].value = _light_channel[i].inputValue * brightness;
```
mcspr: Huh. I think I missed this interaction with the real values. If bound, it needs to do value = inputValue for all of the leftover channels

PR #1827, opened 2019-07-14
Support Shelly 2.5

### LINKED ISSUE #1698: Support for Shelly 2.5
Hi,
Is there a plan to support the new Shelly 2.5 in espurna?
This is the [product page](https://shelly.cloud/shelly-25-wifi-smart-relay-roller-shutter-home-automation) and this is the related [Tasmota issue](https://github.com/arendst/Sonoff-Tasmota/issues/5592). 
Thanks

tonilopezmr:
### :pushpin: References
* **Issue:** fixes #1698

### 💣 Issue

* Add new device [Shelly 2.5](https://shelly.cloud/shelly-25-wifi-smart-relay-roller-shutter-home-automation/) support following [Shelly 2.5 tasmota support](https://github.com/arendst/Sonoff-Tasmota/issues/5592).

### :tophat: What is the goal?

Integrate Shelly 2.5 with espurna to control the relay, control built-in led, read wall switch changes, get voltage, current, and temperature.

### :memo: How is it being implemented?

* Turn relay on/off, Relay 1 `GPIO4`, Relay 2 `GPIO15`
* Read switch state, Relay 1`GPIO13`, Relay 2 `GPIO5`
* Add new **ADE9753** sensor reading by I2C at address `0x38`, SDA `GPIO12` and SCL `GPIO14`
     - We get voltage, current relay 1 and current relay 2
     - I didn't use `i2c.ino` because I see only support for reading and write up to `uint16`, **Do I need to add `uint32` functions version?**.
     - Add three different instances of this sensor with a `_relay` configuration which could be `ALL`, `RELAY_1` or `RELAY_2` (this was my first attempt) to show (RELAY 1 + RELAY 2) consumption, RELAY 1 and RELAY 2. *(See on the picture below)*
* Use the built-in led in `GPIO0` with a reverse signal.
* Read the built-in temperature sensor using `NTCSensor`
* Read built-in button states for simple reset or factory reset using `GPIO2`

### ⚠️ Fix before merge / help wanted

* I didn't use `i2c.ino` and I would like to remove some compile warnings 💯 
* Different sensor instances 

This is how looks like `#0` are (RELAY 1 + RELAY 2) consumption, `#1` Relay 1, and `#2` Relay 2.

Screenshot 2019-07-14 at 23 42 59

> I would look at `pzem` module to know how to use a single instance.


mcspr:
> I didn't use i2c.ino and I would like to remove some compile warnings

Not seeing any warning with the latest commit 🤷‍♂ 
I would add 32 bit methods to the i2c.ino, to be consistent with other sensor implementations.

> Different sensor instances

Already discussed in gitter. You can use just one instance + some internal configuration. If you want to keep combined reading, this can be an optional flag (e.g. specific mode for per-relay readings, same + combined and only combined). As already mentioned in the #1822, BMX280 is one example of magnitude config. PZEM004T takes this further, using magnitude list multiple times for the reasons described here. 

IANAL, but I think the copyright comment should include you too? Maybe even with the comment below

tonilopezmr:
First of all, thanks for your review and your time. I would like to make those changes by myself then you just need to review it. I'm going to continue this PR tomorrow afternoon. 🎉

tonilopezmr:
Hi @mcspr I got an error trying to use vector to read and write the current and power values.

```
Exception Cause: 29  [StoreProhibited: A store referenced a page mapped with an attribute that does not permit stores]

0x40218f23: ADE7953Sensor::pre() at /Users/toni/Documents/sato/espurna/code/espurna/config/../sensors/ADE7953Sensor.h:130
 (inlined by) ADE7953Sensor::pre() at /Users/toni/Documents/sato/espurna/code/espurna/config/../sensors/ADE7953Sensor.h:114
0x40218ee0: ADE7953Sensor::pre() at /Users/toni/Documents/sato/espurna/code/espurna/config/../sensors/ADE7953Sensor.h:100
0x4021232c: _sensorPre() at /Users/toni/Documents/sato/espurna/code/espurna/sensor.ino:402
0x402378b6: wifi_get_opmode at ??:?
0x40216a1c: sensorLoop() at /Users/toni/Documents/sato/espurna/code/espurna/sensor.ino:1502
0x402378b6: wifi_get_opmode at ??:?
0x40238713: wifi_station_get_connect_status at ??:?
0x402282b8: JustWifi::connected() at /Users/toni/Documents/sato/espurna/code/.pio/libdeps/allterco-shelly25/JustWifi/src/JustWifi.cpp:796
0x4021a13c: ESP8266WiFiSTAClass::status() at /Users/toni/.platformio/packages/framework-arduinoespressif8266@1.20300.1/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp:453
0x40238713: wifi_station_get_connect_status at ??:?
0x402282b8: JustWifi::connected() at /Users/toni/Documents/sato/espurna/code/.pio/libdeps/allterco-shelly25/JustWifi/src/JustWifi.cpp:796
0x4021a13c: ESP8266WiFiSTAClass::status() at /Users/toni/.platformio/packages/framework-arduinoespressif8266@1.20300.1/libraries/ESP8266WiFi/src/ESP8266WiFiSTA.cpp:453
0x40206fd0: wifiConnected() at /Users/toni/Documents/sato/espurna/code/espurna/wifi.ino:569
0x40204e97: loop at /Users/toni/Documents/sato/espurna/code/espurna/espurna.ino:254 (discriminator 2)
0x4022df70: loop_wrapper() at /Users/toni/.platformio/packages/framework-arduinoespressif8266@1.20300.1/cores/esp8266/core_esp8266_main.cpp:123
0x401006fc: cont_norm at cont.S.o:?
0x4010f000: ?? ??:0
```

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,223 @@
...
+            );    
+            writeFloat(
+                ADE7953_RELAY_2, 
+                (float)current_rms2 / (ADE7953_IREF * 10), 
+                (float)active_power2 / (ADE7953_PREF / 10)
+            );
+        }
+
+        void writeFloat(unsigned int relay, float current, float power) {
+            float* current_reading = &_readings[relay].current;
```
mcspr: regarding exception, you sure readings array is initialized so [relay] access is usable? vector[] does not check bounds and neither is at(), since there are no exceptions.
btw, `auto& reading_ref = &_readings[relay]; reading_ref.current = current`? (may be fuzzy on the proper syntax)

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,223 @@
...
+            Ade7953Write(0x102, 0x0004);    // Locking the communication interface (Clear bit COMM_LOCK), Enable HPF
+            Ade7953Write(0x0FE, 0x00AD);    // Unlock register 0x120
+            Ade7953Write(0x120, 0x0030);    // Configure optimum setting       
+        
+            for(unsigned int i = 0; i > ADE7953_TOTAL_DEVICES ; i++) {
+                reading_t reading;
+                reading.current = 0.0;                
+                reading.power = 0.0;
+                reading.energy = 0.0;
+                _readings.push_back(reading);
```
mcspr: `_readings.resize(ADE7953_TOTAL_DEVICES);` should do the same thing.
i have vague understanding of what default struct values are in this case, but you can just bind these down below where `struct reading_t { ... };` is.

tonilopezmr:
Thanks! I have applied your changes and now it seems it's working! 

image

Now, what do I need to get those values in MQTT? As far as I know, I see [MQTT Sensors](https://github.com/xoseperez/espurna/wiki/MQTT#sensors) doesn't allow to read more than one `current` or `power` value because I don't see after the `root` topic a `0` for the first sensor or `1` for the second one.. etc...

tonilopezmr:
I don't know how to implement energy to save it, etc... I guess it should have something generic implemented to use, or every current metering sensor it's implementing the energy in its own way ??

mcspr:
ref https://github.com/xoseperez/espurna/pull/1832#issuecomment-513586301, `eneTotal<index>`?
PZEM does use a separate key storage though, so it might need to use a generic one too.

Sidenote: how expensive it would be to use double instead of float? I have noticed that pzem api uses it too.

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,215 @@
...
+        void _init() {                     
+            delay(100);                     // Need 100mS to init ADE7953
+            Ade7953Write(0x102, 0x0004);    // Locking the communication interface (Clear bit COMM_LOCK), Enable HPF
+            Ade7953Write(0x0FE, 0x00AD);    // Unlock register 0x120
+            Ade7953Write(0x120, 0x0030);    // Configure optimum setting       
+                    
+            _ready = true;
+        }
+
+        int Ade7953RegSize(uint16_t reg) {
```
mcspr: This might as well be static, since we don't use any member access. And we already are inside a class, Ade7953 prefix is redundant.

I was somewhat confused by the sizing function, is this mentioned somewhere in the datasheet?

*Kind-of related*
If it is really how it works, I tried making it more compact (idk about what is more readable though...)
```cpp
const int reg_size_new(const uint16_t reg) {

    const uint8_t mask = ((reg >> 8) & 0b11111111);

    if (!mask || (mask & 0b1100)) {
        return 1;
    } else if (mask & 0b0011) {
        return mask + 1;
    }

    return 0;

}
```
You can try this with ` ~/.platformio/packages/toolchain-xtensa/bin/xtensa-lx106-elf-gdb` (you'll need start building something with `espressif8266@2.2.2` platform first, it will move the old one to a separate dir)
```gdb
(gdb) disassemble reg_size
Dump of assembler code for function reg_size(unsigned short):
   0x40201018 <+0>:     extui   a2, a2, 8, 4
   0x4020101b <+3>:     movi.n  a3, 8
   0x4020101d <+5>:     bltu    a3, a2, 0x40201069 <reg_size(unsigned short)+81>
   0x40201020 <+8>:     l32r    a3, 0x40201014
   0x40201023 <+11>:    addx4   a2, a2, a3
   0x40201026 <+14>:    l32i.n  a2, a2, 0
   0x40201028 <+16>:    jx      a2
   0x4020102b <+19>:    and     a6, a3, a0
   0x4020102e <+22>:    excw
   0x40201031 <+25>:    excw
   0x40201034 <+28>:    or      a1, a0, a5
   0x40201037 <+31>:    and     a5, a5, a4
   0x4020103a <+34>:    excw
   0x4020103d <+37>:    excw
   0x40201040 <+40>:    s32i.n  a6, a0, 4
   0x40201042 <+42>:    excw
   0x40201045 <+45>:    excw
   0x40201048 <+48>:    excw
   0x4020104b <+51>:    and     a6, a3, a4
   0x4020104e <+54>:    excw
   0x40201051 <+57>:    s8i     a0, a6, 0
   0x40201054 <+60>:    andbc   b0, b12, b0
   0x40201057 <+63>:    addi.n  a2, a2, 1
   0x40201059 <+65>:    j       0x4020105e <reg_size(unsigned short)+70>
   0x4020105c <+68>:    movi.n  a2, 0
   0x4020105e <+70>:    addi.n  a2, a2, 1
   0x40201060 <+72>:    j       0x40201065 <reg_size(unsigned short)+77>
   0x40201063 <+75>:    movi.n  a2, 0
   0x40201065 <+77>:    addi.n  a2, a2, 1
   0x40201067 <+79>:    ret.n
   0x40201069 <+81>:    movi.n  a2, 0
   0x4020106b <+83>:    ret.n
(gdb) disassemble reg_size_new
Dump of assembler code for function reg_size(unsigned short):
   0x40202f24 <+0>:     extui   a3, a2, 8, 8
   0x40202f27 <+3>:     movi.n  a2, 1
   0x40202f29 <+5>:     beqz.n  a3, 0x40202f3c <reg_size_new(unsigned short)+24>
   0x40202f2b <+7>:     movi.n  a4, 12
   0x40202f2d <+9>:     and     a4, a3, a4
   0x40202f30 <+12>:    bnez.n  a4, 0x40202f3c <reg_size_new(unsigned short)+24>
   0x40202f32 <+14>:    extui   a2, a3, 0, 2
   0x40202f35 <+17>:    addi.n  a3, a3, 1
   0x40202f37 <+19>:    movnez  a4, a3, a2
   0x40202f3a <+22>:    mov.n   a2, a4
   0x40202f3c <+24>:    ret.n
```
At least what it seems like, that the code size goes down 77 bytes.

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,215 @@
...
+                size++;
+                case 0x00:
+                case 0x07:
+                case 0x08:
+                size++;
+            }
+            return size;
+        }
+            
+        void Ade7953Write(uint16_t reg, uint32_t val) {
```
mcspr: just `write`?
same thing about the static as above

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,215 @@
...
+                Wire.write(reg & 0xFF);
+                while (size--) {
+                Wire.write((val >> (8 * size)) & 0xFF);  // Write data, MSB first
+                }
+                Wire.endTransmission();
+                delayMicroseconds(5);    // Bus-free time minimum 4.7us
+            }
+        }
+            
+        uint32_t Ade7953Read(uint16_t reg) {
```
mcspr: `read`?
static?

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,215 @@
...
+                if (size <= Wire.available()) {
+                for (int i = 0; i < size; i++) {
+                    response = response << 8 | Wire.read();   // receive DATA (MSB first)
+                }
+                }
+            }
+            return response;
+        }
+
+    typedef struct {
```
mcspr: *just a style note*
because it is c++, this can be just `struct reading_t {}`

code/espurna/config/hardware.h
```diff
@@ -3122,14 +3122,8 @@
     #define RELAY1_PIN          15
     #define RELAY1_TYPE         RELAY_TYPE_NORMAL
 
-    // Light
-    #define LED1_PIN            0
```
mcspr: merge issue?

tonilopezmr: right! 

Just a note: `LED1_PIN` doesn't work for 1PM and 2.5 I don't know why, if I try to change LED state I can't use `gpio 0 1` or `gpio 0 0`, because the pin doesn't have pin mode output and the led state doesn't work.

code/espurna/sensors/ADE7953Sensor.h
tonilopezmr: Thanks! I'm not a c/c++ writer!

code/espurna/sensors/ADE7953Sensor.h
```diff
@@ -0,0 +1,215 @@
...
+                (float) active_power1 / (ADE7953_PREF / 10)
+            );    
+            writeFloat(
+                ADE7953_RELAY_2, 
+                (float)current_rms2 / (ADE7953_IREF * 10), 
+                (float)active_power2 / (ADE7953_PREF / 10)
+            );
+        }
+
+        void writeFloat(unsigned int relay, float current, float power) {
```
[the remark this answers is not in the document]
tonilopezmr: wow!! I didn't know about `inlined` in c++! thanks!

code/espurna/sensors/ADE7953Sensor.h
tonilopezmr: I need to learn more about moving bytes, I'm not really sure what reg_size does, but it seems to be a very important function.

code/espurna/sensors/ADE7953Sensor.h
tonilopezmr: I have problems with static methods because I'm using `_address` from `I2CSensor`. I'm going to add an extra parameter to send the address too

tonilopezmr:
Here you can play if you want with the new `reg_size` method (replacing it to the new one) and testing the declaration of `struct`, for me doesn't compile

code/espurna/sensors/ADE7953Sensor.h
tonilopezmr: doesn't compile for me! example: `'struct ADE7953Sensor::reading_t' has no member named 'current'`

mcspr:
Should work?
Second reg_size function is on by default, struct moved up.
I think the type warning fix for addresses is wrong though (uint8_t for all), but it should go to the i2c.ino anyways later.
delay / nice_delay is more of a quirk of a platform. it is not technically always needed, but nice to have in case it does break.

code/espurna/config/hardware.h
mcspr: https://github.com/xoseperez/espurna/blob/fbf2b71749307688f7258ad8235bc0f5d068e926/code/espurna/sensors/AnalogSensor.h#L72 can you remove this line?

tonilopezmr:
Thanks!! @mcspr I'm going to test it right now. About energy? 

1. is there any common implementation for all devices?
2. how I should implement it? 
3. It saves energy consumption if turn off and on?

mcspr:
_sensorEnergyTotal(index), _sensorEnergyTotal(index, value) ? see sensor.ino
It saves after reading N times, sensorLoop calls it every read to put it in persistent memory
(aka Rtcmem, does not erase between resets. Implemented as a pointer to that memory chunk and using it as basic struct)
Settings are pretty straightforward
Rtcmem I am not sure, keep it like this and only save index 0? Or, extend energy to energy0, energy1, etc. It needs to have known size on compilation

mcspr:
@tonilopezmr any luck so far? is energy saving still a problem?

tonilopezmr:
I sorry I was on holidays, yes, I need to implement it, also I need to fix a problem with led in GPIO0, it's not working, as you can see the configuration it seems right, but GPIO0 is not set as `OUTPUT`.

code/espurna/config/hardware.h
mcspr: See here. Perhaps it is leftover from some other arduino code, since ESP only has 1 analog input pin and it is already set as INPUT by the Core setup code. This is called via sensor setup code though...
Looks like every other hw that was used to test analog sensors was not using GPIO0

tonilopezmr: Okey! Thank you!

tonilopezmr:
I thought ` _sensorEnergyTotal(index, value)` exists, but I only see `void _sensorEnergyTotal(double value)` should I create it in `sensor.ino` ?

mcspr:
Sure, that's what I was thinking.

And idk why I was always referring to the PZEM sensor, [EmonADS1X15Sensor.h](https://github.com/xoseperez/espurna/blob/dev/code/espurna/sensors/EmonADS1X15Sensor.h) also supports multiple "ports" and is based on the EmonSensor / I2CSensor. It is analog converter, though... So, for example, rms value is manually calculated while ADE chip already does this for us.
As mentioned in gitter, I'd think we can somehow figure out a way to sync up readings with how fast ADE calculates data. What I was not sure about is how exactly make this happen, when documentation only mentioned interrupt pins and per-channel state registers as a method to receive state updates.

It has the same problem with energy saving by the looks of it, since there's no mechanism to restore energy per port, and current API is misleading for single source devices (ref HLW, first argument is value and not idx).

code/espurna/sensors/ADE7953Sensor.h
mcspr: And just to be clear about this, what happens is:
```python
>>> '{:#06b}'.format(0x31C >> 8)
'0b0011'
>>> '{:#06b}'.format(0x21C >> 8)
'0b0010'
```
Then, when checking last 2 bits
```
>>> (0b0011 & 0b0011) + 1
4
>>> (0b0010 & 0b0011) + 1
3
```
So, the 0x3... prefix contains 4 bytes => 32 bits and 0x2... contains 3 bytes => 24 bits

mcspr:
Nice. BTW, while I was describing that energy accumulator, I missed that it needs to be explicitly enabled via ACCMODE, LCYCMODE and LINECYC registers. As I understood it, then the AENERGY... registers will contain per-cycle values and would be immediately reset to 0, waiting for the next cycle.
Have you tried reading those manually via cli commands or periodically as part of sensor read, maybe just for a test? Energy value could be more precise that way (depends on load though)

Calibration is also missing, that is also something to add later. See A/B WGAIN, A/B IGAIN 

Does makes sense to save the first "dev" at all? I'll clean up flow a bit, combining with other sensors to avoid having special conditional handling per sensor type. Maybe also finally fix reset timestamping...

tonilopezmr:
I think for a first "dev" version it's fine, I have been testing this weekend and works fine, I would like to improve my knowledge about I2C and ADE, I would create an ISSUE to work on getting energy from ADE sensor, with this comment's help.  

can you give me more information about reading those values using cli commands? do you mean espurna commands? 

Thanks!

mcspr:
Yes, terminal. Something to inspect sensor state, which in this case is to read configuration and value registers. `i2c.read <bits> <addr>` ?

code/espurna/sensor.ino
```diff
@@ -462,18 +462,37 @@ void _sensorResetTS() {
     #endif
 }
 
-double _sensorEnergyTotal() {
+double _sensorEnergyTotal(unsigned int index = -1) {
```
mcspr: I missed this earlier... note the `unsigned` :) Fixed via #1875

tonilopezmr: Thank you!

Issue #1380, opened 2018-11-30
NtpClientLib issues / alternatives


mcspr:
Based on #682 discussion. Some possible options, not dependent on Core version:
- [ ] use `configTime()` from Core / lwip (possibly reuse now() / year() etc. with appropriate wrappers above libc gmtime())
- [ ] directly fix fork of NtpClient / import it's code here, per [gmag11 comment](https://github.com/xoseperez/espurna/issues/682#issuecomment-427324096) (no need to switch to AsyncUDP, just adjust internals. wip + problems described below need to be addressed)
- [ ] replace with adapted [ezTime](https://github.com/ropg/ezTime) fork (not usable directly as a lib right now, settings are hardcoded)

`configTime()` has offsets with core 2.3.0, proper time zones since 2.4.0, but having it configured on that level is still problematic (ref ntp.ino:`ntpLocal2UTC()`, `NtpClient::decodeNtpMessage()`). Instead, timezone support could be added outside of ntp client, based on the work done in #1295.
Timestamp can be kept as UTC and local time can be applied whenever it is appropriate
`ntpDST` / `ntpOffset` / `ntpRegion` can still be read but proper support should be using TZ variables / otherwise stringified timezone offsets array

While LWIP option is being the easiest to implement, sntp does not return much info besides timestamp being set initially. One reason to keep ntp client on Arduino level is to have access to the connection status / timeouts etc., although those issues are not really actionable.

<summary>*Some additional info* <br>tldr: packet validation is wrong, DST affects sync timestamps, DNS blocking</summary>
<details>

- No udp packet sanity check / kiss-o-death packet. Sync is treated as a success by checking udp packet length, which will not detect problems with packet itself. Also, quoting https://tools.ietf.org/html/rfc5905#page-76
> A kiss-o'-death packet is an NTP header with leap 0x3 (NOSYNC) and
> * stratum 16 (MAXSTRAT). It tells the receiver that something
> * drastic has happened, as revealed by the kiss code in the refid
> * field. The NTP header fields may or may not be correct.
- Multiple raised issues when it is crashing from either ws or mqtt (sys context inside async callbacks), due to TimeLib directly doing network request when now() is called.
- Faulty DNS will stall NTP.getTime() method until the DNS timeout comes
- .begin() immediately trying to sync time
- DST implementation is overriding received timestamp right in the udp packet, having no obvious way 
to receive it untouched (set daylight to false, sync, back to true?)

</details>

Misiu:
[...]

PR #1828, opened 2019-07-17
Set RTC to same timezone and daylight settings upon boot

Niek:
This sets the RTC to the same timezone and daylight settings as the NTP config. This fixes that ``time(nullptr)`` returns the same time before and after NTP sync.

### DIFF
--- a/code/espurna/ntp.ino
+++ b/code/espurna/ntp.ino
@@ -98,6 +98,9 @@ void _ntpConfigure() {
         _ntp_report = true;
     }
 
+    // Configure internal clock (RTC) with same timezone and daylight settings
+    configTime((tz_hours * 3600) + (tz_minutes * 60), (daylight * 3600), "");
+
     String server = getSetting("ntpServer", NTP_SERVER);
     if (!NTPw.getNtpServerName().equals(server)) {
         NTPw.setNtpServerName(server);


mcspr:
Why though? configTime just enables sntp loop & timer, which will do nothing bc there is no server set (besides setting timer and wasting some time in timer func).

AFAIK, this may be different between lwip and lwip2, and lwip2 uses TimeLib-like approach and not the real RTC memory time

Niek:
If you call ``time(nullptr)`` before NTP is synced, it will return the system clock time in wrong timezone (I think it's always China time). It looks like ``configTime()`` is the most compatible way to set the timezone of the system clock before NTP is synced, but maybe there's another way?

mcspr:
I mean, why use it at all? It is not synced anyways using the lwip's sntp and we increase binary size too, because of the way linkage works.

BTW, have you tried with lwip2 builds? (removing -DPIO_FRAMEWORK_ARDUINO_LWIP_HIGHER_BANDWIDTH from the build flags). Would it not just return 0 there?

mcspr:
To answer my own question, time(0) results are:
- 0 (constant) with lwip1.4 builds
- 28800 initially and ticking up every ~1s with lwip2 builds (see `realtime_stamp` in cores/esp8266/sntp-lwip2.cpp)

Initially I thought that there was a way to set this outside of sntp code, but does not seem like it :/

Niek:
I tested this with LWIP 2 (PIO_FRAMEWORK_ARDUINO_LWIP2_LOW_MEMORY_LOW_FLASH). The result of ``time(nullptr)`` (on both older cores as well as staging) are:
- Current time in UTC +7h before NTP sync is completed (right after boot)
- Current time in UTC after NTP sync

The reason it's useful to get the RTC time in the correct timezone is that you don't need to wait on NTP sync to complete in order to connect a BearSSL/CA client (e.g. to MQTT). In most cases the RTC will be close to synchronized with the NTP clock so you can speed up the initial connect with a few seconds (a wrong time will results in a failing connection).

mcspr:
Sorry for being a bit dense, but do you mean that you have it synced between resets somehow? AFAIK it just keeps a global var and does not do network sync. configTime is a separate thing from TimeLib / NtpClientLib.

And it is not RTC per-se, since ESP8266 does not keep time there, but some internal timekeep timer thing that Core devs added.
We can, however, use additional Rtcmem slot to save timestamp there on each sync (or, better off, via ticker every second). That way, you will have up-to-date time whenever the device is rebooted

Niek:
I understand that's it not a (battery powered) RTC in the traditional sense, but it serves the same purpose (keep clock state between reboots).

Only after the NTP sync is done, the system time (from ``time(nullptr)``, I think ``now()`` from TimeLib returns the same) will be in the correct timezone. Maybe configTime is not the best way to set the timezone, I see ``RTC_TZ_SET`` and ``RTC_DST_SET`` are used internally.

mcspr:
> If you call time(nullptr) before NTP is synced, it will return the system clock time in wrong timezone (I think it's always China time). 

I am missing one thing still - how does the time from NtpClientLib gets into the time() results? Whenever time() is called in our context, it should always return seconds from boot (i.e. 0) plus timezone offset and not the real timestamp.

Looking at the routines that you mention, they are only relevant for lwip1.4 code and those RTC register writes are only triggered when `sntp_time_flag` is `true`. But, it is always `false`:
https://github.com/esp8266/Arduino/blob/29bedfa8429100f3ed0d5a960bd2520c9ebf15e2/tools/sdk/lwip/src/core/sntp.c#L305
https://github.com/esp8266/Arduino/blob/29bedfa8429100f3ed0d5a960bd2520c9ebf15e2/tools/sdk/lwip/src/core/sntp.c#L1167

mcspr:
Ok. So there's one way that this may happen:
https://github.com/d-a-v/lwIP/blob/159e31b689577dbf69cf0683bbaffbd71fa5ee10/src/apps/sntp/sntp.c#L800
SNTP server list is populated by the DHCP callback, which is always enabled for LWIP2 builds:
https://github.com/esp8266/Arduino/blob/2130f3ee8c98b5b79be71031ad6cf8f9b62da18b/tools/sdk/lwip2/include/lwipopts.h#L954
First one of which you erase by using configTime "" arg, right after wifi connection finishes:
wifiConnected -> ntpLoop continues -> ntpConfigure called first time -> configTime bam

time() will return valid timestamp from the initial sync, but it will never be updated after that (if that what is really happening, this also is causing really funny issue accessing ntp server twice in a row)
Debug build of lwip + wireshark will tell for sure.

Niek:
Wow good find, I was scratching my head why I couldn't reproduce the issue anymore at a different location - ``time(nullptr)`` suddenly always returned 0 from boot time. Apparently it depends on the DHCP server. I guess this PR can be closed?

mcspr:
Yep, given that this is not doing the intended thing.

This does incentivize revisiting #1380 again, and maybe just dropping NTPClient when lwip2 is detected. Minimal port would be to add timezone conversion rules and minimal TimeLib compatibility layer.
There is also `sntp_getserver` to check if dhcp did something to the sntp server list

Niek:
Ok closing this, I agree that #1380 would probably be the better approach

d-a-v:
@mcspr Why don't you PR on lwip2 repo ?

mcspr:
I have not tested anything yet 🙃 And comments also suggest to check whether arithmetic works ok. Plus, there's also requirement to set the default time closer to today's date
(*"in order for the round-trip calculation to work, the difference between the local clock and the NTP server clock must not be larger than about 34 years"*)

And there are also some other things I wanted to try first, like customizable sntp_set_time (or esp8266/arduino settimeofday) callback a-la IDF ([commit](https://github.com/espressif/esp-idf/commit/7e5be1b58cab05cb9c45f06d519f8845175938c8) / [discussion PR](https://github.com/espressif/esp-idf/pull/1668) / [forum discussion](https://esp32.com/viewtopic.php?f=13&t=4833))

d-a-v:
(Arduino has the time_is_set-like callback)

mcspr:
Yes, but I was looking for *any* reason to make time_is_set cb itself overridable. The way commit above chooses to either use the default, setting time directly via settimeofday or through adjtime (which is another interesting feature). Or it can be overriden by the user app to avoid doing that at all, working as a notification of sntp success instead of system time change

Niek:
https://github.com/esp8266/Arduino/pull/6373 is now merged in core: https://github.com/esp8266/Arduino/commit/ffe5476fc42b06ebde5688527200c6690e8e519f

PR #1829, opened 2019-07-17
MQTT rewrite with SSL fixes (see also: #1751)

Niek:
- Added Arduino MQTT library support (actively maintained)
- Added support for BearSSL (core >= 2.5)
- BearSSL validation: insecure, fingerprinting and CA validation
- AxTLS validation: insecure and fingerprinting
- Support MFLN in order to reduce heap usage

This PR kind of depends on #1751, might be better to wait for that one to be merged first. Update: #1751 is merged now.


mcspr:
Quick questions:
- any particular reason not to use DEBUG_MSG_P?
- can we reduce the size of connect method, avoiding the nested ifdef tree?
- should #1751 implement common setup method for wificlient?
- why combine async server dependencies and secure_client_axtls?

Niek:
> Quick questions:
> 
> * any particular reason not to use DEBUG_MSG_P?

I'm using ``DEBUG_MSG`` instead of ``DEBUG_MSG_P`` if the debug message consist of static text only - please let me know if this doesn't make sense.

> * can we reduce the size of connect method, avoiding the nested ifdef tree?

I agree, it's getting quite messy for sure - I can split up in a separate WifiClient and AsyncMqttClient one if you want? It'll be a bit more verbose obviously and have some duplication, but the code will be more readable.

> * should #1751 implement common setup method for wificlient?

I think it makes sense, but if we want things like separate MFLN/CA/etc settings per module it will be hard to make it a common method. And there may be more exceptions: for example, we might always want to attempt MFLN on MQTT but not on OTA.

> * why combine async server dependencies and secure_client_axtls?

I don't understand this question, can you elaborate? When ``MQTT_LIBRARY == MQTT_ASYNC``, it's using ESPAsyncTCP and thus AxTLS by definition, since you can't use BearSSL with ESPAsyncTCP.

mcspr:
You can just use PSTR always. For any debug message it makes sense to place those there, since they will be kept in heap otherwise; after esp loads the binary, all char strings are placed in heap (unless specified 
 otherwise with PROGMEM macro). And only the format is in PROGMEM anyways.
*sidenote:* but, with 2.5.0+ (not sure on the exact version, since https://github.com/earlephilhower/newlib-xtensa/commit/61f08d03e875a64629deaaa0b975e179bcfa4ea0. unless it was broken on release), libc was patched so that printf will accept PSTR's as arguments too, simply always using aligned reads for `%s`

> It'll be a bit more verbose obviously and have some duplication, but the code will be more readable.

Duplication can be minimized through? At the very least, maybe some flow can be split up into functions. If marked as explicitly inline, binary code would be exactly the same, but (i hope) more readable.

> I think it makes sense, but if we want things like separate MFLN/CA/etc settings per module it will be hard to make it a common method. And there may be more exceptions: for example, we might always want to attempt MFLN on MQTT but not on OTA.

When I get around to updating the HTTPupdate PR, I will try to implement sort of a wrapper for wificlient struct and helpers for some of the flow stages (i.e. here we are checking mfln, here we are setting up security checks). Some minimal config, giving cert pointer and fp string for example.

> I don't understand this question, can you elaborate? When MQTT_LIBRARY == MQTT_ASYNC, it's using ESPAsyncTCP and thus AxTLS by definition, since you can't use BearSSL with ESPAsyncTCP.

I meant changing `ASYNC_TCP_SSL_ENABLED && WEB_SSL_ENABLED` parts.

Niek:
I have cleaned up the code a bit, using PSTR everywhere and made the connect method a lot more concise (still not spit up though). Having the WiFiClient(Secure) wrapper would help a lot in cleaning up the code I think.

> > I don't understand this question, can you elaborate? When MQTT_LIBRARY == MQTT_ASYNC, it's using ESPAsyncTCP and thus AxTLS by definition, since you can't use BearSSL with ESPAsyncTCP.
> 
> I meant changing `ASYNC_TCP_SSL_ENABLED && WEB_SSL_ENABLED` parts.

Ah yes, I probably should have put that in a separate PR - I assume that after introducing ``SECURE_CLIENT`` in the codebase we should get rid of all ``ASYNC_TCP_SSL_ENABLED`` ifdefs in the code, hence those changes.

code/espurna/mqtt.ino
```diff
@@ -128,56 +153,143 @@ void _mqttConnect() {
...
+                if (requested_mfln) {
+                    bool supported = _mqtt_client_secure.probeMaxFragmentLength(_mqtt_server.c_str(), _mqtt_port, requested_mfln);
+                    DEBUG_MSG_P(PSTR("[MQTT] MFLN buffer size %u supported: %s\n"), requested_mfln, supported ? "YES" : "NO");
+                    if (supported) {
+                        _mqtt_client_secure.setBufferSizes(requested_mfln, requested_mfln);
+                    }
+                }
+                #endif
+
+                int check = getSetting("mqttScCheck", SECURE_CLIENT == SECURE_CLIENT_BEARSSL ? SECURE_CLIENT_CHECK_CA : SECURE_CLIENT_CHECK_FINGERPRINT).toInt();
```
lbdroid: The way this is organized prevents BearSSL from using a signature check. It would be preferable to be able to configure this.

Niek: This is just the default setting: on BearSSL that's the CA check and on AxTLS that's the fingerprint (basically the most secure setting on both implementations). Setting ``mqttScCheck`` to ``SECURE_CLIENT_CHECK_FINGERPRINT`` (1) will allow you to do a fingerprint check.

lbdroid: Oh, I see. So presumably, that will have to be brought out to the web UI at some point.

Thanks.

Niek: Web UI configuration is a good point. What do you think @mcspr, should users be able to upload the certificate which is then written on the SPIFFS? Or just keep a toggle to switch the ``mqttScCheck`` setting (none/FP/CA)?

lbdroid: According to discussion in https://github.com/xoseperez/espurna/issues/1839 , it is considered appropriate for all configuration options to be made available via defines.

I understand the idea to be that the defines... Define the DEFAULTS, whereas your implementation enforces a predetermined default.

Niek: Hmm yes, I agree that it would be better to have that configurable via a define as well. It's just that I copied the behavior from PR #1751 ([1](https://github.com/xoseperez/espurna/pull/1751/files#diff-4668235fd2c44f88c437348e5247399aR101) [2](https://github.com/xoseperez/espurna/pull/1751/files#diff-5727c97832f80d3ac6aedd50dbe75fd6R84
) [3](https://github.com/xoseperez/espurna/pull/1751/files#diff-5727c97832f80d3ac6aedd50dbe75fd6R152
)).

But yeah, that means yet another define (``MQTT_SECURE_VALIDATION`` or something) in general.h. As well as another one for the OTA code (``OTA_SECURE_VALIDATION``). It would be great to have something like JSON/ini config file, as mentioned in https://github.com/xoseperez/espurna/issues/1839#issuecomment-518656529.

Niek:
Web UI is still TBD as per comment https://github.com/xoseperez/espurna/pull/1829#discussion_r311912469

mcspr:
Regarding my previous comment about the fixed flash location for cert data, this is exactly what happened in tasmota aws iot integration:
Special conversion .pem -> .der -> translated key -> base64: https://github.com/arendst/Sonoff-Tasmota/wiki/AWS-IoT#step-8-prepare-your-aws-iot-credentials
Ofc that required base64 converter and some hard-coded (probably temporary) assumptions about the flash layout: https://github.com/arendst/Sonoff-Tasmota/pull/6179

mcspr:
Little follow up on client configuration, based on this branch.
https://github.com/xoseperez/espurna/compare/dev...mcspr:mqtt-ssl-rewrite
I wanted to test it first though... 

Gist of what's changed:
- include headers are in prototypes, .ino has the objects
- remove some logging from runtime into compile time
- split connect method into two stages: setup and async/sync_connect
- use the same mfln logic as httpupdate does (see `mfln.probe` terminal command for probing)
And SecureClientHelpers do the rest. Since we want to have up-to-date config, there's new "config" struct with all of the settings as getters. "...scCheck" logic is separated in to "checks" struct, running all of the relevant actions.
I also tried to utilize those for AsyncMqttClient, but I am not sure how to clearly separate those, since  Could also just forget about AxTLS altogether and focus on bssl integration everywhere. Notice a little *optimization* with onconnect.

*Continuing off-topic about the certs...* 
I had also noticed that we can override X509List class with something that returns progmem pointers on getTrustAnchors(), so there's no need to create a whole separate wificlientsecure class. After `make` on bearssl-esp8266 repo, `brssl ta file.pem` util can generate the required struct, only thing needed is to add PROGMEM attribute.

PR #1830, opened 2019-07-20
Add Sonoff Itead S31 Lite Hardware Profile

CrazyIvan359:
This adds a hardware profile for the S31 "Lite" version, which does not have the power monitoring chip installed on the board. The device is otherwise identical. It uses the exact same board and components as the regular S31.

I tried using the regular S31 profile, but Espurna would not boot because it could not open communications with the CSE7766 chip. The Lite profile simply removes the CSE7766 support and re-enables debug serial (might as well since we don't need the quiet for the CSE7766).

IMG_20190720_092306


mcspr:
Huh. This is exactly the same pin layout as Sonoff Basic:
https://github.com/xoseperez/espurna/blob/fbf2b71749307688f7258ad8235bc0f5d068e926/code/espurna/config/hardware.h#L279-L286
...or even going back to the Sonoff SV (minus second button)

Adding for convenience sake, but we need to handle these duplicates at some point

CrazyIvan359:
@mcspr what about a "base" hardware profile and addon profiles with only the extra options needed. Then in `platformio.ini` you make the build env with `-DSONOFF_BASE -DSONOFF_S31` for example, then the S31 light would only be `-DSONOFF_BASE`

tching911:
HI every body.
I just received my s31 lite sonoff and the connections are different from yours. Someone knows which pins p02 p03 or p04
are equivalent to tx and rx
Thank's
sonoff s31 lite

Issue #1529, opened 2019-02-06
Feature Request Sonoff Basic WiFi 


peteeee:
Originally posted this in Gitter and it was suggested that I post it here.

Currently using a SonOff WiFi basic / Tasmota in multiple function mode as a Garage door sensor, opener and have added one DS18B20 sensor for temperature only monitoring. GPIO 4, 14 and a pulse for GDO. (IE: door open sensor, door closed sensor, push button pulse to open garage door and temperature sensor) It has worked fine over the last few months. 

Over the last few months though have switched from Tasmota to Espurna firmware.

Automation sand box here using Homeseer MQTT, Home Assistant, Openhab, Domoticz, and Homegenie. (continue to utilize X10, UPB, Zigbee and ZWave). Personally like the Espurna firmware a bit better than Tasmota firmware. Where would I make a request for same functionality using Espurna firmware versus Tasmota custom firmware? I noticed it is much easier to solder modifications to the SonOff WiFi basic board versus the Magic Home RGB controller (which I am currently using with customized Espurna firmware).

In a recap the firmware would be multi functional for use with the modded SonOff WiFi basic device.
Hardware modifications to testing SonOff WiFi basic:
1 - cut high voltage traces for relay to use low voltage
2 - soldered pins to RX, TX, Ground, GPIO 4, GPIO14. (GPIO 1,3,4,14)
3 - connected GPIO 3 to one door sensor, GPIO 4 to second door sensor, GPIO 14 to DS18B20.

Using MQTT to monitor status of GPIO 3 and GPIO4 and GPIO 14 and toggle relay for garage door opener.

This is a low cost WiFi basic combination garage door opener, door monitor and temperature sensor.

![sonofftasmota](https://user-images.githubusercontent.com/8072890/52348760-ba86e700-29ea-11e9-8d7a-c9b337f472ce.jpg)

lbdroid:
So correct me if I'm wrong, but it looks like your objective is just to be able to read the state of the GPIO pins over MQTT, is that correct?

Try this: (which I am using for a garage door opener)
```
diff --git a/code/espurna/config/hardware.h b/code/espurna/config/hardware.h
index 6068a274..82ce69d7 100644
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -259,19 +259,63 @@
     // Info
     #define MANUFACTURER        "ITEAD"
     #define DEVICE              "SONOFF_BASIC"
[...]

PR #1832, opened 2019-07-21
Enable multiple Digital & Events

pilotak:
I know that there is going to be support for specifying all the things dynamically through browser but before that here is enhancement to support it now. This would also help for anyone finding #1815 as there is no code provided, only guidance

### DIFF
--- a/code/espurna/config/defaults.h
+++ b/code/espurna/config/defaults.h
@@ -553,6 +553,92 @@
 #endif
 
 // -----------------------------------------------------------------------------
+// Digital Inputs
+// -----------------------------------------------------------------------------
+
+#ifndef DIGITAL1_PIN
+#define DIGITAL1_PIN         GPIO_NONE
+#endif
+#ifndef DIGITAL2_PIN
+#define DIGITAL2_PIN         GPIO_NONE
+#endif
+#ifndef DIGITAL3_PIN
+#define DIGITAL3_PIN         GPIO_NONE
+#endif
+#ifndef DIGITAL4_PIN
+#define DIGITAL4_PIN         GPIO_NONE
+#endif
+#ifndef DIGITAL5_PIN
+#define DIGITAL5_PIN         GPIO_NONE
+#endif
+#ifndef DIGITAL6_PIN
+#define DIGITAL6_PIN         GPIO_NONE
+#endif
+#ifndef DIGITAL7_PIN
+#define DIGITAL7_PIN         GPIO_NONE
+#endif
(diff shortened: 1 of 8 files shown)


pilotak:
The only thing i don't know is how to set `pwrRatioE` in web-config for each sensor separately

mcspr:
`pwrRatioE<index>`? where index is just a _sensors position
Reset & calibration are a strange ones. Wouldn't it be easier to use direct action calls instead of going through settings and then purging keyvalues?

You might want to also add mapping from the old `..._PIN` to the `...1_PIN`

pilotak:
`pwrRatioE<index>` is not a problem to add but how to hide the unused in browser - there is only section hiding (in html there will be all 8 listed but lets say you have only two setup which would need `display: inherit`).
How do i do "direct action calls"?

mcspr:
Append html settings section programmatically?

By direct actions I meant ws actions:
https://github.com/xoseperez/espurna/blob/fbf2b71749307688f7258ad8235bc0f5d068e926/code/espurna/relay.ino#L698
https://github.com/xoseperez/espurna/blob/fbf2b71749307688f7258ad8235bc0f5d068e926/code/html/custom.js#L430
(thus directly requesting required actions instead of using one-shot settings)

pilotak:
ok, what if i remove pulse-meter from this PR and create a separate once it's done?

mcspr:
sure, that will work

mcspr:
> You might want to also add mapping from the old ..._PIN to the ...1_PIN

Btw, by that I meant to add something here https://github.com/xoseperez/espurna/blob/dev/code/espurna/config/deprecated.h
to remap old defines into the new one + show a warning. Release notice will be there to, but at least VSCode / Atom with PlatformIO will track those messages and show them after build completes if someone uses those in custom header

pilotak:
@mcspr any other request? i think it can be changed otherwise

mcspr:
Don't think so. Any other changes can be made in bulk with other sensors (and maybe even move sensor init to sensor classes themselves)
Sorry for the delay, was I bit busy past 2 weeks.

Also, cc @lbdroid mentioned this https://github.com/xoseperez/espurna/issues/1529#issuecomment-517808484

pilotak:
That's ok, can this be merged?

mcspr:
Done.
I have noticed some pulsemeter changes left over, but they just chain back into themselves and don't change the existing behaviour

pilotak:
I know let's call them preparation for future

PR #1836, opened 2019-07-25
 * Added support for Etekcity ESW01-USA device

sirpatil:
Adding support for a new device. Tested on a device and basic functionality works ok.


mcspr:
Thanks! If this needs any notes about opening / flashing it, we can add a wiki page for it.

And I have also added amazon link for what seems to be this switch model.

PR #1837, opened 2019-07-31
Add support for the ISelector SM-PW702

lwalkera:


mcspr:
Thanks!

Is this it btw? https://www.amazon.com/ISELECTOR-Function-Required-Compatible-Assistant/dp/B075XL3DRD

PR #1841, opened 2019-08-06
Allow offline building - skip package updater if there is no connection

Niek:
A quick fix to allow building if there is no internet connection. I couldn't build the latest HEAD during a flight, I think the package update should be optional and not break building.

### DIFF
--- a/code/extra_script_pre.py
+++ b/code/extra_script_pre.py
@@ -66,9 +66,12 @@ def get_shared_libdeps_dir(section, name):
 
 
 def ensure_platform_updated():
-    if PIO_PLATFORM.are_outdated_packages():
-        print("updating platform packages", file=sys.stderr)
-        PIO_PLATFORM.update_packages()
+    try:
+        if PIO_PLATFORM.are_outdated_packages():
+            print("updating platform packages", file=sys.stderr)
+            PIO_PLATFORM.update_packages()
+    except Exception:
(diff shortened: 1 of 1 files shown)


mcspr:
I tried to remember what this did exactly, besides fixing lwip build... so maybe it is better to just replace  LWIP -> LWIP2 in the build flags and drop this upgrade thing. espressif8266 platform forces old toolchain for frameworks other than `arduino` but never forces the latest one due to the relaxed version pin.

It still needs to be high bandwidth for fauxmoesp, since the lib does not do buffering by itself
ref: https://bitbucket.org/xoseperez/fauxmoesp/issues/76/adding-more-than-10-devices

mcspr:
Another side effect of the update_packages() is that when env is using `platform = ${common.arduino_core_git}`, it does `git pull` on the framework dir (with submodules)
Somewhat useful for CI (e.g. add travis04 with it), but I'd rather it did not update locally.
gcc-9 adds interesting warnings btw, dtostrf for example was using `1-sizeof(buffer)` as a decimals space, which should've been `decimals + 2`

Niek:
That's how I noticed the issue (the git command throws an exception). CI always starts with a fresh environment anyway, right?

I haven't managed to build the GCC9 toolchain on macOS yet, but interested to see if that leads to improvements in code size/heap usage.

PR #1852, opened 2019-08-12
Core 2.6.0+ fixes

mcspr:
- travis04 to test build with current esp8266/Arduino master
- lwip -> lwip2
- small fix for gcc-9 compilation


mcspr:
@Niek the `template` addition fixes build with gcc-7 & gcc-9

for tests:
`~/.platformio/packages/framework-arduinoespressif8266@src-31d658a59f41540201fc3726a1394910` with checked out https://github.com/earlephilhower/Arduino/tree/gcc9.1
manually fetch https://github.com/earlephilhower/esp-quick-toolchain/releases/tag/3.0.0-gnu1 and unpack into the `~/.platformio/packages/toolchain-xtensa-91`
copy `~/.platformio/packages/toolchain-xtensa/package.json` and modify the version. e.g.
```patch
--- toolchain-xtensa/package.json       2019-08-13 00:02:16.052773600 +0300
+++ toolchain-xtensa-91/package.json    2019-08-13 00:01:44.126792900 +0300
@@ -3,5 +3,5 @@
     "name": "toolchain-xtensa",
     "system": "linux_x86_64",
     "url": "https://github.com/earlephilhower/esp-quick-toolchain",
-    "version": "2.40802.190218"
+    "version": "3.0.220719"
 }
```

Niek:
Nice, LWIP2 as default makes sense as well. I only seem to have some crashes on AP mode with git core, but that might be some of my own code - will investigate further.

Did you notice any difference in build size and heap usage when building with 9.1?

mcspr:
> Did you notice any difference in build size and heap usage when building with 9.1?

With the same debug POW build it is ~5.5KB less in size. Heap metric... around the same. Initial value changes though - 43296 bytes with the current toolchain and 42808 bytes with gcc-9 (ofc everything is built with the same gcc9.1 tree)

PR #1858, opened 2019-08-15
Fixed a bug, any unauthenticated user could reset the device

foxman69:
If you just access /reset the device reboots without any warning or need to authenticate


mcspr:
This one and /upgrade were pretty strange thing to have... I wonder if a single handler can dispatch all simple requests to avoid this

PR #1862, opened 2019-08-16
Fix js sync bug that can cause brick

foxman69:
if you restore configuration (not rebooting the device) and change more settings it makes the device freak out and sometimes even bricks it (wont boot anymore).
making sure to restart after configuration restore fixes it.

Tested on Sonoff 4CH PRO R2


mcspr:
>  it makes the device freak out and sometimes even bricks it (wont boot anymore).

Does this happen with any config file, even with minimal one? settingsRestoreJson expects "app":"ESPURNA", you can try adding wifi settings and check that out.
so you would need to flash via serial again, like dead - dead? can you compare full flash dump before and after? check out the wiki here and esptool.py repo readme

PR #1865, opened 2019-08-19
Check value range for PMSX005 and SenseAir CO2 sensor

Yonsm:
Check value range for PMSX005 and SenseAir CO2 sensor, and drop invalid value (temperature>=100°C or humidity<=1% or humidity>=100%) for MQTT reporting.


mcspr:
Thanks!

PR #1873, opened 2019-08-26
Updates for #1829

mcspr:
https://github.com/xoseperez/espurna/pull/1829#issuecomment-524671908
@Niek tested with https://test.mosquitto.org/ - test.mosquitto.org:8883, https://test.mosquitto.org/ssl/mosquitto.org.crt as custom .pem root
I also noticed that we may need to specify it clearer that it is *not a client* certificate, but root cert of the server. `_mqtt_client_ca` -> `_mqtt_client_server_ca` ?

- clean-up dependencies and headers
- since MQTT_LIBRARY choices are always commented to explain what they mean, use full names for those and add MQTT_LIBRARY_... prefix
- remove some logging from runtime, move some into compile time
- split connect method into two stages: setup and async/sync_connect
- use the same mfln logic as httpupdate does (see mfln.probe terminal command for probing)

And SecureClientHelpers do the rest. Since we want to have up-to-date config, allow module to define "config" struct with all of the settings as getters. "...scCheck" logic is separated in to internal "checks" struct, running all of the relevant actions based on "config" results.
AsyncMqttClient is still special case, because it handles fingerprinting all by itself. Difference from `dev` is that fingerprint only added on module setup, not with every configuration reload.
Also notice a little optimization with onconnect verification for axtls. No need to overcomplicate things if it will be removed anyway.


code/espurna/mqtt.ino
```diff
@@ -17,41 +18,38 @@ Copyright (C) 2016-2019 by Xose Pérez <xose dot perez at gmail dot com>
...
+#if SECURE_CLIENT != SECURE_CLIENT_NONE
+    SecureClient* _mqtt_client_secure = nullptr;
+#endif
 
-    #if MQTT_LIBRARY == MQTT_ARDUINO // Using Arduino-MQTT
-        #include <MQTTClient.h>
-        #ifdef MQTT_MAX_PACKET_SIZE
+#if MQTT_LIBRARY == MQTT_LIBRARY_ARDUINOMQTT
+#ifdef MQTT_MAX_PACKET_SIZE
         MQTTClient _mqtt(MQTT_MAX_PACKET_SIZE);
```
mcspr: And what is up with this btw. Does platformio build_flags need comment / bigger value for it? idk if json works at all with the default 400 bytes buffer

mcspr:
Obviously, httpupdate should use this method.
Thingspeak sync variant too, since I do see SSL as an option.

Later merge of ESPAsyncTCP with proper bssl support would not need anything with before/after logic, only thing what would be required is to have some kind object with the same API as WiFiClientSecureBearSSL (setTrustAnchors, setBufferSizes etc.) that would be somehow exposed on the target client. Template magic will properly handle the rest

Niek:
Very nice work :) I'll leave some comments on the code, but this looks good - I'll do some testing on my end as well. I agree the naming can be a bit better, for example ``_mqtt_client_server_ca`` as you mentioned and the and ``MQTT_LIBRARY_`` prefix like in this PR.

mcspr:
Renamed those to `...trusted_root_ca`, both headers and symbols.

Issue #1864, opened 2019-08-17
HSV command is not HSV set


CrazyIvan359:
**Bug description**
HSV commands do not apply the H, S, and V in the command.

**Steps to reproduce**
Send an HSV command to a device with Color enabled.

**Expected behavior**
HSV set should be the same as the command.

**Device information**
* Arduino Core `2.3.0 rev 159542381`
* ESPurna version `1.13.5`

**Solution**
The "Value" of the command is being used as the brightness, then the calculated RGB channel values are scaled by it. This should not be done, as the conversion from HSV to RGB takes the V component into consideration already.

mcspr:
Can you give an example of expected values?

Following the brightness function, it does some other transformations when "useWhite" mode is on. However, basic brightness method will simply multiply each channel by the brightness, resulting in the same values as this conversion shows:
https://www.wolframalpha.com/input/?i=0,100,57+hsv+to+rgb
https://github.com/xoseperez/espurna/blob/43a44cf96447dc2e04f0ef936573950df359889e/code/espurna/light.ino#L1212
https://github.com/xoseperez/espurna/blob/43a44cf96447dc2e04f0ef936573950df359889e/code/espurna/light.ino#L117

Extracted light.ino algorithm:
https://gist.github.com/mcspr/4a0a7001c9b52c0dd09ca1af13b287c5
https://github.com/xoseperez/espurna/blob/43a44cf96447dc2e04f0ef936573950df359889e/code/espurna/light.ino#L274-L276
```
$ ./hsv 0,100,57
brightness=0, rgb=0,0,0
input=0,100,57
_setRGBInputValue(255,0,0)
brightness=145, rgb=255,0,0
```
*Note about Arduino.h => std::...*
`constrain` => `std::clamp`
`round` => `std::lround`
`floor` => `std::floor`

CrazyIvan359:
I'm running the latest release, not dev. I fixed the issue by preventing the brightness from being calculated when an HSB command is received.

I think if you were to use the maximum Brightness value (100% not `MAX_BRIGHTNESS`) when calculating the RGB values, and then set actual brightness normally, the `_light_brightness_func` should adjust the values to what they should be. I ran out of time to keep testing different solutions though.

I do see that another issue I found has been fixed now, I saw that channels 4 and 5 would be set by brightness if colour was enabled, regardless of white settings.

mcspr:
Ping. Was #1874 sufficient?

CrazyIvan359:
[...]

Issue #1763, opened 2019-06-07
Domoticz H801 RGBWW


physke:
Hi,

I've managed to get espurna on a H801 working with domoticz thru MQTT for RGB and brightness settings, but not for warm/cold white LEDs. Looks like domoticz is sending cmode=2, but there is nothing in the espurna code for cmode=2. only 3 & 4 (ColorModeRGB & ColorModeCustom). Is this an issue with domoticz or espurna or the way I have it configured?

**domoticz setting for CW/WW:**
![image](https://user-images.githubusercontent.com/35822297/59090011-aaaeed80-895f-11e9-9864-8d0bbaa31b27.png)

**received by espurna:**
![image](https://user-images.githubusercontent.com/35822297/59090162-214beb00-8960-11e9-9ec8-d8a5743da8c2.png)

**domoticz setting for RGB:**
![image](https://user-images.githubusercontent.com/35822297/59090230-49d3e500-8960-11e9-9df4-ff553b374c51.png)

**received (and processed) by espurna:**
![image](https://user-images.githubusercontent.com/35822297/59090334-9b7c6f80-8960-11e9-9611-262cb4b36c9a.png)

Thanks
Jason

physke:
;(

mcspr:
As you have noted, domoticz module specifically filters out any other cmode:
https://github.com/xoseperez/espurna/blob/274f71d7d9047bb2bdf6b5c2cca7a6ff9d63142b/code/espurna/domoticz.ino#L61

While we do see the problem... what is the difference between cmodes? Should we care at all about them when receiving values?
I could not find anything about them
https://www.domoticz.com/wiki/MQTT
https://www.domoticz.com/wiki/Special:AllPages
...and google searches in general, since I do not use Domoticz

Only reference is by reading ColorSwitch.h, counting from 0 this is cmode=2:
https://github.com/domoticz/domoticz/blob/22ee3f8b37158f20a62e345d410055a5cd9bd250/hardware/ColorSwitch.h#L18-L28
> ColorModeTemp,     // White with color temperature. Valid fields: t

physke:
Found this:

br, = brightness 0..100
cw, = cold white Range:0..255, Cold white level
ww, = warm white Range:0..255, Warm white level
m, = mode
ColorModeNone = 0, // Illegal
ColorModeWhite = 1, // White. Valid fields: none
ColorModeTemp = 2, // White with color temperature. Valid fields: t
[...]

PR #1874, opened 2019-08-26
Format fixes in light module

mcspr:
- configurable max and min values
- fix printf format specifiers with `double` inputs
- fix `map()` with older cores, add warning
- maybe fix hsv conversion? #1864


code/espurna/light.ino
```diff
@@ -382,11 +384,10 @@ void _toRGB(char * rgb, size_t len) {
...
 void _toHSV(char * hsv, size_t len, bool target) {
     double h, s, v;
-    double brightness = (double) _light_brightness / LIGHT_MAX_BRIGHTNESS;
 
-    double r = (double) ((target ? _light_channel[0].target : _light_channel[0].inputValue) * brightness) / 255.0;
-    double g = (double) ((target ? _light_channel[1].target : _light_channel[1].inputValue) * brightness) / 255.0;
-    double b = (double) ((target ? _light_channel[2].target : _light_channel[2].inputValue) * brightness) / 255.0;
+    double r = static_cast<double>(target ? _light_channel[0].target : _light_channel[0].value);
+    double g = static_cast<double>(target ? _light_channel[1].target : _light_channel[1].value);
+    double b = static_cast<double>(target ? _light_channel[2].target : _light_channel[2].value);
```
mcspr: I noticed this strange calculation earlier. `inputValue` is `value` without brightness, so maybe it makes sense in that case. But it tries adjust it even with `target` which is set from `value`, thus displayed value could be wrong.
@CrazyIvan359 does checking for `target` flag makes any sense here at all? Web sends it with `false`, but mqtt reports with `true`... I am not quite sure what was the reason for this conditional.

mcspr: OK! Thanks

`bool target` param for `_toHSV(output, output_size, >> target <<)` function, selecting `channel.target` aka value+brightness via ternary 'if' above.

Last change to this was 19d0c5d17a576a22e8b7d38c1f09e3a14e41d6a5, which states to report target (adjusted) values to mqtt & web, but that only happened with mqtt (see `lightColor(false)` usage).
Maybe there's no need for conditional at all and it should always use adjusted value

mcspr:
Adding for https://github.com/xoseperez/espurna/issues/1763, to not forget to use lightMap(...)

PR #1875, opened 2019-08-26
Use indexed keys for energy saving

mcspr:
- always use indexes with `eneTotal`
- fix `unsigned int` comparison with -1
- fix sensorID checks, use `magnitude.global` index directly as index
- migrate `pzemEneTotal` to `eneTotal`
- migrate `eneTotal` to `eneTotal0`


mcspr:
@tonilopezmr and I had also for some reason thought that `magnitude.global` is the same as `_magnitudes` index. Instead, it sets itself from `_counts[magnitude.type]`. All eneTotal indexes are updated from that value now

PR #1877, opened 2019-08-27
Test: etharp_gratuitous

### LINKED ISSUE #614: Sonoff Dual R2 IP Address not reachable after a while
To avoid mixing up what may be two different issue that I noted at the bottom of [issue #572](https://github.com/xoseperez/espurna/issues/572) I thought it better to open an new issue here.

I have managed to do some more testing - having finally managed to lose connection to the web /telnet interface. I think my issue may be related to the closed [issue #392](https://github.com/xoseperez/espurna/issues/392).

Briefly my setup is an unmodified Technicolor TG582n 4 port wireless router connected to a couple of devolo "powerline type" wifi extenders. These extenders are ultimately connected to one of the four Ethernet ports. The sonoff Dual r2 is in the same room as the router less than 2m away from it. Signal strength to the router almost maxed out. The nearest devolo is across the hall at the far end of the room - signal strength almost non existent. I use a shared (all use the same SSID) and hidden SSID.

Logging into the router web interface I am able to see the devices connected to its interfaces - wifi /ethernet port. When I am able to connect to the Sonoff it always appears attached to the routers wifi interface. When I can't connect to the sonoff it always appears on the Ethernet port. As the Sonoff is wifi it has to be assumed appearance on the ether port has to be because it has decided to attach to the devolo AP - with almost no signal strength in preference to the router with maximum strength! When I next lose connection I will see if I can log onto the devolo and see if it registers the sonoff as a device connected to it.

I am not sure why even if the sonoff is connected to the devolo AP I can't access the Sonoff web interface. A tablet connected to the same devolo wifi extender, is able to browse the internet, but yet not able to connect to the Sonoff when it goes off line. Conversely when the Sonoff comes back on line (connected to the router wifi) the tablet is then able to access the sonoff web interface while attachd to the same devolo AP.

As previously noted leaving Telnet enabled in the ESPurna Web interface, and repeatedly initiating a telnet connection will in most cases bring the Sonoff back online. The last time I had to run nping --arp <sonoff IP> to get the Sonoff to talk to me.

Wifi on the Sonoff is configured with Scanning turned off. With scanning turned on it is unable to connect to the hidden SSID.

Have I missed something I can do to stop the sonoff hopping onto a different AP? Alternatively, if it can be exposed on the web interface, is there a way to force the sonoff to only bind to a specific BSSID?

mcspr:
fix #614 ?
https://github.com/esp8266/Arduino/issues/5998#issuecomment-525332517

@arihantdaga can you try this one? if I understood correctly the required process. Can be adjusted a bit later
*edit:* not seeing any resets on initial connection. wifi.reset, wifi.ap commands work too.

If you can reliably reproduce lost IP addresses (either modem or none sleep), it would be interesting to see if this helps

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -565,13 +565,13 @@
 #define WIFI_GRATUITOUS_ARP_SUPPORT              1
 #endif
 
-// Interval is randomized on each boot in range from ..._MIN to ..._MAX
+// Interval is randomized on each boot in range from ..._MIN to ..._MAX (ms)
 #ifndef WIFI_GRATUITOUS_ARP_INTERVAL_MIN
-#define WIFI_GRATUITOUS_ARP_INTERVAL_MIN         15
+#define WIFI_GRATUITOUS_ARP_INTERVAL_MIN         15000
 #endif
 
 #ifndef WIFI_GRATUITOUS_ARP_INTERVAL_MAX
-#define WIFI_GRATUITOUS_ARP_INTERVAL_MAX         30
+#define WIFI_GRATUITOUS_ARP_INTERVAL_MAX         30000
 #endif
 
 // -----------------------------------------------------------------------------
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -554,6 +554,26 @@
 #define WIFI_PROPAGATION_CONST      4           // This is typically something between 2.7 to 4.3 (free space is 2)
 #endif
 
+// ref: https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/kconfig.html#config-lwip-esp-gratuitous-arp
+// ref: https://github.com/xoseperez/espurna/pull/1877#issuecomment-525612546 
+//
+// Help solve compatibility issues with some routers.
+// If the ARP table of the AP is old, and the AP doesn't send ARP request to update it's ARP table,
+// this will lead to the STA sending IP packet fail. Thus we send gratuitous ARP periodically to let AP update it's ARP table.
+
+#ifndef WIFI_GRATUITOUS_ARP_SUPPORT
+#define WIFI_GRATUITOUS_ARP_SUPPORT              1
+#endif
+
+// Interval is randomized on each boot in range from ..._MIN to ..._MAX
+#ifndef WIFI_GRATUITOUS_ARP_INTERVAL_MIN
+#define WIFI_GRATUITOUS_ARP_INTERVAL_MIN         15
+#endif
+
+#ifndef WIFI_GRATUITOUS_ARP_INTERVAL_MAX
+#define WIFI_GRATUITOUS_ARP_INTERVAL_MAX         30
+#endif
+
 // -----------------------------------------------------------------------------
 // WEB
 // -----------------------------------------------------------------------------
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -557,9 +557,8 @@
 // ref: https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/kconfig.html#config-lwip-esp-gratuitous-arp
 // ref: https://github.com/xoseperez/espurna/pull/1877#issuecomment-525612546 
 //
-// Help solve compatibility issues with some routers.
-// If the ARP table of the AP is old, and the AP doesn't send ARP request to update it's ARP table,
-// this will lead to the STA sending IP packet fail. Thus we send gratuitous ARP periodically to let AP update it's ARP table.
+// Broadcast gratuitous ARP periodically to update ARP tables on the AP and all devices on the same network.
+// Helps to solve compatibility issues when ESP fails to timely reply to ARP requests, causing the device's ARP table entry to expire.
 
 #ifndef WIFI_GRATUITOUS_ARP_SUPPORT
 #define WIFI_GRATUITOUS_ARP_SUPPORT              1
--- a/code/espurna/wifi.ino
+++ b/code/espurna/wifi.ino
@@ -95,8 +95,6 @@ void _wifiConfigure() {
         _wifi_gratuitous_arp_interval = getSetting("wifiGarpIntvl", secureRandom(
             WIFI_GRATUITOUS_ARP_INTERVAL_MIN, WIFI_GRATUITOUS_ARP_INTERVAL_MAX
         )).toInt();
-
-        DEBUG_MSG_P(PSTR("[WIFI] garp interval=%u\n"), _wifi_gratuitous_arp_interval);
     #endif
 
 }
--- a/code/espurna/wifi.ino
+++ b/code/espurna/wifi.ino
@@ -711,15 +711,10 @@ void wifiSetup() {
 
     // Periodic gratuitous arp to keep the ip alive
     #if WIFI_GRATUITOUS_ARP_SUPPORT
-
-        #if WIFI_GRATUITOUS_ARP_INTERVAL_MIN == WIFI_GRATUITOUS_ARP_INTERVAL_MAX
-            _wifi_gratuitous_arp_interval = getSetting("wifiGarpMax", WIFI_GRATUITOUS_ARP_INTERVAL_MAX).toInt();
-        #else
-            _wifi_gratuitous_arp_interval = secureRandom(
-                getSetting("wifiGarpMin", WIFI_GRATUITOUS_ARP_INTERVAL_MIN).toInt(),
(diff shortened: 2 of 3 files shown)


arihantdaga:
@mcspr Its working. Dint see any reset. I also checked at the same time, n->ip_addr.addr was giving value 0. I think this condition for checking netif up or not works well. 👍

TD-er:
Looks OK to me.
I will also include it in ESPeasy to see if this also may have been triggering some of the WDT reboots.

TD-er:
For ESP32 at least, you need to check the `hwaddr_len`, or else it will crash with an assert as soon as AP+STA is enabled.

So it will be something like this:
```
    if ((n->hwaddr_len == ETH_HWADDR_LEN) && 
        (n->flags & NETIF_FLAG_ETHARP) && 
        ((n->flags & NETIF_FLAG_LINK_UP) || (n->flags & NETIF_FLAG_UP))) {
```

I am not sure if a check for `NETIF_FLAG_ETHARP` flag is needed. (or maybe even contra-productive?)
See [LWIP NETIF_FLAG_](http://www.nongnu.org/lwip/2_0_x/group__netif__flags.html)

mcspr:
...UP check should be && btw, assuming lwip authors know something that we don't and expecting that LINK_UP can be missing independently of UP
https://git.savannah.nongnu.org/cgit/lwip.git/tree/src/core/netif.c?h=STABLE-2_1_2_RELEASE#n880
Old lwip1 code only has UP condition

ETH flag is set by lwip-glue layer, so check is safe with 8266 too:
https://github.com/d-a-v/esp82xx-nonos-linklayer/blob/af6738a22332a92256471e3d29d37a2700ec1298/glue-lwip/lwip-git.c#L397
https://github.com/espressif/esp-idf/blob/fdab15dc76a3464ed10ce03e9b9cf1a5cd2aa0d1/components/lwip/port/esp32/netif/wlanif.c#L76
Probably can be avoided, but you can never be too safe.

And while I read the https://espeasy.readthedocs.io/en/latest/Tools/Tools.html#periodical-send-gratuitous-arp explanation for arp, in what case it should be on? Are we doing this so that router does not kick us, sort-of keep alive?
Is it safe to leave always on? Is 5000msec interval something based on testing or can be more random?

TD-er:
I have been using the code I mentioned in my last comment, since that comment.
The ESP32 unit I have here is still very responsive now. It used to be taking up to a minute to reply after some time.
The ESP8266 ones do not appear to react different from before this change.

It is not really a keep-alive, but more to make sure the MAC tables in switches/AP/router are not forgetting about the node.

It is giving an answer to a question never asked.

For normal ethernet traffic, each switch/router etc. needs to know what MAC address is behind what port.
Since these tables in switches etc. are not infinite large, they have to renew their tables every now and then.
This is done via ARP packets.
So someone asks "who has 192.168.1.1" and the answer is also used by every other host or switch on the network to update its MAC tables.
If the ESP does not reply to such ARP requests, it may become impossible to send data (back) to the ESP, since the switches do not know how to route it.
So therefore we send Gratuitous ARP packets to answer "Mac AA:BB:CC:DD:EE:FF has 192.168.1.1" even though no-one asked for it.

Sending ICMP (ping) packets to the ESP may also have a similar effect, since ICMP packets appear to be handled different by the AP compared to other packets.
Also an ICMP packet does seem to have effect on the power management of the ESP at core level.
Try running an ESP node for a while (running `delay()` when not performing a task) and you will notice the power consumption does decrease after a few minutes.
This lower power consumption does seem to be correlated with not receiving some packets like ARP packets.
However, if you start sending ping packets to the ESP, the first one may take a while (up-to 900 msec sometimes), but it will always be answered. This will cause the ESP to increase its power consumption and the next ping packets will be answered much faster.

This simple test does show us a number of things happening here:
- Lower power consumption is very likely caused by the WiFi radio not listening on every beacon interval of the AP (102.4 msec on most AP)
- Some packet types (ARP/UDP/etc) are not being repeated or buffered by the AP to overcome the actual listening interval of the ESP. (at least the ones not required to be answered)
- ICMP packets are being repeated long enough to be received by the ESP (N.B.  even the first ping you send will get an reply eventually)
- ESP nodes may become hard/impossible to reach after a while, which can be explained by their MAC address not being known any more. You can also test this on your PC to look for the known ARP table, or clearing it yourself.

TL;DR
I don't think the Gratuitous ARP is the same as a keep-alive, but more like prevent to become forgotten.

In ESPeasy I use a more dynamic interval for sending these ARP packets.
If a connection (or DNS lookup) attempt fails, I will make the interval short (and send one right away) and every new call to Gratuitous ARP will increase the interval to about 5 sec again.
Also right after getting connected ("Got IP" event) I send one out.
The actual implementation of clearing MAC tables in switches may differ from brand to brand.
Some may remove one after not been seen for a while and others just may clear the entire table periodically (or when it is full). I have seen reports of 30 seconds, but also lots of "hard to give a single number".
It is hard to know what implementation is being used, but I guess some interval below the watchdog timer may be a good idea.
Maybe also good to have some kind of randomness in the intervals to prevent all nodes sending out ARP packets at the same time (e.g. after some power outage)

mcspr:
Wow, thanks for detailed explanation!

It is surprising that esp32 build works differently, there's this exact functionality enabled by default:
https://github.com/espressif/esp-idf/blob/aa087667dffb8f024a5bf162cef4d5c0ffb152ab/components/lwip/Kconfig#L138-L154
https://github.com/espressif/arduino-esp32/blob/7a574399b17e552800459e3eafb72d9158c05968/tools/sdk/sdkconfig#L601 (i.e. pre-built idf has this flag on)
https://github.com/espressif/esp-lwip/blob/61d840ff4778f4946c8743f7e412345abcd537f1/src/core/ipv4/etharp.c#L142-L154
I guess, 60s is too long?

I think I'd settle with idf approach and randomize arp time between 15 and 30s (and maybe kick in sooner if any connection routine fails, that's an interesting idea)

code/espurna/config/general.h
```diff
@@ -554,6 +554,26 @@
 #define WIFI_PROPAGATION_CONST      4           // This is typically something between 2.7 to 4.3 (free space is 2)
 #endif
 
+// ref: https://docs.espressif.com/projects/esp-idf/en/latest/api-reference/kconfig.html#config-lwip-esp-gratuitous-arp
+// ref: https://github.com/xoseperez/espurna/pull/1877#issuecomment-525612546 
+//
+// Help solve compatibility issues with some routers.
+// If the ARP table of the AP is old, and the AP doesn't send ARP request to update it's ARP table,
+// this will lead to the STA sending IP packet fail. Thus we send gratuitous ARP periodically to let AP update it's ARP table.
```
TD-er: Strictly speaking, it is not the sending of an packet that fails, but receiving.
The other hosts in the network may no longer know how to retrieve what MAC address to use as destination when sending a packet to some IP-address. Also the other components on the network, like switch or AP may have the MAC address removed from their tables and thus no longer be able to route the packet to the correct port.

code/espurna/wifi.ino
```diff
@@ -687,31 +739,11 @@ void wifiLoop() {
...
-        if ((n->flags & NETIF_FLAG_LINK_UP) || (n->flags & NETIF_FLAG_UP)) {
-            DEBUG_MSG_P(PSTR("sending arp...\n"));
-            etharp_gratuitous(n);
-            res = true;
-        } else {
-            DEBUG_MSG_P(PSTR("netif not yet up...\n"));
+    #if WIFI_GRATUITOUS_ARP_SUPPORT
+        // Only send out gra arp when in STA mode
+        if (_wifi_gratuitous_arp_interval && ((WiFi.getMode() & WIFI_AP) == 0)) {
+            _wifiSendGratuitousArp(_wifi_gratuitous_arp_interval);
```
TD-er: With STA+AP mode the WiFi radio is always on, which makes it less likely ARP requests may be missed.
But it may still be useful to send it if you get a (new) IP address.
Just make sure to only send it to the STATION_IF if you want to distinguish this.

N.B. the AP and STA use different MAC addresses, so it should be no problem to send out Gratuitous ARP to the AP interface. (as long as you're sending the right IP of course :) )

code/espurna/config/general.h
mcspr: The wording is weird, true... that was a blatant c/p from the esp-lwip Kconfig entry. Reworded to mention ARP tables.

code/espurna/wifi.ino
mcspr: This part is copied from the idf, basically trying to emulate GARP flag that they check. No flag here, so STATION_IF seems like a solution.

PR #1878, opened 2019-08-28
fix paths in prototypes.h

reaper7:
Arduino on windows machine shows this error:
```
In file included from sketch\config/all.h:36:0,

                 from C:\Programy\arduino_projekty\espurna_dev\code\espurna\espurna.ino:22:

sketch\config/prototypes.h:178:86: fatal error: static/letsencrypt_isrgroot_pem.h: No such file or directory

 #include "static/letsencrypt_isrgroot_pem.h" // Default to LetsEncrypt X3 certificate

                                                                                      ^

compilation terminated.
```
replace include paths from **static/** to **../static/** fix problem

### DIFF
--- a/code/espurna/config/prototypes.h
+++ b/code/espurna/config/prototypes.h
@@ -173,9 +173,9 @@ using mqtt_callback_f = std::function<void(unsigned int, const char *, char *)>;
 #endif
 
 #if MQTT_SECURE_CLIENT_INCLUDE_CA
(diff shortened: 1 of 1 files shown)


mcspr:
Thanks! I'd note that this is cross-platform issue, IDE / arduino-builder does things differently from platformio builder. For example, I do see the same issue with Linux build via arduino-cli
This should probably go into .ino, like WebUI headers... will add into the #1873

Issue #1440, opened 2018-12-19
Depreciated warnings from home-assistant


inverse:
In the last major update of home-assistant I noticed the following warnings in the logs

```
"platform": "mqtt" is deprecated, replace with "schema":"basic"
```

From researching the issue I realise It's to do with the mqtt message with auto-discovery that will need to be adjusted. 

Having no expeirence with this code base would I be right in the assumption that it involves changing this line? 

https://github.com/xoseperez/espurna/blob/dev/code/espurna/homeassistant.ino#L82

From:
```
    config.set("platform", "mqtt");
```

To: 
```
    config.set("schema", "basic");
```

mcspr:
Yep, it seems to be working just fine with {"schema":"basic"} addition. At first I thought this was only related to lights (searching ha github about light.mqtt_json changes), but switch component works the same.

Not really sure when deprecated becomes obsolete. No mention of it here:
https://github.com/home-assistant/home-assistant/blob/a9990c130dc613d83a63128375767458e52f4937/homeassistant/components/mqtt/discovery.py#L43
https://github.com/home-assistant/home-assistant/pull/18227 (pr that changed this)

inverse:
@mcspr thanks for adding more context :+1:

inverse:
@xoseperez that's a very valid point. I think generally people move with the releases but supporting both would be a good solution for a while and will cover people like me who keep up and when they do finally remove the feature not be broken. What do you think?

peterhoeg:
I have filed a feature request with HA for having it publish either its own version or an API version via MQTT. That would allow espurna to detect the format to use.

mcspr:
@peterhoeg Have you posted the issue to the home-assistant/home-assistant about mqtt version? I would not rely on versions for this, but it might be nice to display that ESPurna device found HA presence.

"schema" / "platform" can also be avoided altogether as it falls back to the basic mqtt method (on / off / state topics) and implicitly passes "platform": "mqtt" to the configurator.
There is no json lights support -> no need to specify it right now. And no warning.
Future version must use "schema": "json" (which is from config spec, not discovery specific) to use json payloads

peterhoeg:
@mcspr, I did on discourse as that's where they prefer feature requests: https://community.home-assistant.io/t/publish-ha-version-to-mqtt/86694

I has not received any replies though.

Issue #1085, opened 2018-07-28
Support publishing custom value for relay processing


InduPrakash:
It would be nice if on/off can be sent instead of 1/0. Incoming on/off messages are handled but only 1/0 get sent out. This will make interfacing and automating with Smartthings easier. I had prototyped introducing MQTT_TOPIC_RELAY_ON/MQTT_TOPIC_RELAY_OFF with default value 1/0 and custom value in hardware.h, and that seemed to work well.

xbb1024:
hi, do you have a template smartthings device handler for espurna? I'm new to mqtt.

InduPrakash:
SmartThings events have to through a SmartApp/DeviceHandler into MQTT world through a bridge (https://hub.docker.com/r/stjohnjohnson/smartthings-mqtt-bridge/) to interact with Espurna device. My device is just of the type "Virtual Switch". The ST MQTT messages just end with on/off to indicate the state as well as the action and I was able to get this working by defining "smartthings/Device Name/switch" as the grouper, this felt quite odd but works and adjusting switch in ST gets reflected in Espurna device. But now the other way around since the outgoing messages have 1/0. So I locally end up creating 2 defines MQTT_TOPIC_RELAY_ON/MQTT_TOPIC_RELAY_OFF and used it in homeassistant.ino and relay.ino. I can request a pull if you want to look at it.

xoseperez:
You can use "ON" and "OFF" (and "TOGGLE") for incoming MQTT messages. If you also want it to output strings instead of 0/1 you can use the latest dev branch and change the RELAY_MQTT_ON and RELAY_MQTT_OFF settings in `general.h`

InduPrakash:
Thank you, that is pretty much what I did. In addition, I also modified
homeassistant._haSendSwitch to use the same.

xoseperez:
Good idea, but I'd prefer to use a different setting for that.

PR #1885, opened 2019-09-02
Home Assistant updates

### LINKED ISSUE #1762: show config topic field blank when code uploaded from platformio
The show config in HASS is not working properly when i am uploading code from platform io instead of using the prebuilt firmware for nodemcu lolin.

```
switch:
  - name: ESPURNA_1F7B48
    platform: mqtt
    state_topic: 
    command_topic: 
    payload_on: 1
    payload_off: 0
    availability_topic: 
    payload_available: 1
    payload_not_available: 0
 ```

This is what i am getting where as you can see state_topic, command_topic and availability_topic is blank

### LINKED ISSUE #1883: MQTT relay payload ON / OFF inconsistent usage
**Bug description**
Configuration options for payload ON and OFF strings are independent of each other. For example, if Home Assistant configuration uses payload ON as "1", relay payload needs to be updated too. Otherwise, configuration generator would show a different value from the one that is actually sent.

**Steps to reproduce**
Update HOMEASSISTANT_PAYLOAD_ON
Run `ha.config` in terminal, see `payload_on`
Try to subscribe to the relay topic and toggle the relay. Actual value is from RELAY_MQTT_ON

**Expected behavior**
Dependant MQTT messages should have a single configuration entry
Perhaps, there should be a global setting for every "boolean" payload.

* *ESPurna version* 35ce68710cc51fae2e185a6d5edff31a6d035002

**Tools used**
* *IDE & version* Platformio, Arduino IDE with custom.h / modified .h
* *Compiler & version (if not embedded in IDE)* Any

**Additional context**
As mentioned in the https://github.com/xoseperez/espurna/issues/1762#issuecomment-526709359, if HA config generator omits payload_on/off, Home Assistant will expect default ON and OFF strings. However, these are used:
https://github.com/xoseperez/espurna/blob/dev/code/espurna/config/general.h#L1280-L1294
But they need to be in sync with these:
https://github.com/xoseperez/espurna/blob/35ce68710cc51fae2e185a6d5edff31a6d035002/code/espurna/config/general.h#L407-L413
Because that's what is actually used by mqtt send in relay module:
https://github.com/xoseperez/espurna/blob/35ce68710cc51fae2e185a6d5edff31a6d035002/code/espurna/relay.ino#L879
So there should also be an option to optionally omit certain key-value pairs in the configuration.

mcspr:
- remove "platform" key, see #1440. this implicitly sets schema to "basic". pending some other clean-up regarding json and mqtt queueing, other schema can be added down the line 
- updated ws module queue elem to capture callbacks list, allows to pass more than one callback (for example, when they are generated on the fly as lambdas, see ha wsPost usage)
- modified method to send ha config to use global ws queue, fix #1762 problem with empty topics and ensure json allocation is consistent.
- use existing defines to set mqtt payload options. amend https://github.com/xoseperez/espurna/issues/1085, https://github.com/xoseperez/espurna/pull/1188 to use the set payload value. drop HOMEASSISTANT_PAYLOAD... defines. fixes #1883 
- update MQTT_STATUS_ONLINE/OFFLINE and RELAY_MQTT_ON/OFF with runtime configuration
- filter payload strings so that resulting yaml value is not interpreted as bool
- helper method for settings to streamline string values manipulation

### DIFF
--- a/code/espurna/config/deprecated.h
+++ b/code/espurna/config/deprecated.h
@@ -25,3 +25,20 @@
 #warning EVENTS_PIN is deprecated! Please use EVENTS1_PIN instead
 #define EVENTS1_PIN EVENTS_PIN
 #endif
+
+// 1.13.6 unifies mqtt payload options
+#ifdef HOMEASSISTANT_PAYLOAD_ON
+#warning HOMEASSISTANT_PAYLOAD_ON is deprecated! Global RELAY_MQTT_ON is used instead
+#endif
+
+#ifdef HOMEASSISTANT_PAYLOAD_OFF
+#warning HOMEASSISTANT_PAYLOAD_OFF is deprecated! Global RELAY_MQTT_OFF is used instead
+#endif
+
+#ifdef HOMEASSISTANT_PAYLOAD_AVAILABLE
+#warning HOMEASSISTANT_PAYLOAD_AVAILABLE is deprecated! Global MQTT_STATUS_ONLINE is used instead
+#endif
+
+#ifdef HOMEASSISTANT_PAYLOAD_NOT_AVAILABLE
+#warning HOMEASSISTANT_PAYLOAD_NOT_AVAILABLE is deprecated! Global MQTT_STATUS_OFFLINE is used instead
+#endif
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -408,6 +408,7 @@
 #ifndef RELAY_MQTT_ON
 #define RELAY_MQTT_ON               "1"
 #endif
+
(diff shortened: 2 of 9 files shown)


mcspr:
> -use existing defines to set mqtt payload options. amend #1085, #1188 to use the set payload value. drop HOMEASSISTANT_PAYLOAD... defines. fixes #1883 
> - update MQTT_STATUS_ONLINE/OFFLINE and RELAY_MQTT_ON/OFF with runtime configuration

cc @Yonsm, as that was the latest change to the status payload string. now, only one define needs to be specified

PR #1889, opened 2019-09-03
MQTT: handle custom relay status payloads

mcspr:
- customize TOGGLE payload
- match payload string when receiving data
- reference enum values instead of raw integers, spell out intended status
- remove dead code

amend #1885, capitalize `relayPayload...` suffix instead of using uppercase
add `relayPayloadToggle`

### DIFF
--- a/code/espurna/config/general.h
+++ b/code/espurna/config/general.h
@@ -404,13 +404,17 @@
 #define RELAY_REPORT_STATUS         1
 #endif
 
-// Configure the MQTT payload for ON/OFF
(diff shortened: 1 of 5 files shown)


mcspr:
cc @InduPrakash from https://github.com/xoseperez/espurna/issues/1085
General idea is configure both incoming and outgoing on/off/toggle payload strings from runtime. Outgoing is strictly from the setting, incoming is lax parsing matching old 0/1/2 first, then custom payloads and then "off"/"on"/"toggle" (case insensitive)

PR #1891, opened 2019-09-03
Add color temperature state to homeasssistant config generation

l3d00m:
As it is provided via mqtt anyways

### DIFF
--- a/code/espurna/homeassistant.ino
+++ b/code/espurna/homeassistant.ino
@@ -164,6 +164,7 @@ void _haSendSwitch(unsigned char i, JsonObject& config) {
             }
             if (lightUseCCT()) {
                 config["color_temp_command_topic"] = mqttTopic(MQTT_TOPIC_MIRED, true);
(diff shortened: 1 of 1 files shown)


mcspr:
Cool, thanks!

Moonbase59:
In the Web UI, this should probably also be shown under HASS → Configuration → Show.

I distinctly remember having seen `color_temp_command_topic` in 1.13.3 but it also seems to have vanished in 1.14.0.

It is one of Espurna’s little niceties (read: great features) that it shows the needed Home Assistant config for those of us who prefer manual installs and I tend to forget such "extra" settings after a while. ;-)

mcspr:
See the code above, condition is just false for your configuration.
`lightUseCCT()` depends on `useCCT` setting being on, and `useCCT` depends on `useWhite` on
("Use white channel" and "Use white color temperature" in Web)

But this PR has nothing to do with the setting, condition was there before since the 9ceca44770cc7772afc020ce7bad2d4cf321eadb. I guess it should be `lightUseCCT() || lighthasColor()` instead?

PR #1905, opened 2019-09-10
lights: fix value2pwm range conversion

mcspr:
https://github.com/xoseperez/espurna/pull/1902#issuecomment-529144343
> Sending HSV commands works perfectly though, as does the HSV returned on mqtt, but again no physical output from device.

- revert toPWM type change, value could not fit values larger than 255
- try1 in handling map() output types

however, arduino map() is technically not entirely correct
https://github.com/PaulStoffregen/cores/blob/238b102ac46d1184e7055943cb5fe60bcb8eabbe/teensy4/wiring.h#L47
https://github.com/arduino/Arduino/issues/2466#issuecomment-69873889
https://forum.pjrc.com/threads/44503-map()-function-improvements
as I have not tried it with floating point inputs, it is likely broken

### DIFF
--- a/code/espurna/light.ino
+++ b/code/espurna/light.ino
@@ -458,8 +458,8 @@ void _toCSV(char * buffer, size_t len, bool applyBrightness) {
 
 // See cores/esp8266/WMath.cpp::map
 // Redefining as local method here to avoid breaking in unexpected ways in inputs like (0, 0, 0, 0, 1)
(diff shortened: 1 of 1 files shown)


CrazyIvan359:
Oh Paul, always fixing Arduino libraries lol.

I'll let you know how it goes when I have a chance to test.

PR #1907, opened 2019-09-11
test: dns probing and netstat for terminal

mcspr:
https://gitter.im/tinkerman-cat/espurna?at=5d761a6e74c23124a2e2662f, @Niek 
Do we even need MDNS_CLIENT_SUPPORT then? I can resolve .local hostnames with built-in resolver, no need to use the external library. Only applies to lwip2 builds:
https://github.com/d-a-v/esp82xx-nonos-linklayer/blob/e23a07e5f9ccb43d36777c497fc34d044e3ecf3b/glue-lwip/arduino/lwipopts.h#L1164-L1168


Niek:
I just tested, this works fine for me on LWIP2 without MDNS_CLIENT_SUPPORT

Issue #1835, opened 2019-07-22
ota update Flash Write Failed


gernst48:
Trying to use ota firmware update in a serial terminal leads in the following error:

ota http://192.168.2.4/data/nodemcu-lolin-eg-power-firmware.bin
[274720] +OK
[274725] [OTA] Downloading /data/nodemcu-lolin-eg-power-firmware.bin
ERROR[1]: Flash Write Failed
[316147] [OTA] Progress: 1073667996 bytes
ERROR[1]: Flash Write Failed
[316147] [OTA] Disconnected

tonilopezmr:
it seems `1073667996 bytes` are too much!

mcspr:
@gernst48 What server are you using at the http://192.168.2.4?
1073667996 bytes feels like a bug with http parsing on our side though

mcspr:
@Niek you mentioned in the #1716 fixes for OTA headers, was that the problem?

I will try it out some time this week, and there's another possibility header parsing is broken on our side
- https://bitbucket.org/xoseperez/fauxmoesp/issues/46/cannot-use-setbinarystate-with-pywemo , headers arrive separate from the body
- it needs something like #1806 to remember parsing state (across multiple callbacks) and more safe-checks
- headers again, but too much, so we never get to the body in the same callback

I will add #1751 to clearly compare OTA implementations. And it also includes a little fix to immediately close the connection

Niek:
@mcspr the current OTA code expects 2 * \r\n in the first data chunk and assumes that's where the response headers stop and the data starts. Some web servers might send some weird headers, or very long ones, or maybe the MTU is small, so this pattern can end up in the second chunk (not sure how ESPAsyncTCP's ``onData()`` chunking works internally - but I ran across this in a few tests), so in https://github.com/xoseperez/espurna/pull/1716/files#diff-668d7fcc36707d45e425775d15bc9b5bR112-R120 I changed it to read char-by-char and it should be more fault-tolerant.

#1751 + ``OTA_CLIENT_HTTPUPDATE`` should also fix this since it handles the header stuff already.

mcspr:
True, using `OTA_CLIENT_HTTPUPDATE` should handle things properly (however, idk about if there is anything broken with the Core 2.3.0 builds)
What I tried on the weekend is to migrate HTTP parser logic from the thingspeak to some abstract "HTTP client" - https://github.com/xoseperez/espurna/compare/xoseperez:dev...mcspr:util/http
[...]

PR #1909, opened 2019-09-12
Generic HTTP class: Thingspeak and Async OTA

mcspr:
Reuse thingspeak http enhancements with ota module: handle lwip2 low memory, handle status errors, handle lifetime (WIP)

It is pretty basic, only "real" HTTP parsing is for status and body. Expected way to properly do that is by using llparse // llhttp instead of rolling our own http parsing: https://github.com/xoseperez/espurna/issues/1835#issuecomment-523195504
But I want to try to integrate it with async web server first.

Thingspeak should work, OTA was not tested at all. Using draft PR as a means to track this.

### DIFF
--- a/code/espurna/libs/Http.h
+++ b/code/espurna/libs/Http.h
@@ -281,34 +281,41 @@ class AsyncHttp {
         }
 
         bool trySendHeaders() {
-            if (headers.done()) return true;
+            do {
+                if (!client.canSend()) return false;
+                if (headers.done()) return true;
 
-            const auto& string = headers.current();
-            const auto len = string.length();
+                const auto& string = headers.current();
+                const auto len = string.length();
 
-            if (!len) {
-                return true;
-            }
+                if (!len) {
+                    return true;
+                }
 
-            if (client.space() >= (len + 2)) {
-                if (client.add(string.c_str(), len)) {
-                    if (!headers.next().length()) {
-                        client.add("\r\n", 2);
+                if (client.space() >= (len + 2)) {
+                    if (client.add(string.c_str(), len)) {
+                        if (!headers.next().length()) {
+                            client.add("\r\n", 2);
+                            break;
+                        }
+                        continue;
                     }
                 }
-                client.send();
-            }
 
+            } while (true);
+
+            client.send();
             return false;
+
         }
 
 
     protected:
 
-        static AsyncHttpError _timeoutError(AsyncHttpError::error_t error, const __FlashStringHelper* message, uint32_t ts) {
+        static AsyncHttpError _timeoutError(AsyncHttpError::error_t error, const char* message, uint32_t ts) {
             String data;
             data.reserve(32);
-            data += message;
+            data += FPSTR(message);
             data += ' ';
             data += String(ts);
             return {error, data};
@@ -328,14 +335,14 @@ class AsyncHttp {
 
             AsyncHttp* http = static_cast<AsyncHttp*>(http_ptr);
             http->last_error = AsyncHttpError::NETWORK_TIMEOUT;
-            if (http->on_error) http->on_error(http, _timeoutError(AsyncHttpError::NETWORK_TIMEOUT, F("Network timeout after"), time));
+            if (http->on_error) http->on_error(http, _timeoutError(AsyncHttpError::NETWORK_TIMEOUT, PSTR("Network timeout after"), time));
         }
 
         static void _onPoll(void* http_ptr, AsyncClient* client) {
             AsyncHttp* http = static_cast<AsyncHttp*>(http_ptr);
             const auto diff = millis() - http->ts;
             if (diff > http->timeout) {
-                if (http->on_error) http->on_error(http, _timeoutError(AsyncHttpError::REQUEST_TIMEOUT, F("No response after"), diff));
+                if (http->on_error) http->on_error(http, _timeoutError(AsyncHttpError::REQUEST_TIMEOUT, PSTR("No response after"), diff));
                 client->close(true);
             }
         }


mcspr:
https://github.com/xoseperez/espurna/pull/1909/commits/3076bc99cb9746db65266e60e0380b189eef53bf addresses the continuous sending, but buffer is held by the api user.
https://www.nongnu.org/lwip/2_1_x/group__tcp__raw.html#ga6b2aa0efbf10e254930332b7c89cd8c5
notes that the pointer needs to live until ack is received, but I am a bit fuzzy how this works without it 🤔 
perhaps it is better to follow the ESPAsyncTCP buffered client and keep window-sized / mss-sized buffer until ack is received and only call client callback then

Tested with Thingspeak API on low-mem LWIP and `ota <url>` finally works with `python -mhttp.server`

mcspr:
@mhightower83 sorry to "summon" you and bringing up other pr, ref https://github.com/me-no-dev/ESPAsyncTCP/pull/115
As far as I understood, error-related changes there are only required for the SyncClient operation i.e. intermixing loop & sys contexts? It feels like I am silently avoiding this specific issue by not touching asynclient object at all and keeping it part of the parent object as a simple member (and not cleaning that promptly too)
But I don't think I've seen tcp_err callback crashing like https://github.com/me-no-dev/ESPAsyncTCP/issues/94 describes 

And also thanks for the comment there regarding LWIP_NETIF_TX_SINGLE_PBUF, finally that explains no need for data copy flag.

mhightower83:
Sorry, I missed your "summon" :) I have xoseperez/espurna in my watch list so your message got drowned out by the other topics. I am surprised there is no additional annotation for the @ reference.

me-no-dev/ESPAsyncTCP#94 was interesting.  I think I was just lucky :)  I was using the fauxmoESP (wemo version) with around 10 devices configured. The fewer the devices the better it worked. The other element was a Windows system with some additional software that wanted to do UPnP discovery, a lot! With the Windows system off, my sketch stayed up most of the day. With it on, a few hours at most. The crashes appeared to be related to TCP RSTs sent to idle TCP connections on the ESP8266. Back then fauxmoESP did not have `client->setRxTimeout()`, so there were more lingering connections.

With this fixed, the problem with NULL _client emerged in SyncClient type interfaces that may have done a yield, etc.

Another corner case was in `fauxmoESP::_onTCPClient` when all of the `_tcpClients` were used up and control passed down to `client->close(true);`. This could result in generating an ERR_ABORT which must be returned to lwIP by the callback that did the abort. However, there was no place in ESPAsyncTCP to preserve this result and pass it back. Because, in the client's `onDisconnect` callback handler, it would delete AsyncClient.

While my modified version of faumoESP diverged a bit from its roots, I don't remember anything specific I had to change for these issues. These issues were dealt with in the ESPAsyncTCP. It was some effort to preserve the API calling interface and make things work.

Correction I remembered one bug: You may want to change these lines in fauxmoESP:
https://bitbucket.org/xoseperez/fauxmoesp/src/f60c46d80f9b8a400319ab8af9af2957eeb276aa/src/fauxmoESP.cpp#lines-345,346
Reverse lines 345 and 346, `i` in the `DEBUG_MSG_FAUXMO` statement becomes invalid after `delete c;`

>It feels like I am silently avoiding this specific issue by not touching asynclient object at all and keeping it part of the parent object as a simple member (and not cleaning that promptly too)

One thought, if you are not ready to clean the AsyncClient object from the parent object, don't delete AsyncClient in the `onDisconnect` callback. Keep it so you can reference the object's state. I would then think it would be safe. IMHO deleting AsyncClient in `onDisconnect` is tricky and makes my head hurt. :)

mcspr:
Thank you! Telnet server uses the same technique btw, but I hope that causes less problems than more aggressive WeMo / Hue discovery. I would probably re-do it though, checking for connection state later in loop, sort of like wifiserver does. fauxmo stuff is more tricky, because I am both not an active user and not really sure which one emulation layer should've stayed at all (if not both at the same time).

Using updated ESPAsyncTCP bumps .bin size by ~6kb and is currently failing for ssl builds, however those broken, https://travis-ci.org/mcspr/espurna/jobs/587857201#L522 so this still bums me out :/ *(why it wasn't esp/netconn underneath in the first place...)*

mhightower83:
I am really sorry about the broken SSL builds. There are so many build permutations. I really thought I had done that one; however, I can tell by the remaining unused variable warnings that I had missed it. I have to get rid of those things or I cannot see my errors. I didn't have any examples for ESPAsyncTCP with SSL to build against; however, I thought I had at least set the flag and ran compile. Again I am really sorry. 

Here is a drop-in replacement for the function that call is in.
```cpp
err_t AsyncServer::_poll(tcp_pcb* pcb){
  err_t err = ERR_OK;
  if(!tcp_ssl_has_client() && _pending){
    struct pending_pcb * p = _pending;
    if(p->pcb == pcb){
      _pending = _pending->next;
    } else {
      while(p->next && p->next->pcb != pcb) p = p->next;
      if(!p->next) return 0;
      struct pending_pcb * b = p->next;
      p->next = b->next;
      p = b;
    }
    ASYNC_TCP_DEBUG("### remove from wait: %d\n", _clients_waiting);
    AsyncClient *c = new (std::nothrow) AsyncClient(pcb, _ssl_ctx);
    if(c){
      c->onConnect([this](void * arg, AsyncClient *c){
        (void)arg;
        _connect_cb(_connect_cb_arg, c);
      }, this);
      if(p->pb) {
        auto errorTracker = c->getACErrorTracker();
        c->_recv(errorTracker, pcb, p->pb, 0);
        err = errorTracker->getCallbackCloseError();
      }
    }
    // Should there be error handling for when "new AsynClient" fails??
    free(p);
  }
  return err;
}
```
With this replacement, it should at least compile. If you know of any examples that use ESPAsyncTCP with SSL, that you can point me to? I can do a little more.

mcspr:
No need to be sorry :) While we track down that it builds, idk if I would really advice using it right now since it only supports axtls builds (and as the result tls1.1). I have tried to compile someone else's work regarding bearssl (ref: https://github.com/me-no-dev/ESPAsyncTCP/issues/95#issuecomment-533829004), which does include a small SyncClient example, but I wanted this PR implemented first to have at least something else to test with.

Not really small examples, but there are also 2 other users of ESPAsyncTCP & SSL that I know about:
https://github.com/esphome/esphome
https://github.com/homieiot/homie-esp8266/tree/develop-v3
Which in turn depend on async-mqtt-client and it does have a small example:
https://github.com/marvinroger/async-mqtt-client/blob/master/examples/FullyFeaturedSSL/src/main.cpp

nvm my comment about the size, I was looking at travis01 .bin wrong because it includes influxdb support there. real increase is `+928 .irom0.text,  +224 .rodata`, compared on local build

Issue #1130, opened 2018-08-12
ESP-01 + 1CH 5V RELAY FROM LC TECHNOLOGY


rossmax:
Hi!

I try to modificate some parts of code to make it work. I have ESP-01 + 1CH 5V RELAY FROM LC TECHNOLOGY, https://www.az-delivery.de/products/esp8266-01s-mit-relais?ls=de , https://www.hackster.io/makerrelay/esp8266-wifi-5v-1-channel-relay-delay-module-iot-smart-home-e8a437 

This board does not use GPIO0 or GPIO2 in order to activate the relay, it use the serial communication instead it.
First: you need to open Serial port communication at 9600B, any other speed does not works fine.
Ex. : Serial.begin(9600);

Second: 
To enable the Relay send it by serial port:

const byte miBufferON[] = {0xA0, 0x01, 0x01, 0xA2};
Serial.write(miBufferON, sizeof(miBufferON));

To disable the Relay send it by serial port:
const byte miBufferOFF[] = {0xA0, 0x01, 0x00, 0xA1};
Serial.write(miBufferON, sizeof(miBufferOFF));

I add in hardware.h

```
#elif defined(GENERIC_ESP01S_LCTECH_RELAY)
    // Remove UART noise on serial line
    #define DEBUG_SERIAL_SUPPORT    0

    #define MANUFACTURER        "GENERIC"
    #define DEVICE              "LCTECH_RELAY"
    #define RELAY_PROVIDER          RELAY_PROVIDER_LCTECH
    #define SERIAL_BAUDRATE         9600

    // Relays
    #define DUMMY_RELAY_COUNT             1

    // Buttons
    #define BUTTON1_PIN         	       3    //external button connected to "Rx" pin
    #define BUTTON1_MODE        	       BUTTON_PUSHBUTTON
    #define BUTTON1_PRESS       	       BUTTON_MODE_TOGGLE
    #define BUTTON1_CLICK       	       BUTTON_MODE_NONE
    #define BUTTON1_DBLCLICK    	       BUTTON_MODE_NONE
    #define BUTTON1_LNGCLICK    	       BUTTON_MODE_NONE
    #define BUTTON1_LNGLNGCLICK 	       BUTTON_MODE_NONE
    #define BUTTON1_RELAY       	        1

    // LEDs
    #define LED1_PIN                 1
    #define LED1_MODE                LED_MODE_RELAY
    #define LED1_PIN_INVERSE         1
    #define LED1_RELAY               1

[...]

Issue #713, opened 2018-03-22
ESP8266 WiFi 5V 1 Channel Relay


zafrirron:
Probably the low end of esp relays possible...

https://www.hackster.io/makerrelay/esp8266-wifi-5v-1-channel-relay-delay-module-iot-smart-home-e8a437
https://www.ebay.com/itm/ESP8266-5V-Wifi-Relay-Module-TOI-APP-Controled-For-Smart-Home-Automation-Board/311736287457?ssPageName=STRK%3AMEBIDX%3AIT&_trksid=p2057872.m2749.l2649

The challenge was to get it work "as is"with espurna firmware, including all web features and button + external led, once its done you may consider to add it to supported HW.

Problem Description:
esp01 - 512k

The challenge is that this relay is similar to STM with small modifications, and since the module exposes only the serial connections (5v, GND, Rx,Tx). 
Rx can also set to GPIO 3
Tx can also set to GPIO 1 (also internal blue Led).

So we need to change GPIO1 from serial mode (for STM like relay operation) to GPIO and Back.

Solution:
(this is not an elegant solution but a working prototype for consideration)

Create new "Relay Provider"- RELAY_PROVIDER_HW655

hardware.h:
    #define DEVICE                         "HW655_RELAY"
    #define RELAY_PROVIDER          RELAY_PROVIDER_HW655
// Relays
    #define DUMMY_RELAY_COUNT       1
    #define RELAY1_DELAY_ON               0
    #define RELAY1_DELAY_OFF              0

	// Buttons
    #define BUTTON1_PIN         	       3    //external button connected to "Rx" pin 
    #define BUTTON1_MODE        	       BUTTON_PUSHBUTTON
    #define BUTTON1_PRESS       	       BUTTON_MODE_TOGGLE
    #define BUTTON1_CLICK       	       BUTTON_MODE_NONE
    #define BUTTON1_DBLCLICK    	       BUTTON_MODE_NONE
    #define BUTTON1_LNGCLICK    	       BUTTON_MODE_NONE
    #define BUTTON1_LNGLNGCLICK 	BUTTON_MODE_NONE
    #define BUTTON1_RELAY       	        1

    // LEDs
    #define LED1_PIN            	                 1
   #define LED1_MODE            	         LED_MODE_RELAY
    #define LED1_PIN_INVERSE    	        1
    #define LED1_RELAY                           1

in relay.ino add  special handling for this type 

[...]

PR #1910, opened 2019-09-15
HW-655 Adding support for ESP8266 HW-655 Relay board

daniftodi:
Adding support for Esp8266 HW-655 board.

hw-655


code/espurna/config/hardware.h
```diff
@@ -4059,6 +4059,20 @@
...
+
+    #define MANUFACTURER            "GENERIC"
+    #define DEVICE                  "HW655_RELAY"
+    #define RELAY_PROVIDER          RELAY_PROVIDER_HW655
+    #define SERIAL_BAUDRATE         9600
+
+    // Relays
+    #define DUMMY_RELAY_COUNT             1
+    #define RELAY1_DELAY_ON               0
+    #define RELAY1_DELAY_OFF              0
```
daniftodi: observed they are defined for other relay board, are they necessary?

code/espurna/relay.ino
```diff
@@ -98,6 +98,16 @@ void _relayProviderStatus(unsigned char id, bool status) {
         rfbStatus(id, status);
     #endif
 
+    #if RELAY_PROVIDER == RELAY_PROVIDER_HW655
+            Serial.write(0xA0);
+            Serial.write(0x01);
+            Serial.write(status);
+            Serial.write(0xA1 + status);
+            Serial.flush();
```
daniftodi: A0 A1 00 A1 - close
A0 A1 01 A2 - open

code/espurna/system.ino
```diff
@@ -243,7 +243,7 @@ void _systemSetupSpecificHardware() {
 
     // These devices use the hardware UART
     // to communicate to secondary microcontrollers
-    #if (RF_SUPPORT && !RFB_DIRECT) || (RELAY_PROVIDER == RELAY_PROVIDER_DUAL) || (RELAY_PROVIDER == RELAY_PROVIDER_STM)
+    #if (RF_SUPPORT && !RFB_DIRECT) || (RELAY_PROVIDER == RELAY_PROVIDER_DUAL) || (RELAY_PROVIDER == RELAY_PROVIDER_STM) || (RELAY_PROVIDER == RELAY_PROVIDER_HW655)
```
daniftodi: we need Serial communication initialized (control of HW-655 relays is done through serial port)

mcspr: Yes, this should probably be a runtime flag, but we never got around managing debugging like that bc that depends on serial begin too (and reverse-depends on this condition)

code/espurna/config/hardware.h
mcspr: Other one? 0 is the default delay, so this does nothing

code/espurna/relay.ino
mcspr: STM relay protocol strikes again... slightly different each time

daniftodi:
@xoseperez could you please review this?

mcspr:
if merged, should depend on https://github.com/xoseperez/espurna/issues/713

just to be clear, this is *not* the same as STM relay protocol? i do remember another issue dealing with a similar board, but was resolved as an user error
https://github.com/xoseperez/espurna/issues/1130#issuecomment-482198341

daniftodi:
It works with this STM commands, probably it is another one. not sure.

mcspr:
Can you please try?

Based on the referenced issue, it should be something like this:
```cpp
...
#elif defined(GENERIC_HW_655)

    #define MANUFACTURER            "GENERIC"
    #define DEVICE                  "HW_655"

    #define DUMMY_RELAY_COUNT       1
    #define RELAY_PROVIDER          RELAY_PROVIDER_STM

    #define DEBUG_SERIAL_SUPPORT    0
...
```

mcspr:
>  not sure.

?

PR #1916, opened 2019-09-18
Fix SSL check callback function definition

Niek:
Took me a loooong time to figure out why ``config.on_check()`` was returning 1 (fingerprint) instead of 2 (CA). This fixes the function definition, added some debug lines as well to indicate which check is being used.

### DIFF
--- a/code/espurna/libs/SecureClientHelpers.h
+++ b/code/espurna/libs/SecureClientHelpers.h
@@ -20,6 +20,15 @@ using fp_callback_f = std::function<String()>;
 using cert_callback_f = std::function<const char*()>;
 using mfln_callback_f = std::function<uint16_t()>;
 
+const char * _secureClientCheckAsString(int check) {
+    switch (check) {
+        case SECURE_CLIENT_CHECK_NONE: return "no validation";
+        case SECURE_CLIENT_CHECK_FINGERPRINT: return "fingerprint validation";
+        case SECURE_CLIENT_CHECK_CA: return "CA validation";
+        default: return "unknown";
+    }
+}
+
 #if SECURE_CLIENT == SECURE_CLIENT_AXTLS
 using SecureClientClass = axTLS::WiFiClientSecure;
 
@@ -61,7 +70,7 @@ struct SecureClientChecks {
         int check = getCheck();
 
         if(config.debug) {
-            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %d\n"), config.tag.c_str(), check);
+            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %s\n"), config.tag.c_str(), _secureClientCheckAsString(check));
         }
 
         if (check == SECURE_CLIENT_CHECK_NONE) {
@@ -157,7 +166,7 @@ struct SecureClientChecks {
         bool settime = (check == SECURE_CLIENT_CHECK_CA);
 
         if(config.debug) {
-            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %d\n"), config.tag.c_str(), check);
+            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %s\n"), config.tag.c_str(), _secureClientCheckAsString(check));
         }
 
         if (!ntpSynced() && settime) {
--- a/code/espurna/libs/SecureClientHelpers.h
+++ b/code/espurna/libs/SecureClientHelpers.h
@@ -15,7 +15,7 @@
 namespace SecureClientHelpers {
 
 using host_callback_f = std::function<String()>;
-using check_callback_f = std::function<bool()>;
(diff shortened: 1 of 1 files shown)


Niek:
Yes I thought about converting it back to the macro name, but as far as I know that's not possible without creating a manual mapping? If you know a nice way to do it, please add it.

mcspr:
https://godbolt.org/z/b7N4QU
Check out the PROGMEM part in prototypes header

Niek:
I don't really see how that would work, you can't use ``STRINGIFY(check)`` since it's been assigned to a variable. Maybe I'm missing something?

mcspr:
Maybe I misunderstood the question about manual typing... otoh I can only come up with something like this.
```cpp
String _secureClientCheckAsString(int check) {
    switch (check) {
        case SECURE_CLIENT_CHECK_NONE: return F(STRINGIFY(SECURE_CLIENT_CHECK_NONE));
        case SECURE_CLIENT_CHECK_FINGERPRINT: return F(STRINGIFY(SECURE_CLIENT_CHECK_FINGERPRINT));
        case SECURE_CLIENT_CHECK_CA: return F(STRINGIFY(SECURE_CLIENT_CHECK_CA));
        default: return F("UNKNOWN");
    }
}
```
Or using progmem and typing out `const char SC_CHECK_NONE[] = STRINGIFY(SECURE_CLIENT_CHECK_NONE);` etc. and retrieve things from array where "check" is index. Still via String or manually via memcpy_P / strncpy_P.

mcspr:
BTW any reason not to use F() strings? I remember trying this approach with state->string mapper from here:
https://github.com/xoseperez/espurna/blob/726e302a3ec037a39d809feec5615f19c6cc3207/code/espurna/terminal.ino#L91-L106

Slight difference is that those 4 strings are always loaded in RAM, not on-demand via String buffer allocation.

Niek:
Ah sorry, you're right. Here is the diff in case you want to apply to HEAD:
```diff
index 0a17e28c..8070a108 100644
--- a/code/espurna/libs/SecureClientHelpers.h
+++ b/code/espurna/libs/SecureClientHelpers.h
@@ -20,12 +20,12 @@ using fp_callback_f = std::function<String()>;
 using cert_callback_f = std::function<const char*()>;
 using mfln_callback_f = std::function<uint16_t()>;
 
-const char * _secureClientCheckAsString(int check) {
+String _secureClientCheckAsString(int check) {
     switch (check) {
-        case SECURE_CLIENT_CHECK_NONE: return "no validation";
-        case SECURE_CLIENT_CHECK_FINGERPRINT: return "fingerprint validation";
-        case SECURE_CLIENT_CHECK_CA: return "CA validation";
-        default: return "unknown";
+        case SECURE_CLIENT_CHECK_NONE: return F("no validation");
+        case SECURE_CLIENT_CHECK_FINGERPRINT: return F("fingerprint validation");
+        case SECURE_CLIENT_CHECK_CA: return F("CA validation");
+        default: return F("unknown");
     }
 }
 
@@ -70,7 +70,7 @@ struct SecureClientChecks {
         int check = getCheck();
 
         if(config.debug) {
-            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %s\n"), config.tag.c_str(), _secureClientCheckAsString(check));
+            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %s\n"), config.tag.c_str(), _secureClientCheckAsString(check).c_str());
         }
 
         if (check == SECURE_CLIENT_CHECK_NONE) {
@@ -166,7 +166,7 @@ struct SecureClientChecks {
         bool settime = (check == SECURE_CLIENT_CHECK_CA);
 
         if(config.debug) {
-            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %s\n"), config.tag.c_str(), _secureClientCheckAsString(check));
+            DEBUG_MSG_P(PSTR("[%s] Using SSL check type: %s\n"), config.tag.c_str(), _secureClientCheckAsString(check).c_str());
         }
 
         if (!ntpSynced() && settime) {
```

mcspr:
Thanks! 
I wish char strings were always in flash, but these PSTR / F / PROGMEM shenanigans are here for historical reasons. F() seems like the easiest approach right now, possibly replacing them back to `const char*` again whenever esp8266 Core starts using esp-idf / rtos sdk as base (even if this bumps code size ~500bytes compared to const char approach)

Current git Core allows things like `DEBUG_MSG(PSTR("%s %s"), PSTR("hello"), PSTR("world"))` if `vsnprintf` is used instead of rom [`ets_vsnprintf`](https://github.com/xoseperez/espurna/blob/deb465763fec9492b177988ba083d2b8182395c2/code/espurna/libs/DebugSend.h#L12) in the debugSend method, but I don't think there's a way to hide it from the user inside of the macro because of VA_ARGS :/

PR #1920, opened 2019-09-23
Added support for reverse telnet

Niek:
Not sure if this is a useful feature to add, but I could see it being useful for some people. This basically adds the ability to telnet to an Espurna device that is NATed/firewalled. After sending the ``host:ip`` of a publicly reachable ncat listener to the MQTT topic ``telnet_reverse``, the device opens a telnet connection.

Note: untested on Async telnet server, but should work. There's one more issue that the first command is ignored, probably easiest is to move ``_telnetFirst = true`` to ``_telnetLoop`` / ``_telnetNewClient`` and set it to false in ``_telnetReverse``?

### DIFF
--- a/code/espurna/telnet.ino
+++ b/code/espurna/telnet.ino
@@ -42,6 +42,58 @@ void _telnetWebSocketOnConnected(JsonObject& root) {
 
 #endif
 
+#if TELNET_REVERSE_SUPPORT
+    void _telnetReverse(const char * host, uint16_t port) {
+        DEBUG_MSG_P(PSTR("[TELNET] Connecting to reverse telnet on %s:%d\n"), host, port);
+
+        unsigned char i;
+        for (i = 0; i < TELNET_MAX_CLIENTS; i++) {
+            if (!_telnetClients[i] || !_telnetClients[i]->connected()) {
+                #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
+                    _telnetClients[i] = std::unique_ptr<WiFiClient>(new WiFiClient());
+                #else // TELNET_SERVER_ASYNC
+                    _telnetClients[i] = std::unique_ptr<AsyncClient>(client);
+                #endif
+
+                if (_telnetClients[i]->connect(host, port)) {
+                    return _telnetNotifyConnected(i);
+                } else {
+                    DEBUG_MSG_P(PSTR("[TELNET] Error connecting reverse telnet\n"));
+                    return _telnetDisconnect(i);
+                }
+            }
+        }
+
+        //no free/disconnected spot so reject
+        if (i == TELNET_MAX_CLIENTS) {
+            DEBUG_MSG_P(PSTR("[TELNET] Failed too connect - too many clients connected\n"));
+        }
+    }
+
+    #if MQTT_SUPPORT
+    void _telnetReverseMQTTCallback(unsigned int type, const char * topic, const char * payload) {
+        if (type == MQTT_CONNECT_EVENT) {
+            mqttSubscribe(MQTT_TOPIC_TELNET_REVERSE);
+        } else if (type == MQTT_MESSAGE_EVENT) {
+            String t = mqttMagnitude((char *) topic);
+
+            if (t.equals(MQTT_TOPIC_TELNET_REVERSE)) {
+                String pl = String(payload);
+                int col = pl.indexOf(':');
+                if (col != -1) {
+                    String host = pl.substring(0, col);
+                    uint16_t port = pl.substring(col + 1).toInt();
+
+                    _telnetReverse(host.c_str(), port);
+                } else {
+                    DEBUG_MSG_P(PSTR("[TELNET] Incorrect reverse telnet value given, use the form \"host:ip\""));
+                }
+            }
+        }
+    }
+    #endif
+#endif
+
 void _telnetDisconnect(unsigned char clientId) {
     // ref: we are called from onDisconnect, async is already stopped
     #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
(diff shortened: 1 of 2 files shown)


mcspr:
It could be useful, but something to consider:
- can this be just mqtt terminal? #1505 
- maybe we should specify a way for mqtt actions (see `MQTT_TOPIC_ACTION`) to have arguments?

Telnet is a bit tricky, because we are not exactly doing real telnet and would need to look at the network stream to check what is broken there (line-endings, some control codes or something else)
[`telnet.netkit`](https://packages.debian.org/en/stretch/telnet) works on the first try, but [`netcat`](https://packages.debian.org/en/stretch/netcat-traditional) or busybox's `telnet` do not

Niek:
I didn't see #1505, but MQTT seems to be less suited for this. First of all I don't think it's very useful to send full terminal output over MQTT. Secondly, this is especially useful when debugging connection issues like MQTT/OTA/etc.

I tested with ncat only, netcat doesn't seem to work indeed (and I don't see an option to use busybox telnet in listen mode). It'll need some further testing for sure.

Niek:
Ah, you were talking about the MQTT topic only. Yes, I think it does makes sense - if the action command is extended some topics can be removed (OTA, relay, and this one come to mind). I think the plain one (not JSON) is the easiest approach.

Coming back to the PR: I have changed the telnetFirst code now, it works fine in both netcat and ncat now.

mcspr:
Will test that asap.
Note the first mention of this is from here: https://github.com/xoseperez/espurna/commit/d69d6bfd5488847e775987f90c8ff1e97a77a2d5
Does this work with real telnet clients thought? And stuff like C-d still resets the connection?

Do you mean pulse action for relays? I do remember that this was initially proposed as `root/relay/N/pulse/set` (https://github.com/xoseperez/espurna/pull/902), but Xose had reverted it into the `root/pulse/N/set` instead. There's overlapping dependency on HTTP API.
I guess this *can* be merged as-is, but I'd rather think a bit about how actions can work here instead of adding one more top level topic.

Niek:
> Will test that asap.
> Note the first mention of this is from here: [d69d6bf](https://github.com/xoseperez/espurna/commit/d69d6bfd5488847e775987f90c8ff1e97a77a2d5)
> Does this work with real telnet clients thought? And stuff like C-d still resets the connection?

With reverse telnet, Espurna is basically the client and the netcat/nc/whatever listening socket is the server. But yes, everything works as expected like Ctrl+C, quit command, etc.

The firstline stuff is in there because this is part of the [telnet negotiation](http://mud-dev.wikidot.com/telnet:negotiation) (should probably be handled differently than naively expecting it to be the first packet). Because the socket that we connect to is not an actual telnet client, there's no need to skip the first data packet.

> Do you mean pulse action for relays? I do remember that this was initially proposed as `root/relay/N/pulse/set` (#902), but Xose had reverted it into the `root/pulse/N/set` instead. There's overlapping dependency on HTTP API.
> I guess this _can_ be merged as-is, but I'd rather think a bit about how actions can work here instead of adding one more top level topic.

If the action topic would be extended, I think that every command that can be done with terminal commands should be done with this topic, right?

mcspr:
What I meant is how that worked with existing functionality. I do see that is ok, so no problems there.

Thanks for the link! I did end up reading telnet source anyway, but the gist of it that we should probably ignore sequences like `255 something something` and `255 250 ...multiple bytes... 240` instead of ignoring the first packet.

code/espurna/telnet.ino
```diff
@@ -42,6 +42,58 @@ void _telnetWebSocketOnConnected(JsonObject& root) {
...
+    void _telnetReverse(const char * host, uint16_t port) {
+        DEBUG_MSG_P(PSTR("[TELNET] Connecting to reverse telnet on %s:%d\n"), host, port);
+
+        unsigned char i;
+        for (i = 0; i < TELNET_MAX_CLIENTS; i++) {
+            if (!_telnetClients[i] || !_telnetClients[i]->connected()) {
+                #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
+                    _telnetClients[i] = std::unique_ptr<WiFiClient>(new WiFiClient());
+                #else // TELNET_SERVER_ASYNC
+                    _telnetClients[i] = std::unique_ptr<AsyncClient>(client);
```
mcspr:
```suggestion
                    _telnetClients[i] = std::make_unique<AsyncClient>();
```

code/espurna/telnet.ino
```diff
@@ -42,6 +42,58 @@ void _telnetWebSocketOnConnected(JsonObject& root) {
...
 
+#if TELNET_REVERSE_SUPPORT
+    void _telnetReverse(const char * host, uint16_t port) {
+        DEBUG_MSG_P(PSTR("[TELNET] Connecting to reverse telnet on %s:%d\n"), host, port);
+
+        unsigned char i;
+        for (i = 0; i < TELNET_MAX_CLIENTS; i++) {
+            if (!_telnetClients[i] || !_telnetClients[i]->connected()) {
+                #if TELNET_SERVER == TELNET_SERVER_WIFISERVER
+                    _telnetClients[i] = std::unique_ptr<WiFiClient>(new WiFiClient());
```
mcspr:
```suggestion
                    _telnetClients[i] = std::make_unique<WiFiClient>();
```

Niek: Fixed in https://github.com/xoseperez/espurna/pull/1920/commits/d62ecf2b6d8bfa814db8373255d4730699c9845e

Niek:
> What I meant is how that worked with existing functionality. I do see that is ok, so no problems there.
> 
> Thanks for the link! I did end up reading telnet source anyway, but the gist of it that we should probably ignore sequences like `255 something something` and `255 250 ...multiple bytes... 240` instead of ignoring the first packet.

I think it can be as easy as changing the code to this, right?
```c++
if (len >= 2) {
  // Ctrl+C
  if ((p[0] == 0xFF) && (p[1] == 0xEC)) {
      _telnetDisconnect(clientId);
      return;
  // Negotiation stuff
  } else if (p[0] == 0xFF) && (p[1] != 0xFF)) {
    return;
  }
}
```

mcspr:
Yep, something like that. As a separate pr though.
Second condition would only handle data at the beginning. As I understood it telnet client would insert this anywhere it finds 0xff as a part of the data stream, but in what case user would send it?

PR #1929, opened 2019-10-02
Telnet: buffered output for AsyncTcp server

mcspr:
- buffer outgoing data in the internal linked list after client would not accept any more data
- schedule telnet client destruction on loop after disconnection routine, see me-no-dev/ESPAsyncTCP#115

This is something I had noticed with Sonoff Basic board running lwip2 low-memory
scheduling is +400B
buffering is +1.4KB

Why not [AsyncTCPbuffer](https://github.com/me-no-dev/ESPAsyncTCP/blob/master/src/ESPAsyncTCPbuffer.cpp)? -1KB code size and slightly less memory allocations.
And technically, nothing stops us from using this class for something else.


mcspr:
It would also make sense to always buffer / add callback (#1909) for shared data (debug, ws announce). But it feels like it would be much easier to implement this with another class that only does that specific thing on a list of clients, pushing chunks of data in a queue until every client has it (...onAck for each client should count written data)

Debugging itself needs a queue of sorts to help with that, but that is a whole other issue.

Also a honorable mention of [AsyncPrinter.cpp](https://github.com/me-no-dev/ESPAsyncTCP/blob/master/src/AsyncPrinter.cpp), but technically we don't need (or want, because of delay() usage) Stream API.

Issue #1931, opened 2019-10-03
Restore backup to a newly flashed device


cconde:
**Is your feature request related to a problem? Please describe.**
After messing with firmware I tend to have to restore a backup. 
In order to restore a backup I have first to reset the default password and then go to "Admin" to restore the backup.

**Describe the solution you'd like**
It would be superb to have a "Set password & restore config" button at that very first page.

**Describe alternatives you've considered**
I don't know if it's possible, it would be totally awesome if the restore would also restore the original password.

**Additional context**
I keep in mind this may only be useful to few people.

Thanks for this great work!

mcspr:
On a sort-of related note, the starting page could use WiFi setup as well.

You can also skip WebUI completely, see:
https://github.com/xoseperez/espurna/wiki/Configuration#direct-http-settings-api
This will overwrite configuration using the current password

Tofandel:
Was thinking about improving the first step as well for the UI, it's a tad annoying

Isn't the password stored in the config backup?
Cause then you can just make it

Backup or (Password and Wifi)

And then connect to the device via local network instead of the usual
Change password => reconnect => setWifi => reconnect again

Tofandel:
Also it's possible to skip the change password step by setting the build flag WEB_FORCE_PASS_CHANGE to 0

sfilinto:
i reset my devices sometimes so in order to bring them to a desired state very quickly i have made 2 .JSON files which have the desired configuration items. i run the below command to configure the device `curl -F "file=@1.json" -u admin:fibonacci http://192.168.4.1/config` and it take a couple of seconds.
1st JSON file changes the default password & adds my wifi ssid & password.
login with a browser and reboot the device.
2nd JSON file has all the desired configuration and hostname.
again i login and reboot the device.

entire procedure take under 5 minutes to bring a device to a desired state & also avoids spelling mistakes.
HTH.

Tofandel:
@sfilinto if you want there is a trick to do that in one step even faster, use `jq` to merge your two json files, send the output via curl and tadam done should take less than 20 seconds
`jq -s '.[0] * .[1]' 1.json 2.json | curl -H "Content-Type: application/json" -d @- -u admin:fibonacci http://192.168.4.1/config`

Issue #1899, opened 2019-09-06
Test web interface


mcspr:
**Is your feature request related to a problem? Please describe.**

Following #1898 , it would be useful to have a tests / test mode for WebUI

**Describe the solution you'd like**

https://gitter.im/tinkerman-cat/espurna?at=5d717445ae44a8412487f6c3
One of the options is to have a dummy web-server (node.js is a likely choice) that can WebUI page can interact with.
As an alternative, web page itself should have a special mode for running from `file:///` (current check could be extended)

General test suite for .js code would also be useful.

**Describe alternatives you've considered**
\-

**Additional context**
\-

tonilopezmr:
Hi, I'm learning web development and using [Cypress](https://www.cypress.io/) to make E2E testing using screenshot testing, and if we can mock the WebSocket using jest-websocket-mock commented here https://github.com/xoseperez/espurna/pull/1933#issuecomment-543166651, I think its a good way to go to create Web UI tests very easy.

mcspr:
Cypress sells itself as a all-in-one thing, which sounds interesting.
However, if I properly understood how it works, in case of interactive stuff the only target browser is Chrome?
https://github.com/cypress-io/cypress/blob/dc251cc5d6c18da2be64ae3de269a00b75985099/packages/launcher/lib/browsers.ts#L6-L31

Re-quoting https://github.com/xoseperez/espurna/pull/1933#issuecomment-543166651, the point of mentioning local webserver was to possibly reuse some of the code in case Arduino & support libraries are mocked / redone to run on the host if that would be useful. And I think I also thought about a way to locally run the WebUI as-would-be on the device, without going about unhiding menus and panels manually.
fwiw the current webui's only entrypoint is the `processData` function (minus `connectTo...`), so local tests could already somehow work if something loads up index.html as `document` and feed data objects to it.

*...and I would say I have not done as much progress on this as I would've liked to...*

Issue #1898, opened 2019-09-06
Test utility functions


mcspr:
**Is your feature request related to a problem? Please describe.**
ref #1864 , there needs to be a concise way to check that utility functions (brightness calculation, channel setting, value conversion etc.) work without problems with some set of expected inputs and outputs.

**Describe the solution you'd like**
- special build option to replace application code with bunch of assert calls. if it is possible, partly run application code. runs on the device
- build and run application on the host system, including utility functions source (`#include *.ino` or otherwise)

**Additional context**
https://tinkerman.cat/post/automated-unit-testing-metal
https://github.com/esp8266/Arduino/tree/master/tests/device
https://github.com/esp8266/Arduino/tree/master/tests/host

cc @xoseperez

mcspr:
Current implementation via https://github.com/bxparks/UnixHostDuino in the https://github.com/xoseperez/espurna/tree/dev/code/test
To run host test, simply use `pio test` in the `code/test` directory. Only test at this time is for tuya module.

I also tried using https://github.com/arduino/ArduinoCore-API
- clone the repo while inside `test/` directory
- `cp -r ArduinoCore-API/api lib/Arduino`
- add `Arduino.h` in the `lib/Arduino`, which includes `ArduinoAPI.h`
- add `Arduino` to the lib_deps in `platformio.ini`
- (lol) don't use filesystems with case sensitivity, because Arduino "deprecated" WString for String.h, I could not get `#include <string.h>` to work for `strlen`, `memove` etc.

mcspr:
Also in regard to the ArduinoCore-API usage - because it is just an API reference, there are host functions missing:
https://github.com/esp8266/Arduino/blob/master/tests/host/common/Arduino.cpp
https://github.com/bxparks/UnixHostDuino/blob/bafce102e2c39f205e6cf527d7ca02fa19e9d8b5/Arduino.cpp#L27-L58
https://github.com/arduino/ArduinoCore-API/blob/398e70f188e2b861c10d9ffe5e2bfcb6a4a4f489/api/Common.h#L92-L128

And another small example of WiFi scanning, plus millis and delay implementation (but for a proper esp8266 simulation, there should probably be two threads that will behave like CONT and SYS):
https://github.com/xoseperez/justwifi/compare/master...mcspr:mocked#diff-d4059dfc84ae48bb13df27ad23761758

mcspr:
[...]

PR #1843, opened 2019-08-06
Fixed size json payload & other WS bugfixes

mcspr:
Resolve #1731 (?), #1823 (original question), #1824, #1831 
cc @gernst48 @tonilopezmr @Interpyme @ruimarinho 

I'll clean it up later from some experimental leftovers and fix some dumb behaviour (like, not immediately sending sensor readouts)
POW runs this reasonably well. Refresh does not crash 🤷‍♂
(but I could not make it crash even with old builds. must be something related to OS / browser / network conditions)

And probably try again to rethink how to push static data before everything else, to reduce total number of messages (although, it is smaller than before #1387 / 1.13.3)

Keeping ArduinoJson 5 for now.

PR #1933, opened 2019-10-07
[WIP] New Vue Web UI

Tofandel:
#TODO#


tonilopezmr:
I :) I'm learning web development for embedded applications and I'm testing [preact](https://github.com/preactjs/preact) and just curious about Vue. 

Can you explain in the PR message which will your strategy for build the project, also it is important to test the website, have different components and in the build create a single HTML with all styles and JS. How are you going to manage that?

After some research using LitElement and preact, preact is way smaller, with the same application LitElement is 21Kb size and Preact 4Kb. I used Webpack to inline the css and JS code in a single HTML

Tofandel:
Build will be made by webpack so with a bit of configuration it will handle the minification of html/js/css and inlining them into one html which then will need to be gziped

Never heard of preact but vue makes a lot of sense for single page UIs it's really easy to understand what's happening and having worked with react the component logic and writing is simpler (meaning more intuitive for the community to support and more lightweight as well)

https://vuejs.org/v2/guide/comparison.html#Preact-and-Other-React-Like-Libraries

4kb is indeed a very small framework size, I wonder if they include all important features of react

tonilopezmr:
Yes, Vue was another alternative I studied but I've never used it before (I used react) and it was easy for me to create a small application to test preact, it includes the important features.

Thank for your explanation 🎉 🎉 , I hope I can test Vue as well to see the comparison.

mcspr:
> Configure the build process
 Implement a npm build process on PIO compilation

Some previous attempts were here #970 and #1680 , but the way it was done is by updating PlatformIO SCons filelists. Recent platformio versions make it so that each environment has it's own scons file list, so if I were to add web files to the track list there it would always rebuild the UI when the firmware is built. It can also not do that and just call npm-build and hope it did something (but the question is now, how it would know about which modules it wants)

Configuration should be easy enough to finally implement, but all the versions so far used to run gcc preprocessor and manually parse the result via regexp. I wonder if that could be avoided somehow and still show an expected build config.

As you mentioned in the other issue, backend updates can go straight to the dev optionally / updating existing webui too.

code/espurna/relay.ino
```diff
@@ -719,39 +719,23 @@ String _relayFriendlyName(unsigned char i) {
...
-    JsonArray& pulse = relays.createNestedArray("pulse");
-    JsonArray& pulse_time = relays.createNestedArray("pulse_time");
 
-    #if MQTT_SUPPORT
-        JsonArray& group = relays.createNestedArray("group");
-        JsonArray& group_sync = relays.createNestedArray("group_sync");
-        JsonArray& on_disconnect = relays.createNestedArray("on_disc");
-    #endif
+void _relayWebSocketSendRelays(JsonObject& root) {
+    JsonArray& relays = root.createNestedArray("relays");
```
mcspr: Note for this change and further structs as objects, that this was intentional: https://github.com/xoseperez/espurna/issues/1367, #1614, #1387, #1824, #1843 
Each structure adds up in size, both memory for storage and for transfer buffers

*edit:* MTU should not be a problem in the future because of Core version upgrade, just a historical reference

Tofandel: Mmm, I was still thinking about this design part, whether the UI should compute the data or the websocket send a cleaner output, I believe a cleaner output is the way to go

About the memory, isn't the buffer cleared after the json message is sent?
It should actually not take a significant amount more (It's btw just to send it to the websocket and not for EEPROM storage), storage won't be modified (at least not in this PR)

Tofandel: My main issue is the data received is not easily readable/iterable, the root should be clustered by modules instead of having all the settings intersect.

To save them however it's a non issue, I can just get the data in all modules and process it like nameOfModule + UCFirstSettingName

Tofandel:
Little update: around 50% of the work is completed

code/espurna/sensor.ino
```diff
@@ -234,73 +231,79 @@ void _sensorWebSocketSendData(JsonObject& root) {
...
+        
+    //TODO why is this not nested?
+    sensor["pwrUnits"] = _sensor_power_units;
+    sensor["eneUnits"] = _sensor_energy_units;
+    sensor["tmpUnits"] = _sensor_temperature_units;
+    sensor["tmpCorrection"] = _sensor_temperature_correction;
+    sensor["humCorrection"] = _sensor_humidity_correction;
+    sensor["snsRead"] = _sensor_read_interval / 1000;
+    sensor["snsReport"] = _sensor_report_every;
+    sensor["snsSave"] = _sensor_save_every;
```
Tofandel: In this module, what are all this values and why are they not per magnitude or per sensor?
Lost me on this code

code/espurna/relay.ino
mcspr: JSON buffer would only keep the data until it is stringified, then the string itself needs to be kept until the network accepts all of the data. RFBridge is a good case where there are a lot of relays at the same time, but you can add them on any board by using DUMMY_RELAY_COUNT=N

No problem using a key per module even when they are not all sent in the same message, but nesting things and having a lot of objects is not trivial for memory. 
But is there any specific reason for avoiding property=[n1,n2,n3] as a way to compose the data? Mentioning rfbridge again, there is also a way to partially send the data (via start_index and size values)

Tofandel:
Adding #1898 #1899 #1931 in the WIP queue
A server is not needed, the data can be mocked with jest in unit tests

See https://www.npmjs.com/package/jest-websocket-mock

code/espurna/sensor.ino
mcspr: Correction code makes more sense to have per sensor instead of globally, perhaps the assumption was that the device only uses a single sensor or sensors are all in the same place / have similar problems that require correction

Read / report values could go per sensor too, since those could have a different requirements. idk about saving, because the only use case right now is energy (which, btw, is saved per sensor as eneTotalN)

Units could be kept global? Is there a reason to have them specialized?

code/espurna/relay.ino
mcspr: And to get estimates on size without running any code:
https://arduinojson.org/v5/assistant/ (we are currently using this one)
https://arduinojson.org/v6/assistant/ (slightly larger than v5, so there's a hold up here)

Tofandel: I'll keep the start_index in mind for partial data transfer, I did some calculations with the link provided, the code adds 50 bytes of memory per relay (minus one relay), is it gonna be a problem?

The reason for preferencing [{property:n1,id:1},{property:n2,id:2},{property:n3,id3}] over {property:[n1,n2,n3],id:[1,2,3]} is OOP and data compartmentalization, my teachers used to talk a big deal about this when I learned C, about how it's dangerous to have data representing an object sparsed into arrays (If you delete or add an item in an array but not the other you can break the entire representation, so you need to make sure you add/delete in every arrays, whereas with the other representation if you delete or add an item, well there is no problem)

So sure it adds a bit of memory because you'll have to duplicate the keys multiple times but in the long run if we start adding features, it's less prone to bugs and mistakes and also easier to parse, edit and display properties on the other side

mcspr: For example: https://gist.github.com/mcspr/63f5213f425fa9af95cf60f893c755c3
More around 1...1.5KB, the resulting string grows 3 times in size. Group string is probably needless, just an example. You can also prepare sensor example with multiple magnitudes, something like HLW8012, which are probably a second most common case with a lot of data.
Issues with memory blowing up are rather annoying things to have :/

The case of missing values or size mismatch could be enforced by the sender, we control both sides. However there may be an issue of having too much things in relay object itself or any other object where we need to have a lot of such arrays for properties.

So it feels rather backwards to just undo this optimization (note the PR! this was the way Xose implemented it initially, barring the namespace keys you are introducing and updated queueing mechanism).
Perhaps the thing we want is to send objects as arrays i.e. ctor arguments and prepend some kind of schema? See last json in the gist

mcspr:
> Create a new build mode "PWA" that will be the default, it will generate a full PWA app that can be hosted on espurna.io for all users or locally as well, and the device will not need a web interface since the PWA will be accessible offline reducing the build size considerably

:+1:

Tofandel:
Anybody wants to make a nice svg icon for Espurna?
Would be nice for the PWA and the favicon

This is how the UI currently looks
image

tonilopezmr:
>Create a new build mode "PWA" that will be the default, it will generate a full PWA app that can be hosted on espurna.io for all users or locally as well, and the device will not need a web interface since the PWA will be accessible offline reducing the build size considerably

How can you deal with the device offline using PWA? You need to write the IP or something like that?

Tofandel:
@tonilopezmr Originally yes, but I'm writing a clever way to discover local devices via the PWA, the app will include a new header: "Powered-by: espurna, version"

I can call a webRTC service to get the local IP address of the device and then ping all the local ips on a new endpoint exposing only their names without authentication and list them

The user can then choose from the list and authenticate himself when the device is selected

mcspr:
> 
> 
> Anybody wants to make a nice svg icon for Espurna?
> Would be nice for the PWA and the favicon
> 
> This is how the UI currently looks
> [image](https://user-images.githubusercontent.com/6115458/67586501-13053c80-f752-11e9-8313-bcb9ce3d140c.png)

From recent memory, @xoseperez used this https://twitter.com/espurna_io/status/1172545472742993920 as a graphics for twitter account & stickers. So it is either this one or more high-res current one?

tonilopezmr:
Can I propose some new logo for a new webpage?

Here is my proposition based on the current one, what do you think? @xoseperez 

Screen Shot 2019-10-30 at 01 21 40

Tofandel:
@tonilopezmr Nice
Do you have an svg file of this?

Could you try with this color #479fd6?

tonilopezmr:
haha !! That's right, I don't use facebook messenger :P 

With the new color you suggested:



What do you think about this new spark rotation? Spark and the triangle have parallel lines. 👀 

WhatsApp Image 2019-10-30 at 13 20 55 (1)

Tofandel:
@tonilopezmr I prefer the first triangle without the rotation, so you can share an svg of this one, for the fonts, better use websafe fonts like sans-serif, Arial or Helvetica

Tofandel:
Super compressed svg (without the triangle and with the primary color)
`<svg width="256" height="256" version="1.1" xmlns="http://www.w3.org/2000/svg"><path transform="matrix(.913 0 0 .913 -19.8 .231)" d="m80.5 154c-16.6-3.01-18.5-22.1-3.62-33.1l155-100-101 101 113 1.35c30.7 13.9-5.49 34-19 44.1l-144 93.3 106-103z" fill="#ff952f"/></svg>`

Tofandel:
Favicon is lit like this :)

tonilopezmr:
@Tofandel How did you compress the svg ?? :O The favicon can be an svg? 

I love it 👏 👏 👏

Tofandel:
@tonilopezmr A favicon can be an svg but it's not well supported (safari, firefox, opera only, https://caniuse.com/#feat=link-icon-svg), I can still have a 32x32 or 16x16 png generated in the non pwa app

I compressed it by reducing the curved lines needed (only 2 curves the rest are lines) and then in inkscape when exporting I set the decimal number to 2 instead of 5, final optimization by hand (remove the group and useless properties)

Tofandel:
70% rolled, depending on the weather it will be finished in 2 or 3 weeks

code/ui/src/tabs/common/Setup.vue
```diff
@@ -0,0 +1,233 @@
...
+            </div>
+        </div>
+    </Form>
+</template>
+
+<script>
+    import Btn from "../../components/Button";
+    import Inpt from "../../components/Input";
+    import Hint from "../../components/Hint";
+    import C from "../../layout/Col";
```
mcspr, line 97: (nit) Not specifically related to Setup code, just in regard to naming
Any reason not to use full name like C -> Column, Btn -> Button?

tonilopezmr: +1

tonilopezmr:
When the web finished I'd like to help testing it, now I have more knowledge about web development what do you think about forgetting about unit testing and add more acceptance tests doing screenshots of the website depending on some states.

Tofandel:
@tonilopezmr You can integrate both, but it's always a good idea to have unit tests for functions to avoid edge cases that you wouldn't really get in acceptance testing and there is also things you cannot check that way (eg: make sure the data received and sent is the correct format)

Also while having 10 screenshots to check is fine, having 100 screenshots to check on each release will be prone to human error and some bugs will definitely be missed, so I would be inclined to do a maximum of unit tests and some acceptance test

tonilopezmr:
I agree with you, it was only a solution to create a test easily than unit test if people don't do it. It's better to have a screenshots test than nothing. I think the best way is the scenario you said. 👍

code/ui/src/common/websocket.js
```diff
@@ -0,0 +1,80 @@
+// #if process.env.NODE_ENV === 'development'
```
mcspr: Q: which part of the build process understands this line?

I was playing around with babel by adding it to the existing gulp. There could be a real code `if (process.env.BLA === "true")` and it will be translated into `if (false)` or `if (true)` based on the env variable, thus possibly eliminating that branch as dead code with babel-minify (and I think transform-inline-environment-variables is also required?) or I'd imagine any other minifier.
*edit: reword*

mcspr: A: https://www.npmjs.com/package/webpack-conditional-loader
(figures)

Tofandel: Yep that's the one

mcspr:
> I might need some help getting the public API built though
(To output device name, description, board name and espurna version on a public endpoint)

Some time ago there was this example of local discovery app. I simply iterates local subnet sequentially and tries to open up `/discovery` with some public data:
https://github.com/KimNyholm/espurna-device-locator/blob/master/src/services/ESPurnaDeviceClass.js
Maybe device itself could help with discovery a bit, through mDNS or raw UDP multicasts since it is not restricted by the browser stack

BTW, how one would go about adding a module? Effectively App.vue hides loaders manually, I would need to type out everything like that? Or is there some way to have "sources" / directories with components that are loaded dynamically? e.g., if user does not want to change original sources at all to allow easy updates between versions

Tofandel:
> BTW, how one would go about adding a module? Effectively App.vue hides loaders manually, I would need to type out everything like that? Or is there some way to have "sources" / directories with components that are loaded dynamically? e.g., if user does not want to change original sources at all to allow easy updates between versions

I'm not sure I understand


Iterating on the subnet is what I'm doing for now (with auto detection of the subnet)
But does the `/discovery` endpoint exists? And is it public?

mcspr:
What I meant are these modules, if I correctly understood the loading process:
```vue
            <!-- #if process.env.VUE_APP_RFBRIDGE === 'true' -->
            <template v-if="data.modules.rfb" #rfb>
                <Rfb v-bind="data"/>
            </template>
            <!-- #endif -->
```
As previously mentioned, webpack conditional loader will strip this from the resulting bundle when env `VUE_APP_RFBRIDGE != "true"`. Thus I was wondering if that could be generated somehow, akin to concatenation of `code/espurna/*.ino` for the firmware itself but with some conditional loading still present e.g. `<filename>` is only added when `VUE_APP_<uppercase-filename>="true"` 

--- 

=> [`/discover` endpoint is without any auth](https://github.com/xoseperez/espurna/blob/a2599de4d11481ba9f9712087472ee63bc7e9a7f/code/espurna/web.ino#L75-L94)

Tofandel:
Okay I see, that would require a code generator of some sort or a custom loader

I don't know of any webpack plugin that would allow this but it may be possible to edit webpack conditional loader to get a code preprocessor basically, but that's not a priority, being 5 line per module that can be copy pasted

The normal solution would be to use a dynamic import but that wouldn't work in a single compressed file like currently

> => /discover endpoint is without any auth

Wow, just tried it, I just need to add the description in that, that's amazing, gonna save a lot of work

mcspr:
No problem. Looking at https://github.com/caiogondim/webpack-conditional-loader/blob/master/src/index.js, transformation is pretty straightforward anyway (so I guess hiding things like `import Rfb ...` would work too)

mcspr:
Kind-of a ping with questions:

- anything I can improve from ws / api side of things?
- does schema / arrays still make sense or is there some way better to stream data to the UI? Keeping in mind RAM requirements previously discussed
- from semi-recent gitter comments, Xose had idea about generating "schema-like" descriptions of sensors to generate UI elements. Considering some options for this, would making the device dump array like {group, cfgkey, type, default} (e.g. {`wifi`, `ssid`, CfgType::List, ""}) make sense for such a thing? I am still thinking of how this would even work without eating RAM, but I wonder...

Tofandel:
@mcspr Quite busy so don't have time to roll it up for beta testing yet

There is a few settings I added on the interface that don't exist yet

Under relays:

- `dblDl` (int, time in ms time in between two clicks to retain a double click, replaces a general setting to a relay basis, prev setting `btnDelay`, flag `BUTTON_DBLCLICK_DELAY`) 

- `lngDl` (int, long click delay in ms, replaces a general FLAG on a relay basis, prev flag `BUTTON_LNGCLICK_DELAY`)

- `lngLngDl` (int, long long click delay in ms, replaces a general FLAG on a relay basis, prev flag `BUTTON_LNGLNGCLICK_DELAY`)

- `sndAllEvts` (boolean, to send all mqtt events or not, replaces a general FLAG on a relay basis, only if module mqtt, prev flag `BUTTON_MQTT_SEND_ALL_EVENTS`)

I think that's all I added, they are in the schema sent already but not used yet

If you could implement those options on the .ino side that'd be amazing (on this PR or on dev up to you)

I agree that the sensors could use a little refactoring but I am not too familiar with it, maybe first let's roll this out and then think on how to refactor the sensor to minimize feature change in this PR

code/espurna/wifi.ino
```diff
@@ -453,15 +453,16 @@ void _wifiWebSocketOnConnected(JsonObject& root) {
...
     JsonArray& schema = wifi.createNestedArray("schema");
-    schema.copyFrom(keys, 7);
 
-    JsonArray& networks = wifi.createNestedArray("networks");
+    schema.add("ssid");
+    schema.add("pass");
+    schema.add("ip");
+    schema.add("gw");
+    schema.add("mask");
+    schema.add("dns");
```
mcspr, line 445: Note that I was kind of doing the same thing, hardcoded 7 can also be (sizeof(keys)/sizeof(keys[0]))
I changed the behaviour of wifi settings, getSetting() will return hardcoded default. `stored` flag is to avoid removing hard-coded entry accidentally, since on ui reload it will appear again anyway

Tofandel: I just didn't want to use copyFrom because it was removed from version 6 of the arduino Json lib (Using v5 but if ever we move away)
Also not sure but doesn't it double the ram usage for the array (define + clone, seems redundant)
Edit: Never mind, add also clones the value

I don't get the stored flag
Don't `if (!getSetting({"ssid", index}, _wifiSSID(index)).length()) break;` prevent it from ever being set to false anyway, by not sending wifis that are not set?

mcspr: See https://arduinojson.org/v6/doc/upgrade/#copyarray

When we have `WIFI1_SSID=something`, getsetting is never empty

mcspr:
idk if you are testing something, but what is up with newline removal of debug? why is it needed?

Tofandel:
Just experimenting

code/espurna/wifi.ino
Tofandel: Oh I finally got it, so the WIFI can be hardcoded with defines and this flag tells us if it comes from the define or not

But can't they be overwritten? The defines acting just as a default value

mcspr: Yes, see `code/espurna/wifi_config.h`
They are hard-coded defaults, we can't erase them. It is still possible to "shadow" them, either by setting indexed setting to something else or setting "" empty value

Tofandel: Okay so I got it all but then there might be an issue, if you overwrite the settings the delete button will still be disabled, but it shouldn't (it wouldn't technically delete it, but would reset it to default)
So maybe just changing the name of the button to 'Reset to default' instead of 'Delete' but the API wouldn't know the default value and would need to retrieve it again

Having a hardcoded flag sometimes complicates things more than it simpifies it :p

mcspr: Kind of like that. You can always hit reset-all button though, I guess there could be a specific 'reset wifi settings' one.
pre 1.14.2 versions read hardcoded settings on boot, overwriting runtime settings with those values, so it had the same problem only delayed until the next reset

mcspr: btw, not to speculate on code improvements regarding arduinojson:
https://github.com/xoseperez/espurna/pull/1957#issuecomment-545905580

you can use it as a local library and build program on host. you can repeat the code from above for yourself, array size does not change when we copyfrom() or add individually, it only stores pointers since it is array of pointers.

Tofandel: Yeah that's what I figured, so there is no difference between the two? It's a matter of personal preference then, taking into account that some schemas are variable with macros (defines), is it possible to create an array with macro conditional to copy from?

mcspr: True, effectively it would be the exact same code generated for v5:
https://github.com/bblanchon/ArduinoJson/blob/v5.13.4/src/ArduinoJson/JsonArray.hpp#L151
Slightly different for v6
https://github.com/bblanchon/ArduinoJson/blob/8f8c82d400a2e860e538eed8f92674330186f76f/src/ArduinoJson/Array/Utilities.hpp#L47
But [index] access is also available in v5 (not sure what empty values are though, nulls?)

mcspr:
> Under relays:

*buttons? as:
`btnDblDl` <-> BUTTON_DBLCLICK_DELAY (btw this already was `btnDelay`)
`btnLngDl` <-> BUTTON_LNGCLICK_DELAY
`btnLngLngDl` <-> BUTTON_LNGLNGCLICK_DELAY
`btnSndAllEvts` <-> BUTTON_MQTT_SEND_ALL_EVENTS

All of those are indexed, i.e. per button?

code/espurna/relay.ino
```diff
@@ -894,60 +903,76 @@ String _relayFriendlyName(unsigned char i) {
...
+    schema.add("type");
+    //schema.add("resetGPIO"); //This is not needed
+    schema.add("name");
+    schema.add("boot");
+    schema.add("pulse");
+    schema.add("time");
+
+    schema.add("dblDl");
+    schema.add("lngDl");
+    schema.add("lngLngDl");
```
mcspr: re my previous comment, clicks & click delays are button settings. I'll add the specific part to the button code then

Tofandel: Hmm didn't think about that since the relationship between buttons and relays is almost always 1/1 but there might be cases where buttons don't have a specific relay or vice versa..
In which case it might be a good idea to expose the settings on the button module and to do the relation later on

In the future, that could open the door for a device module where the buttons/leds/relays GPIO, manufacturer and device could be configured. That would allow for non hardcoded generic builds in pio where you could configure your buttons/relays/led on different GPIOs, edit the manufacturer and device freely where one build fits a lot of buttons/relays/leds only devices and you can just have a list/database of configurations online depending on the device (Eg: sonoff 1ch, 2ch, 3ch, touch, sv etc) to just import to the device

mcspr: I think I wanted to do buttons in dev branch first :>
Note on the https://github.com/xoseperez/espurna/pull/1933/commits/b352284ea9f3c676b5a35805d8d8716c1583c93c - .ino should preferably be using constexpr funcs that return preprocessor vars, that way you already know the type.
*edit:* or like with delays - UL suffix will force the type to unsigned by default

But I agree on board ID thing, I don't think I want that after all. Same with migrate, only keeping that for now to reference keyvalues that were used before

Tofandel: Yep I changed all that already :p
I'll be organizing some more stuff (a few settings to rename) and then start testing on a device to finish up the vue app,

For the board id it's not used anywhere and we already have the MANUFACTURER and DEVICE info, for the migration we can keep it (though it may take some build space)
I'll need to add a new config version and migrate some settings anyway so I'll use it

I'll add the buttons settings under the leds tabs and rename it Leds/Buttons
Avoid that clustered mess that is relay

mcspr: I tried some board stuff in the #1680, but was still different from the old v2 approach. Main idea was to support device 'sets' that consisted of key-value pairs and swap config based on that build-time ID. migrate.ino was also safe-guarded against overwriting, like it does right now. 
Headers are finally fixed, so sourcing config/all.h with any external app is much more easier than it was in that PR.

My only concern is to not get too broad with modifications (while staying on the topic of Vue UI, see the title :). But, if that helps with an overall design of the UI, sure; better integrate the stuff now when it is less pain to change.

And as I was saying - I may break when adding the same thing to dev in the meantime, which will probably involve manual merging of stuff related to the configuration. Mostly how things are chained together and supporting indexed stuff at build time.

Tofandel: Hmm for the migrate, does that mean it's not implemented yet? From what I see none of the settings set there are actually used yet..
(Eg relayGPIO, relayType, btnGPIO)

And it was actually an attempt to try what I suggested with the database of devices/settings except here it's hardcoded in the binary

With the pwa it will be possible to have a json file with the list of devices and config and have a modal on init to select the device and load the config at the same time you set the password, won't do it in this PR but to keep in mind


I'm trying to limit modification but I'm refactoring the whole websocket interface (eg: callbacks, schema etc), so I'd rather have some bases covered here than later and have all the breaking changes at once (for a v2 basically) so I'd prefer having all the settings at least added in the schema and add the functionality later

mcspr: Nah, forget about migrate, we already have device settings in the hardware.h + key association is implemented in each module separately instead of having it all in one place. It was a prototype, but there is no actual point of having it anymore.
The general idea from v2 & pr is to keep key-value pairs for each device, based on current MANUFACTURER_DEVICE flag (either all at once, or limited to user-specified set).

As this PR is still wip there is no point of really restricting yourself, but just keep in mind that I'd prefer to merge settings and API changes beforehand. Let's just see what works

code/espurna/button.ino
```diff
@@ -45,12 +57,44 @@ unsigned char buttonCount() {
...
+    );
+}
+
 #if MQTT_SUPPORT
 
 void buttonMQTT(unsigned char id, uint8_t event) {
     char payload[4] = {0};
     itoa(event, payload, 10);
-    mqttSend(MQTT_TOPIC_BUTTON, id, payload, false, false); // 1st bool = force, 2nd = retain
+    mqttSend(MQTT_TOPIC_BUTTON, id, payload, false, getSetting({"btnMqttRetain", id})); // 1st bool = force, 2nd = retain
```
mcspr: Note that 2nd argument's type is always mapped to the return type, getSetting<...>() syntax can be avoided.
(preferably PREPROCESS_FLAGS are adjusted to be the correct type directly instead of casting)
getSetting(key) always returns String obj as-is from the underlying storage, since we don't know the type beforehand

Tofandel: So `getSetting({"btnMqttRetain", id}, false)` is preferred over `getSetting<bool>({"btnMqttRetain", id})` ?

mcspr: Yes, since there is no `getSetting<bool>(key)`
Which could be implemented in theory, but instead of explicit `false` then you'd need to remember what exactly is the default of the type that you want.

code/espurna/debug.ino
```diff
@@ -214,25 +214,17 @@ void _debugSendInternal(const char * message, bool add_timestamp) {
 
 #if DEBUG_WEB_SUPPORT
 
-void _debugWebSocketOnAction(uint32_t client_id, const char * action, JsonObject& data) {
```
Tofandel, line 217: Moved to terminal

mcspr:
While you are looking at HA module, check out https://github.com/xoseperez/espurna/compare/dev...mcspr:ha/ext-structs

Idk if this is still relevant, but the idea is to put string-string container on top and generate either yaml of json from that. c++ is weird though, since we probably need std::unordered_map for hashmap instead of std::map, b/c it uses rbtree + binary search comparisons (and idk how well that works huh)

PR #1948, opened 2019-10-15
Added restore last schedule support

foxman69:
I have added an option to restore the last schedule state on the device boot.

This option is useful in case of power loss or any kind of device reset that you still want your schedules to continue as usual.

At the moment if the device had a "down time" while a scheduled action should have happen, your action wont get applied, with this new feature the device will check for the closest schedule action that should have happened before boot, and restore its state to the desired schedule action state.

I have also added an option in the "SWITCHES" menu to choose on which relay you would like this feature to run on.

Disclaimer: it only counts "Turn ON" and "Turn OFF" actions and ignores toggle actions

Capture

### DIFF
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -733,7 +733,7 @@ void _relayWebSocketSendRelays(JsonObject& root) {
     JsonArray& pulse_time = relays.createNestedArray("pulse_time");
 
     #if SCHEDULER_SUPPORT
-        JsonArray& sch_last = relays.createNestedArray("relayLastSch");
+        JsonArray& sch_last = relays.createNestedArray("sch_last");
     #endif
 
     #if MQTT_SUPPORT
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -753,7 +753,7 @@ void _relayWebSocketSendRelays(JsonObject& root) {
         pulse_time.add(_relays[i].pulse_ms / 1000.0);
 
         #if SCHEDULER_SUPPORT
-            restore_last_schedule.add(getSetting("relayLastSchedule", i, SCHEDULER_RESTORE_LAST_SCHEDULE).toInt() == 1);
+            relayLastschedule.add(getSetting("relayLastschedule", i, SCHEDULER_RESTORE_LAST_SCHEDULE).toInt());
         #endif
 
         #if MQTT_SUPPORT
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -733,7 +733,7 @@ void _relayWebSocketSendRelays(JsonObject& root) {
     JsonArray& pulse_time = relays.createNestedArray("pulse_time");
 
     #if SCHEDULER_SUPPORT
-        JsonArray& restore_last_schedule = relays.createNestedArray("restore_last_schedule");
+        JsonArray& relayLastschedule = relays.createNestedArray("relayLastschedule");
     #endif
 
     #if MQTT_SUPPORT
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -34,7 +34,6 @@ typedef struct {
     unsigned long change_time;  // Scheduled time to change
     bool report;                // Whether to report to own topic
     bool group_report;          // Whether to report to group topic
-    unsigned char restore_last_schedule;  // Boolean, Restore to last schedule or not
 
     // Helping objects
 
@@ -657,7 +656,6 @@ void _relayConfigure() {
             //set to high to block short opening of relay
             digitalWrite(_relays[i].pin, HIGH);
         }
-        _relays[i].restore_last_schedule = getSetting("relayLastschedule", i, 1).toInt();
     }
 
     #if MQTT_SUPPORT
@@ -733,7 +731,10 @@ void _relayWebSocketSendRelays(JsonObject& root) {
     JsonArray& boot = relays.createNestedArray("boot");
     JsonArray& pulse = relays.createNestedArray("pulse");
     JsonArray& pulse_time = relays.createNestedArray("pulse_time");
-    JsonArray& restore_last_schedule = relays.createNestedArray("restore_last_schedule");
+
+    #if SCHEDULER_SUPPORT
+        JsonArray& restore_last_schedule = relays.createNestedArray("restore_last_schedule");
+    #endif
 
     #if MQTT_SUPPORT
         JsonArray& group = relays.createNestedArray("group");
@@ -750,7 +751,10 @@ void _relayWebSocketSendRelays(JsonObject& root) {
 
         pulse.add(_relays[i].pulse);
         pulse_time.add(_relays[i].pulse_ms / 1000.0);
-        restore_last_schedule.add(_relays[i].restore_last_schedule);
+
+        #if SCHEDULER_SUPPORT
+            restore_last_schedule.add(getSetting("relayLastSchedule", i, SCHEDULER_RESTORE_LAST_SCHEDULE).toInt() == 1);
+        #endif
 
         #if MQTT_SUPPORT
             group.add(getSetting("mqttGroup", i, ""));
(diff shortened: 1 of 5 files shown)


mcspr:
Can you explain a bit the algorithm for finding the missed schedule? How far does it go is hard-coded?
There could also be a small issue when dst is happening, causing utc timestamp to be invalid in that instance

foxman69:
Basically what it does, is enumerating the schedules, checking the different options of them (enabled or not, UTC or not) and for each schedule check if it supposes to happen in the current day.
If so, check the minutes left to the schedule from current time (_schMinutesLeft) and if its less then 0 and bigger than the last value i saved (minimum_restore_time), then i update the saved value and the saved action. (I call this function recursively to check for the missing schedule in the days before, in case i didn't found a schedule for the specific relay at the current day).
At the end of the enumeration, it supposes to have the missing schedule's action saved in the saved_action variable.
The function supposed to be called for maximum 7 times (checking each day of the week).

About the DST, it could be problematic but i`m not sure what is the proper solution for that. (It also affects the current scheduling algorithm if there is a schedule between 1:00-2:00  or 2:00-3:00 when DST is happening).

mcspr:
Ah, thank you for that. I missed the part where it moves the timestamp back.

Maybe instead of using part of schCHeck it should be inside of schCheck inself via bool / look_back=N-days param + timestamps as params? Since lights are also handled that way and it is not dependant on relays

Following that idea, would it make sense to use `schLookBack<ID> <N-days-to-look-back>` to force initial schCheck to do the backwards search? I am not sure about adding things specific to the scheduler into relay struct & switches tab

foxman69:
I have merged the RestoreLastScheduleState function into the _schCheck function

I think its makes more sense to have things per-relay and not per-schedule as you mentioned...

foxman69:
Thank you for your code changes and popping the issue about the light provider.
I have tested the new code on my Sonoff CH4 PRO, the changes caused a few issues but i fixed them in my last commits and now everything seems to work great.
About the light provider issue (with `sch_action != 2`) I don't have such device to check it with, but i have added another check to the switch type before calling the `_schAction` and now it should work on that too, if you have such device please check it...

code/espurna/relay.ino
```diff
@@ -732,6 +732,10 @@ void _relayWebSocketSendRelays(JsonObject& root) {
     JsonArray& pulse = relays.createNestedArray("pulse");
     JsonArray& pulse_time = relays.createNestedArray("pulse_time");
 
+    #if SCHEDULER_SUPPORT
+        JsonArray& relayLastschedule = relays.createNestedArray("relayLastschedule");
+    #endif
```
mcspr: IMO this should be something like `sch_last`
Also note that neighbour PR re-does things in vue, and we are using short things here

code/espurna/scheduler.ino
```diff
@@ -208,19 +239,36 @@ void _schCheck() {
...
 }
 
 void _schLoop() {
 
     // Check time has been sync'ed
     if (!ntpSynced()) return;
 
+    if (_sch_restore == 0) {
+        for (int i = 0; i < _relays.size(); i++){
+            if (getSetting("relayLastschedule", i, SCHEDULER_RESTORE_LAST_SCHEDULE).toInt() == 1)
```
mcspr: Keep capitalization as `relayLastSchedule`?
Maybe even `relayLastSch`?

code/html/custom.js
```diff
@@ -265,6 +265,11 @@ function addValue(data, name, value) {
         name = "adminPass";
     }
 
+    // join both relayLastschedule values
+    if (name.startsWith("relayLastschedule")) {
+        name = "relayLastschedule";
```
mcspr: Same as above

mcspr:
I was also thinking about boot modes and about how this could be a separate setting for boot mode, since we already depend on relay id for doing things?
For example, replacing the `relayLastSch` setting with a new `relayBoot` type, similar to the current ones ON, OFF, TOGGLE, LOCK ON, LOCK OFF and finally this one.
Or would it be more complicated that way?

code/espurna/relay.ino
foxman69: Changed to sch_last, also fixed it in the JS...

code/espurna/scheduler.ino
foxman69: Replaced relayLastSchedule to relayLastSch as you mentioned

foxman69:
I have changed all the variables naming like you said.
About the boot mode, I have thought about it even before (during this feature implementation), in my opinion these two features are orthogonal and needs to be separate in order to give the users more possibilities of configuring their device, with these 2 features separately the user have more control of what he wants to do with each relay during boot than just give them one option at a time.

code/html/custom.js
```diff
@@ -1025,6 +1030,10 @@ function initRelayConfig(data) {
         $("select[name='relayBoot']", line).val(data.boot[i]);
         $("select[name='relayPulse']", line).val(data.pulse[i]);
         $("input[name='relayTime']", line).val(data.pulse_time[i]);
+        $("input[name='relayLastSch']", line).prop('checked', data.sch_last[i]);
+        $("input[name='relayLastSch']", line).attr("id", "relayLastSch" + i);
+        $("input[name='relayLastSch']", line).attr("name", "relayLastSch" + i);
+        $("input[name='relayLastSch" + i + "']", line).next().attr("for","relayLastSch" + (i));
```
mcspr: Why do this part? Every other relay element is unindexed and depends on parent elem ordering. Could this be just the `prop('checked', ...)` and remove the rest? Then the data gathering does not need this exceptional handling

code/html/custom.js
```diff
@@ -265,6 +265,11 @@ function addValue(data, name, value) {
         name = "adminPass";
     }
 
+    // join all relayLastSch values
+    if (name.startsWith("relayLastSch")) {
+        name = "relayLastSch";
+    }
```
mcspr: > Then the data gathering does not need this exceptional handling

code/html/custom.js
foxman69: This is what i tried to do at the beggining but what happened is that the `createCheckboxes` function was called before the `initRelayConfig` got initialized, this caused the checkbox the not function correctly.
The same issue already happens in the schedule template and you deal with it in the same way
`$(line).find("input[name='schEnabled']").prop("id", "schEnabled" + (numSchedules + 1))
        .next().prop("for", "schEnabled" + (numSchedules + 1));`

mcspr: Yikes, I see. Is createCheckboxes the culprit though? [After little refresher about how checkboxes were implemented here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#attr-for), label for=... needs unique id as it seems.
Sorry about that.

For specifics:
```
<html>
    <body>
        <div class="block" id="block-a">
            <input name="input-name" id="input-name" type="checkbox"></input>
            <label for="input-name">Name a</label>
        </div>
        <div class="block" id="block-b">
            <input name="input-name" id="input-name" type="checkbox"></input>
            <label for="input-name">Name b</label>
        </div>
        <div class="block" id="block-c">
            <input name="input-name" id="input-name-c" type="checkbox"></input>
            <label for="input-name-c">Name c</label>
        </div>
        <div class="block" id="block-d">
            <input name="input-name" id="input-name-d" type="checkbox"></input>
            <label for="input-name-d">Name d</label>
        </div>
    </body>
</html>
```
Clicking Name B will not select the closest one, inside block-b, but the one from block-a
Unique ids work as intended and this is what happening for input+label combo

code/html/index.html
```diff
@@ -1890,6 +1890,10 @@ <h2>
                 <div class="pure-u-1 pure-u-lg-1-4"><label>Pulse time (s)</label></div>
                 <div class="pure-u-1 pure-u-lg-1-4"><input name="relayTime" class="pure-u-1" type="number" min="0" step="0.1" max="3600" /></div>
             </div>
+            <div class="p-g module module-sch">
+                <div class="p-u-1 p-u-lg-1-4"><label>Restore last schedule</label></div>
```
mcspr: Note that this should probably be `pure-g`, gulp will shorten it afterwards.
Opening things locally breaks styling.

Since everything else is done, I'd rebuild the webui files and add them here + this fix.

PR #1951, opened 2019-10-17
Test: add an ability to run only one build configuration + fix related gcc warnings

mcspr:
- Instead of adding header by name directly use file paths.
- Allow to skip default configurations and use just a single one (`scripts/test_build.py -n -a test/build/basic.h -e travis-git`). <s>This feature depends on getopts and bash being nice. Should work fine (?) crossplatform, but it is tempting to use python instead. @Niek does this work ok with macos?</s> Script re-done in python instead.
- Removed travis folds

Clean-up warnings given by the -Wall, most notably fix ADS1X15 resetting device type when reading adc.
Use abs() to constrain values read as strings by using toInt() - crash handler max bytes and lights channel id


Niek:
This break on macOS unfortunately, because of the ancient Bash version it ships with:
```bash
./test_build.sh -e travis-git "-a test/build/extra/secure_client.h"
\e[1m> Selected configurations:\e[0m

test/build/basic.h
test/build/emon.h
test/build/light_dimmer.h
test/build/light_my92xx.h
test/build/nondefault.h
test/build/sensor.h
 test/build/extra/secure_client.h

\e[1m> building test/build/basic.h\e[0m
#define MANUFACTURER "TEST_BUILD"
./test_build.sh: line 44: ${cfg^^}: bad substitution
tee: illegal option -- -
usage: tee [-ai] [file ...]
```
I recommend changing ``${cfg^^}`` to something like ``echo $cfg | tr '[:lower:]' '[:upper:]'`` and use ``tee -a`` instead ``--append``.

mcspr:
Huh, that is bash 3.2 and tee is also different... then I am not sure if using shellscript is a way to go.
Python looks like an better choice not only being more sane compatibility wise, memanalyzer and ota scripts already are using it. And there is also some shared code with extra scripts anyway

Niek:
The python version looks good! Less consise but at least it's more compatible (I guess even Windows would work).

mcspr:
Windows does work.
Running in the latest Windows Terminal Preview it even properly handles ansi codes. Paths are displayed weird (`test/build\basic.h`), but the files are properly opened

PR #1956, opened 2019-10-23
Telaire T6613 Support

james-coder:
Telaire T6613 is a UART CO2 sensor that comes integrated into some other Telaire devices like Telaire T8100. (Available on ebay for $35 to $100: https://www.ebay.com/sch/i.html?_nkw=telaire+t8100)
Use J2 (the white 0.1" header) for 5V/TX/RX/GND, it connects to the T6613.

Datasheet: https://www.amphenol-sensors.com/en/component/edocman/295-telaire-t6613-t6615-co2-sensor-development-kit-user-manual/download?Itemid=8486

It has a CO2 sensor with +/- 75 ppm accuracy.

This library was based on copying bits from MHZ19 CO2 sensor code.

I've tested it here with a generic ESP8266 dev board connected to Telaire T8100 (with T6613 module.)

The black wire below is me using ground for TX/RX/GND as I was powering it with another 5V/GND adapter.

image


README.md
```diff
@@ -91,6 +91,7 @@ Since November 2018, Max Prokhorov (@mcspr) is also actively working as a collab
         * **AM2320** temperature and humidity sensor over I2C
         * **Dallas OneWire sensors** like the DS18B20
         * **MHZ19** CO2 sensor
+        * **T6613** CO2 sensor
```
mcspr: :+1: thanks for noticing the readme entry!

code/espurna/sensors/T6613Sensor.h
```diff
@@ -0,0 +1,186 @@
+// -----------------------------------------------------------------------------
+// T6613 CO2 sensor
+// https://www.amphenol-sensors.com/en/telaire/co2/525-co2-sensor-modules/321-t6613
+// Uses SoftwareSerial library
+// -----------------------------------------------------------------------------
```
mcspr: (C)opyright string?

code/espurna/sensors/T6613Sensor.h
```diff
@@ -0,0 +1,186 @@
...
+
+#if SENSOR_SUPPORT && T6613_SUPPORT
+
+#pragma once
+
+#include "Arduino.h"
```
mcspr:
```suggestion
#include <Arduino.h>
```

code/espurna/sensors/T6613Sensor.h
```diff
@@ -0,0 +1,186 @@
...
+        // Initialization method, must be idempotent
+        void begin() {
+
+            if (!_dirty) return;
+
+            if (_serial) delete _serial;
+
+            _serial = new SoftwareSerial(_pin_rx, _pin_tx, false, 32);
+            _serial->enableIntTx(false);
+            _serial->begin(19200);
```
mcspr: Add as const / define above?

code/espurna/sensors/T6613Sensor.h
```diff
@@ -0,0 +1,186 @@
...
+
+        // Descriptive name of the slot # index
+        String slot(unsigned char index) {
+            return description();
+        };
+
+        // Address of the sensor (it could be the GPIO or I2C address)
+        String address(unsigned char index) {
+            char buffer[6];
+            snprintf(buffer, sizeof(buffer), "%u:%u", _pin_rx, _pin_tx);
```
mcspr: (minor) current gcc-4 does not mark this as warning, but gcc-9 notices the 3 char size requirement. buffer should be 8

code/espurna/sensors/T6613Sensor.h
james-coder: Added.

code/espurna/sensors/T6613Sensor.h
james-coder: It is like "Arduino.h" in other .h files.
¯\_(ツ)_/¯

code/espurna/sensors/T6613Sensor.h
james-coder: For this sensor, the baud rate is fixed to 19200 and it was also an inline integer in MHZ19.

code/espurna/sensors/T6613Sensor.h
james-coder: Getting rid of warnings is great. I agree!

james-coder:
Added to Wiki: https://github.com/xoseperez/espurna/wiki/Sensors#t6613-co2-sensor

james-coder:
Sensor PDF says "max 2000ppm", but I just tried rebuilding/ota reflashing and got "Out of range" in my room. (Should I be concerned about the high CO2 value? :laughing:) I bumped it up to 5000 and rebuilt / ota-ed and sure enough now it reads 2397ppm.

mcspr:
Elevation setting? Or maybe calibration is a bit off.

There are still some commands left over to check for current mode / status, set / get elevation (+-500ft, from 0 to 5000) and check / do calibrations

james-coder:
It would be nice to expose the elevation offset command, maybe through the web interface if possible. (Not sure how that works yet.) Calibration is ABC Logic done automatically (assuming the earth stays at 400ppm :laughing:) unless you can get your hands on some CO2 gas / known concentrations. I'm not sure if there's a way to turn on / off ABC Logic, that spec sheet seems confusing about that. (It might be the exact same way as the MHZ19, since now I realize the MHZ19 is basically a rip-off version of the T6613. :laughing:)

james-coder:
Or the T6613 is very voltage sensitive and I started giving it 4.45V instead of 4.95V when I change esp 8266 boards. (Wemos D1 Mini has a pin labeled "5V", but I guess it really means "Not Quite 5V".)
My CO2 levels have returned to normal. (It matches another CO2 meter I have set up next to it.)

mcspr:
Auto-calibration as described is veeery dependent on proper ventilation, so it might not always be right, but I guess you sorted that out anyway with a different way to power it.

PR #1974, opened 2019-11-05
Use generic typechecks, OTA add followRedirects

Niek:
This PR changes the following:
* Add ``CREATE_CHECK(class, function, params...)`` macro to generate typechecks
* Change typecheck in HeapStats to use above macro
* Add typecheck in ``OTA_CLIENT_HTTPUPDATE`` module instead of ``OTA_CLIENT_HTTPUPDATE_2_3_0_COMPATIBLE`` macro
* Add typecheck in ``OTA_CLIENT_HTTPUPDATE`` module for ``followRedirects()``, enable if available


mcspr:
Based on gitter discussion, I tried adding this as a struct member:
```cpp
static constexpr std::integral_constant<bool, value> value_type{};
```
Where `value` is bool above, thus value_type is mapped to false or true type. But I can't build it locally with gcc-7 resulting in `undefined reference` errors for value_type. esp's gcc-4.8.2 builds this (...idk if it runs properly though), not sure what the problem is with this kind of use of static variable.

[Maybe something like is_detected is better for the reader?](https://godbolt.org/e#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAViAIzkA2AhgHaioCkATAEK8%2B5AM7oArqVTYQAcl4BmAm1RMxOANRd5AYQLphRUthYBbLbi4AGAIIKlKtdk06iATwAO2APqGWxYeZWtjyKyqoaWtrYAB6epAQm2GxELEwA9G6ePqR%2BRAHyFjZBBqRiqETqAHLYAO7arMLCmgDsAjbqHVW19SyNEACULW3WnaPqBpggaOIVkZHqXl4ACgBKuAAq6wCaXgBiAKqV2usAkgDylYvO2vMTU0mYTFrDo1zNACJBowB%2B1XUNwgGQy%2BYw6d2mYlmOnmi1WG22e0Ox3Oly811uREmIAeT3kL06b0%2B7U6ADd0ARMOoAGbGIgSbAQJQVFjkdRM9QAI1Z7NQgze%2BNB4IwkPROgWyzWmx2ByOpwuVxBoNe0LFvB4EDVou06hYWs0PB45E1czFHL1aqNBr1qHNBv6asVSo6Ju14Jxz0dhKCXrx3psJTKFTOTEwPUawOJHWDoYBQP5jtGQpmethkoRMuR8rRLvGmPubEeHsjLSJI0632jYcBfNaCc6SZFOdT8OlSLlqL1boLuIFJe9HyLthsRGwJncrBH10ySVMTnWgRsYmESmA6iQvS8NJYdKMzne6hwKmnEHBh6YJNSkXnBQGADotzuGYxDfJ%2BvbfUVh6Px9unJFp2ws7qNehTWEuK7jGI7juGQeSbrS9J7rmWIxHECRJCkTBTAQwheDgI7lNgmCROuuEPvSrIgYOQQjmOE5/i4HgzokwELtYZIUgshiuLsCFGBAGBsAYyFTFSqTCN4068AAbJRMlskJKTKNg3LJDqqkVFyCkVLyEZlmCeYQlCNyqgaYlMBJPhMeoqCpE8Vo5l2hYfrYA6ftYtE/pO/5MYBLFUTYHGUtkPF8QygnCeChhiJJTEyXJPDSQpBjsFIGnqdpnLpbp8bFg2xnzGq0WxZ4Nl2caKquoZ7ouaMSgpcp95hRALJcry1FuUOHnfvRU6%2BUBAXseSlLcbx270hA86JclSlSDWvYhWNj4nlBMGkHB5FGFe5gDKpDVpeojDqIa6ivh1pbeiEOBUkoc4rNKuDWOs%2BxrBA2gADLWAAyl983qB931feo6AcjQzzqKNzUgzQ758H27nsiYfhsIyaksKQwCoKyqDrqQABUeM6hjJLzY66z3Xsj3Pa9fxVrDZMU7sVMvbgECVgC9NFJ1Mj9MwsgAKwyOQbCyJYQvoLINz8II4ziJIf4hPQQtEKLPO8wA1iA/PSbePAABw8PQAAs/NG3r9CJfIzR63zMhG0LIsyGL5ASzIQvCCAljkCrTs8%2BQcCwCgGBjgQTDYBQVACegIdh6QIDAMIgHuMISDoEQ5A3UwI6kB7EAcqr5Acko6OuLISvkMHiTJGcbBMKXvvkDgSMcGHBeEEY5QECS2Aew3MTYKgkLSA3TLYFhDdMAQHI5KQrjaFgw/l4YCRl37rAcGggiCMwU8e5AvPoO4RB6EJsju3LUgMLz4%2BC8LBeu9EevSQAtNJRvqJjNoQNFbDq4MED4GIGQfU8gjrzxjuHEB9BBhSwEPwZWqs3zkE1kbeQt5mj0Gtqbfm9A9bNHkNJA0tt7Z3wbq7d2ntvaIOvrIHgQsTBaz1rePWlh9b0GkgATksMbSwesTakOduQqhvteYB2QCAGY7hISUGoMHdwodw4MHIERIBcdGA1ByO4VeNCZC30doI2QCh6DqBqMQJA6hH4vzfh/VAX8f5/wQSI3mSBjA4DjrtceJCGFG3oLeegHCMF63kJYDhet%2BYcPoPzfmAjxZnxEJQn2YskHj3kA7e%2BcTElq3IN3HOJ8QBGyAA)
My main concern is having class name and class instance in a different places. Since this is now generic macro, maybe `check` variable should be replaced with `typed_check<decltype(OBJ)>` right where it is used?

Niek:
Yes, there are many nicer solutions (``std::experimental::is_detected`` being the best looking one), but those do not compile on "older" compilers. I couldn't even get that to work on GCC < 9. But I agree that the namespace clutters quite a bit. Can you add your proposal as a commit to this PR?

mcspr:
Will do.

gcc headers are a bit tricky to navigate to c/p everything needed, but I have tried implementation from clang and small example from sol2 issue tracker, which both build just fine:
https://github.com/llvm-mirror/libcxx/blob/master/include/experimental/type_traits
https://github.com/ThePhD/sol2/issues/784#issuecomment-469722486

Current implementation was made up based on *something* in the gcc sources (i think i just grep'ed for true_type) and I wonder what is the reason for `using value_t = {true,false}_type;` instead of inheritance like sol2 example and the following proposal paper. I guess I need to read this more closely.
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4502.pdf

mcspr:
Adding `Serial.println(...)` into the appropriate types and trying to build with 2.3.0, 2.5.0 and 2.5.2 shows appropriately detected function versions simply by using `strings firmware.bin`: 
```
# 2.3.0
false_type getheapstats
false_type follow redirects
false_type wificlient arg
# 2.5.0
true_type getheapstats
false_type follow redirects
true_type wificlient arg
# 2.5.2
true_type getheapstats
true_type follow redirects
true_type wificlient arg
```

code/espurna/ota_httpupdate.ino
```diff
@@ -30,6 +31,41 @@ Copyright (C) 2019 by Maxim Prokhorov <prokhorov dot max at outlook dot com>
...
+
+template <typename T>
+void _otaFollowRedirects(const std::false_type&, T& instance) {
+}
+
+template <typename T>
+t_httpUpdate_return _otaClientUpdate(const std::true_type&, T& instance, WiFiClient* client, const String& url) {
+    if (client == nullptr) {
+        client = std::make_unique<WiFiClient>().get();
+    }
```
mcspr: btw shouln't this be back where it was initially? check for `has_WiFIClient_argument_t` there and supply it when true.
i don't remember the specifics (search is not helping...), but there were also a couple of issues with httpclient & wificlient destructor ordering. but, this may not be relevant if httpclient can handle wificlient missing and update() does not keep lingering connection / expects more actions.

Niek: I moved it inside the function because that means one place less to check the update() signature: https://github.com/xoseperez/espurna/pull/1974/files#diff-5727c97832f80d3ac6aedd50dbe75fd6L81-L90 - but yes, not sure if this is the best approach.

code/espurna/libs/TypeChecks.h
```diff
@@ -0,0 +1,70 @@
...
+    template <class Default, template<class...> class Op, class... Args>
+    using detected_or = implementation::detector<Default, void, Op, Args...>;
+
+    // ...
+    // implement the rest as needed. some things may not work though
+}
+
+using experimental_type_traits::is_detected;
+
+#endif // __cplusplus >= 201703L
```
Niek: I like this solution a lot, it's nice a clean and works across a wide range of compilers :)

code/espurna/ota_httpupdate.ino
mcspr: Note the second comment, WiFiClient is destroyed right after get().

Niek: Hmm, that was not the case previously when it was called in the ``_otaClientFromHttp()`` function? I tested OTA with these changes and I didn't run into any issues so far.

mcspr: Not the case, unique_ptr lived while parent function was active.
idk how that works exactly... effectively, wificlient is no longer allocated and anything can claim it through new / malloc

Niek: Ah, I guess it "accidentally" works because the pointer is used immediately after the .get() call. I think the best is to always create the WifiClient pointer? I'll commit it now

mcspr: Right. It could still check for arg availability though. Only cost is +2 lines, everything is known at compile time.

code/espurna/config/prototypes.h
```diff
@@ -12,6 +13,28 @@ extern "C" {
...
+        {                                                                                  \
+            using result = decltype(std::declval<detector>().template detect<T>(0));       \
+        };                                                                                 \
+        template <typename T>                                                              \
+        struct typed_check : public detector<T>::result                                    \
+        {                                                                                  \
+        };                                                                                 \
+        typed_check<CLASS> check{};                                                        \
+    };
+
```
mcspr: This should also go btw

code/espurna/libs/HeapStats.h
```diff
@@ -6,51 +6,36 @@ Show extended heap stats when EspClass::getHeapStats() is available
 
 #pragma once
 
-#include <type_traits>
+#include "TypeChecks.h"
```
mcspr: And general question - TypeTraits.h?

Niek: As long as it's only an ``is_detected`` implementation, I guess it's mostly type checks/type detections.

code/espurna/ota_httpupdate.ino
Niek: So split ``_otaClientFromHttp()`` method and call the appropriate one from ``_otaClientFrom()`` with the type check? Or you mean a different way?

mcspr: I meant right in the fromhttp, optionally creating wificlient. unique_ptr structure could be initialized with nullptr, `.get()` will return that null pointer.

Niek: Sorry bu I'm not sure if I follow - feel free to commit what you have in mind :)

mcspr: https://godbolt.org/z/yu9fxq <-
Maybe I misunderstood the question. `make_unique<T>() == unique_ptr<T>(new T())`, so we just hide make_unique under the check

Issue #1918, opened 2019-09-21
Espurna doesn't seem to recover from error in sensor DHT


JavierAder:
Hi; my data 
- build on commit cf3bd49bdecb137d206755df0e0712bc39a7f53f (2019-09-10)
- Sonoff Pow R2
- DHT22 in GPIO 4; added using custom.h
-  Thingspeak enabled

Probably I have some problem in hardware, but I am not totally sure. 
When I power my Sonoff, it reads the sensor correctly; some time later I get (via telnet) timeouts errors:

```
[306269] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[312277] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[318281] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[324532] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[330285] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[336293] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[342297] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[348305] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[354307] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)
[360564] [SENSOR] Error reading data from DHT22 @ GPIO4 (error: 3)

```

Then, if I reset Espurna via command "reset" in telnet, the error persists after the boot; but if power off and power on the sonoff, the problem disappears (at least for a time).

I  suspect the following code in DHTSensor.h

```
// Send start signal to DHT sensor
        	if (++_errors > DHT_MAX_ERRORS) {
                     _errors = 0;
                    digitalWrite(_gpio, HIGH);
                    nice_delay(250);
               }


```
because digitalWrite is not preceded by  "pinMode(_gpio, OUTPUT);". Is it possible that this is the reason why it does not seem to recover?

By the way; DHT_MIN_INTERVAL is set to 2000; it is not to low?

PD:  I will try to force the error timeout disconnecting  and connecting GPIO 4

mcspr:
> because digitalWrite is not preceded by "pinMode(_gpio, OUTPUT);". Is it possible that this is the reason why it does not seem to recover?
> PD: I will try to force the error timeout disconnecting and connecting GPIO 4

Please do also try setting pinMode.

[...]

PR #1979, opened 2019-11-08
DHTSensor fixes

mcspr:
- increase time between HIGH and LOW switch per Adafruit library example, noticed by @JavierAder (https://github.com/xoseperez/espurna/issues/1918#issuecomment-537178533) and https://gitter.im/tinkerman-cat/espurna?at=5dc2cc76196ffb4db91888fc test by @structuralB
- move pinMode one block up to handle error condition (https://github.com/xoseperez/espurna/issues/1918#issuecomment-537178533)
- update _last_ok in begin() so that _read fails with ERROR_WARM_UP for 2 seconds after sensor is initialized.

### DIFF
--- a/code/espurna/sensors/DHTSensor.h
+++ b/code/espurna/sensors/DHTSensor.h
@@ -137,13 +137,14 @@ class DHTSensor : public BaseSensor {
             unsigned char byteInx = 0;
             unsigned char bitInx = 7;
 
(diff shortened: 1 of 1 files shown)


mcspr:
I wonder if additional timeout when just starting up does anything...
Increased timeout should though, pinMode too.

Issue #1922, opened 2019-09-23
E1.31 (Streaming ACN) Protocol Support


CalcProgrammer1:
I've been using Espurna on some RGB light bulbs (AiLight) and was planning to use these bulbs as part of my Halloween/Christmas light setup this year.  The majority of my lights are WS28xx pixels controlled with ESPixelStick firmware on DIY ESP-01 breakout boards.  This firmware uses E1.31 to receive color data from PC sequencing software.

I'm wondering if Espurna could support this protocol as well.  ESPixelStick uses this E1.31 library:

https://github.com/forkineye/ESPAsyncE131

My thought on this would be a configuration page where you select which universe to listen to and then which channel for each output.  Channel values are 0-255.  For relays, could just be zero = off, nonzero = on.

CalcProgrammer1:
I'm taking a shot at adding support myself.  The ESPAsyncE131 library looks very simple to use.  I have it receiving packets from the Vixen sequencing software in unicast mode but all it's doing so far is printing a debug message when it receives a packet.  I tried in multicast mode (which is how I'm driving my ESPixelSticks) and it did not seem to receive successfully so I need to do more investigating.

mcspr:
Also note that ESPixelStick builds from git version of Arduino Core:
https://github.com/esp8266/Arduino
https://github.com/forkineye/ESPixelStick

Which may cause some subtle differences
https://github.com/forkineye/ESPixelStick/blob/178a2e90fe567cf5af986499931308b000b745d6/.travis.yml#L31

CalcProgrammer1:
I had to update Arduino Core to make it build if I remember correctly.  I had build issues, wiped, and rewrote parts of my code so I'll have to double check that it's still using the newer version.  It also puts the firmware.bin over the OTA upload size.

I was able to get it setting the four light channels on an E1.31 packet last night so next I'm going to try to make it configurable in the web UI.  Maybe something like this:

![Notes_190925_082606_009_1](https://user-images.githubusercontent.com/1051772/65605142-3a5bc480-df6e-11e9-9121-8938d3962d34.jpg)

mcspr:
Sort of like that. idk if lights module specifically needs to know about e131 presence, right now that requirement is only for data output. Can we just parse the packet and handle all of it right there?

[...]

PR #1989, opened 2019-11-12
Fix SSL builds on cores < 2.4.2

Niek:
On core 2.3.0/2.4.0/2.4.1 the WiFiClientSecureAxTLS.h header doesn't exist. The best would be to dynamically check for the header with ``__has_include("WiFiClientSecureAxTLS.h")``, but that doesn't have good compiler support. Hence the series of checks for old cores - on < 2.4.2 include the WiFiClientSecure.h header instead and do not use the ``axTLS`` namespace.

### DIFF
--- a/code/espurna/libs/SecureClientHelpers.h
+++ b/code/espurna/libs/SecureClientHelpers.h
@@ -9,7 +9,12 @@
 #if SECURE_CLIENT == SECURE_CLIENT_BEARSSL
 #include <WiFiClientSecureBearSSL.h>
 #elif SECURE_CLIENT == SECURE_CLIENT_AXTLS
-#include <WiFiClientSecureAxTLS.h>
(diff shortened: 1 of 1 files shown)


mcspr:
Do we want that though? It will build, but probably unusable anyway.
And given some examples (#1922) i'd even remove 2.4.0 references from pio altogether

Niek:
I don't think it's unusable, it seems to work on my (minimal) build. But if you decide to drop support for old cores then don't bother merging of course.

mcspr:
Running 2.3.0 or 2.4.1?

I'd keep 2.3.0 but don't bother with anything on 2.4.*, cleaning up old compat statements (as I previously thought those would be useful thing to have). 2.3.0, latest release + git best-effort-support seems easier to have.
Maybe even issue a build warning that it is not supported beyond those 3 options

PR #2001, opened 2019-11-20
Bug fix: wrong representation of switch selection in schedules

foxman69:
Because of a synchronization problem (that started to happen during recent updates from some reason) the "schedules" key was sent before the "relayState" key in the processData function.
This caused the options in "schSwitch" selector not to be initialized before the value was set during the "schedules" key handling and therefor to show the wrong selection (always showing the first option available).

I was able to reproduce the bug in the last nightly build on my Sonoff CH4 PRO

### DIFF
--- a/code/html/custom.js
+++ b/code/html/custom.js
@@ -13,6 +13,7 @@ var numReboot = 0;
 var numReconnect = 0;
 var numReload = 0;
 var conf_saved = false;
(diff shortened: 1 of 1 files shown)


mcspr:
Also noticed and fixed in the #1986 (still pending though)
I'd rather not do this with timeout, we already know when the event happens.

PR #2003, opened 2019-11-21
Relay: update setup code, rework masking

mcspr:
- add u32{to,from}String that use strtoul instead of itoa, add 0b 0o 0x prefix checks
- update boot mask to use kv storage instead of raw eeprom byte, add migration code
- update boot mask only after relay status really changes
- move relay prototypes.h entry into the relay.h, update .ino files to reference it  
- move relay header defaults (RELAY1_PIN etc.) to the relay_config.h
- refactor c struct into a simple c++ struct, add default and minimal pin+type+reset_pin based on id constructors, thus removing the need to push_back into the array.
- fix DUMMY_RELAY_COUNT=0, use default constructor via `_relays.resize(DUMMY_RELAY_COUNT + _relays.size)`

### DIFF
--- a/code/espurna/alexa.ino
+++ b/code/espurna/alexa.ino
@@ -8,6 +8,7 @@ Copyright (C) 2016-2019 by Xose Pérez <xose dot perez at gmail dot com>
 
 #if ALEXA_SUPPORT
 
+#include "relay.h"
 #include "broker.h"
 
 #include <fauxmoESP.h>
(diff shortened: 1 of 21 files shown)


mcspr:
*wrong issue the first time...*
> Skipping configuration **'EEPROM_RELAY_STATUS'** since the value of **'EEPROM_RELAY_STATUS'** is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
```cpp
uint8_t mask = EEPROMr.read(EEPROM_RELAY_STATUS);
```

> Skipping configuration **'EEPROM_RELAY_STATUS'** since the value of **'EEPROM_RELAY_STATUS'** is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
```cpp
EEPROMr.write(EEPROM_RELAY_STATUS, 0xff);
```

Codacy / cppcheck false positive. Something about the headers :/

PR #2010, opened 2019-11-24
add support for Hykker Smart Plug with power meter

reaper7:
Type: Power Plugs
Board: Hykker Smart Home Power Plug
Build flag: HYKKER_SMART_HOME_POWER_PLUG
Description: Power plug w/ power metering


mcspr:
Thanks!

PR #2016, opened 2019-11-27
Secure client / server cleanup

mcspr:
- use SecureClient in OTA HTTPUpdate (as was originally intended) @Niek 
- add warnings for asynctcp define, add warning about asynctcp and mqtt
- also a small change to the httpupdate mqtt integration to use scheduled function instead of local variable to capture url (I think it should work and I am not imagining this c++ lambda syntax)
- small comment change in pio ini to avoid confusion as to why we have all these platforms
- cleanup some module-specific defines back to avoid another confusion between secure client and asynctcp flags. while dependency flag makes sense and we have _AXTLS set as client, use module-local checks.


mcspr:
...and I'll answer my own question - that capture syntax (`[capture = Obj(arg)]() { ... }`) is from c++14:
https://godbolt.org/z/KpTUTj (-std=c++11 with capture `[copy = ....]`)
https://godbolt.org/z/QDWx6V (-std=c++11 with local var and simple capture `[copy]`)

Adding `OTA_MQTT_SUPPORT` to the secure_client build test did not show it, surprisingly, as latest core adds `-Wall` implicitly. *But*, it will show it with `-Wpedantic` in cflags, and many other interesting things

PR #2018, opened 2019-11-27
Update Arduino core in PIO

Niek:
This adds core 2.6.1 and updates the core 2.5.2 release


Niek:
And of course they just released an updated version which reverts the SDK to 2.2.x-190703 by default. I will make a separate PR

PR #2014, opened 2019-11-26
Scripts: memanalyser fixes

mcspr:
- cleanup duplication
- cleanup external dependencies
- properly use subprocess when calling pio, add debug mode
- cleanup argparse usage, use store_true actions instead of counters
- use python3 syntax and python3 shebang by default
- fix some issues with module dependencies

There are still some issues with cross-dependencies, like various _SUPPORT flags actually pulling SENSOR_SUPPORT
This also reveals some issues with memanalyser because of dependencies.h checks. For exampke, API_SUPPORT is turning itself off when WEB_SUPPORT=0
Or RFM69_SUPPORT not having any checks for disabled features, but I need more time to actually go through all of those

*upd*
RPN rules are also showing some weird numbers

PR #2020, opened 2019-11-27
Scripts: update ota.py

mcspr:
- follow up for #2014 , drop py2, try to follow pylint advice (not everything yet), restructure script into functions, fix argparse
- follow up for #1526, use threading helpers and zeroconf discovery through class object
- change default behaviour to print immediately when device is discovered, make --sort optional
- slightly increase timeout and add --timeout NUM


mcspr:
Note about the https://github.com/xoseperez/espurna/commit/7bf7aa690637e6852a6298fda7daac0594b31bf3 merge change. Using `//` to avoid nesting ints

per https://docs.python.org/3.8/tutorial/introduction.html#numbers:
> Division (/) always returns a float. To do floor division and get an integer result (discarding any fractional result) you can use the // operator; to calculate the remainder you can use %

PR #2028, opened 2019-11-28
OTA and system callbacks adjustments

mcspr:
- add nofuss custom reset reason
- nice_delay instead of ticker for arduinoota success callback, since the library will reset the board immediately after the callback returns
- move system.ino prototypes into system.h, alias cb type
- naming change for system rtcmem


mcspr:
Codacy false positive:
https://trac.cppcheck.net/ticket/8909
https://trac.cppcheck.net/ticket/9226

Issue #1998, opened 2019-11-19
Support for new Gosund SP111 and SP112


knopserl:
Has anybody sucess and a tested configuration for the new Gosund SP111 and or SP112?
Therey are very attractive and already working with Tasmota but I rather wanzted to have all my plugs working with Espurna.
Blitzwolf BW SHP2 does not work but SP112 is equal to a Blitzwolf BW SHP5 and Gosund SP111???

mcspr:
I guess the existing `BLITZWOLF_BWSHPX` / `blitzwolf-shpx` profile should work?

Googling for any info on these, I found tasmota device listings with both devices present:
https://blakadder.github.io/templates/gosund_SP111.html
https://blakadder.github.io/templates/gosund_SP112.html

GPIO pins are decoded through reading this enum (starting from 0 aka GPIO_NONE / unused pin):
https://github.com/arendst/Tasmota/blob/63b001021c5166b63078f782f177b6412dc7718f/tasmota/tasmota_template.h#L26

knopserl:
Thanks I found at least a solution for the SP111. It can use BLITZWOLF_BWSHPX but the pin-assignment for CF1 14 does not work, it is 4.
The SP112 does not work, as it has a new power monitor chip with serial communication. Tasmota has a driver for it. So it would require some extention (using code from tasmota?)

mcspr:
Sorry, I missed the last comment.
Could you please post the required defines?

Tasmota config for SP112 uses CSE7766 connected as receive-only which we already support, Sonoff POWr2 was the first to use it

JoanMCD:
I'm also working at this device - as additional info I found: 
How to program
https://www.malachisoord.com/2019/11/24/flashing-custom-firmware-on-a-gosund-sp111/

knopserl:
it would be very helpful and extend the flexibiliy of espurna, if we could change pins after flashing such as tasmota is doing with the profiles.
So the SP111/Blitzwolf BW-SHP5 vendor exchanged recently the CF1 pin from 14 to 4 and the power monitor did not work. If that could be "soft-changed" I could have simple changed that after flashing. Maybe that is worth to make a seperate/new issue? What do you think?

mcspr:
[...]

PR #2029, opened 2019-11-28
add support for BW-SHP5

aligator:
I added support for the BW-SHP5. It is very similar to the other BW-SHX devices but has a different pin layout, uses the CSE7766 for the sensors and has two relays. One for the power plug and another one for two USB ports.

Everything is working.

There may exist other brands of this device. (e.g. https://www.amazon.de/Gosund-USB-Anschl%C3%BCssen-%C3%BCberspannungsschutz-Stromverbrauch-Fernsteurung/dp/B07PMW88L7/ref=pd_sbs_23_2/262-8978108-0107566?_encoding=UTF8&pd_rd_i=B07PMW88L7&pd_rd_r=34e3944d-8685-48c7-a828-11e4c3305c38&pd_rd_w=OdbHi&pd_rd_wg=Ls2mu&pf_rd_p=184816e4-edb5-4587-8faf-776e0027d8d1&pf_rd_r=P80PPE7CA07MFBJ8TDNJ&psc=1&refRID=P80PPE7CA07MFBJ8TDNJ looks very similar, but I havent tested it)


mcspr:
Thank you!

Speaking of Gosund, SP111 & SP112 models (#1998) are a new revision of those boards

PR #2030, opened 2019-11-29
Hykker Smart Home Power Plug update

reaper7:
moved device from Wall Socket to Power monitoring devices,
update device image


mcspr:
Yes, that makes more sense, thanks!

PR #2048, opened 2019-12-09
Refactor get/set/del/hasSetting

mcspr:
- change`...Setting(key, index, default)` ->`...Setting({key, index}, default)` via helper struct. Normal `...Setting` without index works the same
- getSetting will deduce return type from defaultValue. Empty default will return String, special treatment for `const char*`. `bool` is no longer `.toInt() == 1`, but `getSetting("key", false)`. `getSetting("key")` does not specialize.
The general idea for this approach is to have conversion functions specialised in a single place.
- continue with configuration prototypes, move WIFI1_..., WIFI2_... etc. into methods (preserving flash-string status)
- directly use defined wifi settings. webui now knows about built-in settings and wont allow deletion (since the setting will come back anyway. previously, on boot. now, right after reload callback)
- cleanup type narrowing in for-loops

Migrate part is still a tricky one, since modules will know better about which types of settings are used and which ones are not. Keeping in mind that defines are still there, module itself can generate current configuration and provide it to user, so it is not required to overwrite any settings like that.

Latest update also adds conversion template struct override to provide custom conversion functions.
One way is to specialise existing `settings_convert_t` based on `settings_internals.h` examples. Another one is to provide a second argument to `getSetting<..., TConvert>` template. Not sure that the last option is strictly needed right now, but I wonder if some special conversions can be covered by that. For example, wifi sleep or secure client settings.


mcspr:
It feels like migrate configuration block needs to go completely. It is easier to track settings recording per-module instead of per-board and it will manage defaults without having each board duplicate getSetting({blah, index}, _default(index))

Also added wifi management without overwriting settings:
https://gitter.im/tinkerman-cat/espurna?at=5e1c26270aba1f0d80263391

PR #2061, opened 2019-12-14
Influxdb: fix sensor send frequency, use async client, send data in batches

### LINKED ISSUE #2044: BROKER_REAL_TIME=1 causes crashes in the INFLUXDB module
The problem is with POW R2, it crashes few minutes after reboot when BROKER_REAL_TIME is set to 1. In web debug i see INFLUXDB messages in every 5-6 seconds.

_Originally posted by @MihataBG in https://github.com/xoseperez/espurna/issues/1968#issuecomment-562615803_

---

90% certain this is related to the issue with ESPAsyncTCP, fixed by the https://github.com/me-no-dev/ESPAsyncTCP/commit/75c2513c57ac8415eab3670dba5c64e73090f5ee

I don't really remember why exactly I skipped the update, besides the ASYNC SSL build problem. Need to reassess this again.
Or, do things like https://github.com/xoseperez/espurna/pull/1806 and just use AsyncClient directly.

mcspr:
fix #2044 

- use separate Broker instances for sensor read and report
- use async client directly
- send multiple measurements in a single request
- quote string measurements

Avoiding SyncClient for now, since there is no clear benefit of using it over WiFiClient. Move to async client directly (possibly this will be a base for generic client object). For helper object to work it needs to:
- have easily accessible buffer data structure for direct writes. control allocation, maybe allow persistent structures instead of creating / destroying everything each connection.
- allow writes to happen asynchronously, we don't need to block everything if for N seconds.
 probably via a separate method, to keep existing Client interface
- rx buffer should be configurable
Also keeping in mind possible migration to basic sockets.


code/espurna/broker.h
```diff
@@ -49,7 +50,10 @@ TBrokerCallbacks<TArgs...> TBroker<type, TArgs...>::callbacks;
 // --- Some known types. Bind them here to avoid .ino screwing with order ---
 
 using StatusBroker = TBroker<TBrokerType::STATUS, const String&, unsigned char, unsigned int>;
-using SensorBroker = TBroker<TBrokerType::SENSOR, const String&, unsigned char, double, const char*>;
+
+using SensorReadBroker = TBroker<TBrokerType::SENSOR_READ, const String&, unsigned char, double, const char*>;
+using SensorReportBroker = TBroker<TBrokerType::SENSOR_REPORT, const String&, unsigned char, double, const char*>;
```
mcspr: A note to the reader:
The very first arg is needed to distinguish between two otherwise identical templates, without it both ReadBroker and ReportBroker would be the exact same class.

PR #2063, opened 2019-12-15
KingArt WiFi Curtain Switch

AlbertWeterings:
This will add support for the KingArt Wifi Curtain Switch. There are no buttons in the web interface as I could not figure out how to create them (any help on that is welcome).

For now the switch can be controlled over MQTT. I use this flow in Node-Red for it.
```json
[{
    "id": "c07d4f67.4ec8e8",
    "type": "ui_button",
    "z": "c37d0c69.43b2a8",
    "name": "",
    "group": "154b15df.ef346a",
    "order": 3,
    "width": "2",
    "height": "1",
    "passthru": false,
    "label": "Stop",
    "color": "",
    "bgcolor": "#333333",
    "icon": "fa-pause-circle-o",
    "payload": "pause",
    "payloadType": "str",
    "topic": "",
    "x": 470,
    "y": 120,
    "wires": [
        ["3fd9ac4e.68c4c4"]
    ]
}, {
    "id": "1fe35685.fbc771",
    "type": "ui_slider",
    "z": "c37d0c69.43b2a8",
    "name": "slider",
    "label": "",
    "group": "154b15df.ef346a",
    "order": 1,
    "width": 0,
    "height": 0,
    "passthru": false,
    "topic": "",
    "min": "0",
    "max": "100",
    "step": 1,
    "x": 470,
    "y": 40,
    "wires": [
        ["c1808af5.eef7b8"]
    ]
}, {
    "id": "3fd9ac4e.68c4c4",
    "type": "mqtt out",
    "z": "c37d0c69.43b2a8",
    "name": "",
    "topic": "{hostname}/curtain/set",
    "qos": "",
    "retain": "",
    "broker": "43512b83.f7f624",
    "x": 780,
    "y": 120,
    "wires": []
}, {
    "id": "c1808af5.eef7b8",
    "type": "delay",
    "z": "c37d0c69.43b2a8",
    "name": "",
    "pauseType": "timed",
    "timeout": "5",
    "timeoutUnits": "seconds",
    "rate": "1",
    "nbRateUnits": "5",
    "rateUnits": "second",
    "randomFirst": "1",
    "randomLast": "5",
    "randomUnits": "seconds",
    "drop": true,
    "x": 630,
    "y": 40,
    "wires": [
        ["3fd9ac4e.68c4c4"]
    ]
}, {
    "id": "99df64d2.62c48",
    "type": "ui_button",
    "z": "c37d0c69.43b2a8",
    "name": "",
    "group": "154b15df.ef346a",
[...]

PR #2065, opened 2019-12-16
Add support for the LSC LED LIGHT STRIP

sehraf:
This LED strip is sold by ACTION in Germany and comes with a Tuya (TYWE3L) chip. It can be "freed" by using tuya-convert. It has (yet another) IR remote control that uses another set of codes. Which is added, too. The color codes are copies from SET 1 and 2. 
I kind of think that they don't match my remote control but since i'm using my smart home system for controlling, i don't care enough to figure out the correct color codes.

The circuit board supports also a CW channel on PIN 5 (which is not used on by version), nevertheless I've added a comment in case anybody has a similar device with CW.

cpu-pinout


mcspr:
Thanks!

sehraf:
Tanks for merging this. Sadly you cut off the ota lines from platformio.ini https://github.com/xoseperez/espurna/pull/2065/commits/a8253ae06fdc88452cf4e1335231de289dce60f6#diff-25ef070b6156fdfa35dd363916f0f98aR1713

mcspr:
Yep, git merge is a weird at the end of file for some reason.
Will fix while merging the last device in the queue - #2063

alexsahka:
Any build 1M firmware for this module? Thanks!

PR #2096, opened 2020-01-03
Bug fixes in latest release

foxman69:
Fixed a few bugs:

1) You forgot to define the data-settings-max value in the html, it caused "Max number of schedules reached" alert message
2) Because the change of the use of `numSchedules` you mistakenly used the defenition of the function instead of the value (Line 1089 in the js)
3) You incremented the `schedules `variable even thou it already starts from 0...

### DIFF
--- a/code/html/custom.js
+++ b/code/html/custom.js
@@ -1086,7 +1086,7 @@ function addSchedule(values) {
         values = {};
     }
 
-    var tabindex = 200 + numSchedules * 10;
+    var tabindex = 200 + schedules * 10;
     var template = $("#scheduleTemplate").children();
     var line = $(template).clone();
 
--- a/code/html/custom.js
+++ b/code/html/custom.js
@@ -1030,7 +1030,7 @@ function addNetwork(values) {
         values = {};
     }
 
-    var tabindex = 200 + numNetworks * 10;
+    var tabindex = 200 + number * 10;
     var template = $("#networkTemplate").children();
     var line = $(template).clone();
     $(line).find("input").each(function() {
--- a/code/html/custom.js
+++ b/code/html/custom.js
@@ -1102,10 +1102,10 @@ function addSchedule(values) {
     $(line).find(".button-del-schedule").on("click", delSchedule);
     $(line).find(".button-more-schedule").on("click", moreSchedule);
 
-    var schUTC_id = "schUTC" + (schedules + 1);
+    var schUTC_id = "schUTC" + schedules;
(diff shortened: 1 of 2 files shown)


mcspr:
> You forgot to define the data-settings-max value in the html, it caused "Max number of schedules reached" alert message

This does not have any effect?
https://github.com/xoseperez/espurna/blob/62ad7da332f3f904ad8241a2b738be9820b196e4/code/html/custom.js#L1814-L1815
https://github.com/xoseperez/espurna/blob/62ad7da332f3f904ad8241a2b738be9820b196e4/code/espurna/scheduler.ino#L35-L36

Nice find about the func though! Function definition in expression is somehow passable issue for the js runtime :/ While iterating on that pr it somehow passed both local and codacy's eslint.

IDs seem to only be used for indexing and label ids? I don't fully understand the tabindex thing across the webui, because [MDN points out](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex) that it should be extremely rare thing to manually change it all over the place. My guess is that it has something to do with the menu, but I wonder if that can be more generic (+ ported out for Vue PR, since there are hardcoded values too)

foxman69:
I have checked about the `data-settings-max` again, after a factory reset it works correctly
maybe from some reason it didn't initialized in the websocket after the update, anyhow I think its still better to hard-code the attribute in the html to prevent synchronization problems.

I have added another fix in my last commit about another misuse of function as a variable, JS is just a very forgiving language when it comes to arithmetic operations on variables this is way it didn't pop as an error.

About the IDs as you mentioned its only for labeling, but still in all the other places in the code you start from 0 so it think its better keeping the convention.

The `tabindex` variable determine the focus order of the elements on the page when you press "tab" button on the keyboard, before the fix it caused the `tabindex` value to be `NaN` all the time.

mcspr:
Hm. So if the whole scheduler ws packet never gets sent for whatever reason, only then we don't have `max` set. We would still have scheduler tab present, since `schVisible` key is sent separatly.
Have you checked the ws data log in the Network tab of web development tools when that happened? In that case, you'd overwrite the existing settings, since you never have parsed them to begin with and schedules panel should have 0 schedules?

I'd rather defaults stay in one place to more easily track them e.g. like html placeholder for wifi dns setting was 8.8.8.8 for some reason, while this was not the default value in code.

`tabindex` stuff is more of question as to why we need to manually do it in every function (not only scheduler) and can the menu be the only hard-coded one. But the fix is correct, no problem there.

foxman69:
Unfortunately i haven't checked the log while it happen and i cannot reproduce it again:(

Issue #1498, opened 2019-01-19
platformIO OTA update fails on password containing `$@`


sarming:
**Bug description**
OTA update via PlatformIO fails on some complex passwords.
After some investigations I believe the `$@` combination is to blame.
In the debug output below `123456x$@` becomes `123456x`.

**Steps to reproduce**
```
$ export ESPURNA_AUTH='123456x$@'
$ echo $ESPURNA_AUTH
123456x$@
$  pio run -e wemos-d1mini-relayshield-ota -t upload
...
Uploading .pioenvs/wemos-d1mini-relayshield-ota/firmware.bin
20:20:13 [DEBUG]: Options: {'esp_ip': '192.168.1.199', 'host_port': 18914, 'image': '.pioenvs/wemos-d1mini-relayshield-ota/firmware.bin', 'host_ip': '0.0.0.0', 'auth': '123456x', 'esp_port': 8266, 'spiffs': False, 'debug': True, 'progress': True}
20:20:13 [INFO]: Starting on 0.0.0.0:18914
20:20:13 [INFO]: Upload size: 468544
20:20:13 [INFO]: Sending invitation to: 192.168.1.199
Authenticating...FAIL
20:20:13 [ERROR]: Authentication Failed
*** [upload] Error 1
```

**Device information**
```
[544809] [MAIN] ESPURNA 1.13.4-dev
[544809] [MAIN] xose.perez@gmail.com
[544809] [MAIN] http://tinkerman.cat

[544809] [MAIN] CPU chip ID: 0x1C1953
[544812] [MAIN] CPU frequency: 80 MHz
[544815] [MAIN] SDK version: 1.5.3(aec24ac9)
[544820] [MAIN] Core version: 2.3.0
[544823] [MAIN] Core revision: 159542381
[544826] 
[544827] [MAIN] Flash chip ID: 0x1640C8
[544830] [MAIN] Flash speed: 40000000 Hz
[544834] [MAIN] Flash mode: DOUT
[544839] 
[544840] [MAIN] Flash size (CHIP)   :  4194304 bytes / 1024 sectors (   0 to 1023)
[544845] [MAIN] Flash size (SDK)    :  4194304 bytes / 1024 sectors (   0 to 1023)
[544855] [MAIN] Reserved            :     4096 bytes /    1 sectors (   0 to    0)
[544859] [MAIN] Firmware size       :   468544 bytes /  115 sectors (   1 to  115)
[...]

PR #2099, opened 2020-01-05
PIO: remove -ota envs, handle OTA condition in extra script

mcspr:
Detect uploader protocol from our configuration (ESPURNA_IP and ESPURNA_AUTH) with any environment (cc @Niek since you had to modify .ini to fit this use case, but I can't find the mention and I think this was ~6month ago...)
Modify uploader arguments through python / SCons, hopefully help with #1498 down the line

Kind-of dangerous, but that code is not changed very often. Referencing platform code:
https://github.com/platformio/platform-espressif8266/blob/develop/builder/main.py
`UPLOADERFLAGS` seems to be the needed variable, although it is replaced entirely in that piece. The \*magic\* `UPLOAD_FLAGS` is supposed to simulate `upload_flags = ...` ini variable, combining that replacement with it's contents

Generic -ota envs are removed, **but** not esp8266-SIZE-ota
Those are still there, since we still need them for ota.py and theoretical shared env building. So, also added:
esp8266-latest-SIZE-ota 
esp8266-git-SIZE-ota 
(same as travis envs)


Niek:
Looks nice and this cleans up the platformio.ini file a lot! But what is the reason to keep the esp8266-size-ota envs? I think it would be easier to remove those and edit the ota.py to use the regular (non-ota) env instead?

mcspr:
BTW does it work?: )

I mean, there is no esp8266-size to begin with, dash ota only purpose was to be called via the ota script (or if someone read https://github.com/xoseperez/espurna/wiki/OTA#manually-driven-ota-updates note at the end). At least `ESPURNA_BOARD=<board-value-from-the-device>` does the trick, no need to magically substitute `NODEMCU_LOLIN` for `nodemcu-lolin`.

And it is also a followup for this forum question:
https://community.platformio.org/t/build-cache-dir-will-not-share-object-files-between-envs/10011 Since we can avoid cache stuff altogether and really re-use build environment without external files (like travis build tests do by changing custom.h)

PR #2101, opened 2020-01-07
Alarm support, pattern for GPIO pin writes

### LINKED ISSUE #2081: Flip/Flop or slow pulse output while relay is in ON state.
I have an `D1 mini` device connected to a buzzer on one of the GPIOs.

Is there a way to have the GPIO pulse or flip/flop with a set duty cycle, say 1 second on, 1 second off while the 'RELAY/GPIO' state is ON?

I currently use a rule in openHAB to switch the GPIO output ON/OFF so the buzzer pulses, but it would be nice to just set the device to ON and the `D1 mini` do the pulsing.

mcspr:
WIP?
resolve #2081 @davebuk

`alarmGPIO<N>` for GPIO setting
`alarmInv<N>` for GPIO inverse mode (where HIGH is OFF and LOW is ON)
`alarmPattern<N>` for ON OFF pattern as discussed in the issue (`ms ON,ms OFF, N repeats`), default string is `1000,1000,0`

`<N>` is relay ID. When relay is on, alarm starts up.
 
There is no disarming though. While relay action only "Arms" the alarm, app is configured to check each alarm on each event loop iteration and trigger when it is armed.
I wonder if something can lock or unlock the switch instead of the alarm?
Or, there should be an additional lock for trigger itself?


davebuk:
I was thinking the other day about how espurna controls the `LED_MODE` functions. When a device is in `LED_MODE_WIFI` the GPIO assigned to the LED makes that LED blink. This is what I am looking for as my 'ALARM BUZZER ON' state making the buzzer connected effectively "blink". It would be controlled by an 'ON' command via mqtt, report its state to mqtt and bind `ALARM_BUZZER` to an espurna `RELAYx`. Similar to https://github.com/xoseperez/espurna/wiki/Buttons-and-switches#binding-buttons-relays-and-leds 

openHAB (or any other control REST, MQTT, TELNET etc) would turn 'ON' `ALARM_BUZZER` in either **my** 'ALARM ARMING or ALARM TRIGGERED' state scenario and and the bound `RELAYx` would "blink" the GPIO logic ON/OFF repeatedly while `ALARM_BUZZER` is 'ON'. Once **my** alarm is in either the 'DIS-ARMED or ARMED' state, the `RELAYx` would get turned 'OFF' and this controls that GPIO back to an 'OFF' logic.

mcspr:
How about this then. Effectively, switch state now controls either ARMED or TRIGGER state.
`<N>` no longer maps to the switch id, but is just a basic alarm id number so that there could be multiple alarms.
`alarmArm<N>` sets a switch ID, starting from 0, for Alarm's ARMED state.
`alarmTrigger<N>` sets a switch ID to trigger the alarm, but it only does does that when alarm is ARMED.

Previously mentioned RPN can change switch state based on sensor data. Or just a basic `relay/<id>/set` payload to switch ON or OFF
*(Or, to be fancy, maybe this needs customised payload string. There are custom strings that were added in #1889, but unfortunately those can only affect all relay payloads and not just one specific ID)*

And btw, pattern thing is not limited to two triples, it can be as big as ESP RAM can hold.

davebuk:
Please excuse my coding ignorance! The code in the commits is mostly over my head. :-) 
Just to clarify, switch ID `alarmArm<N>` and `alarmTrigger<N>` are two separate virtual switches that can be either ON or OFF. Each switch ID can be bound to a `RELAYx`. 

Each one of the bound `RELAYx` outputs will then pulse the GPIO HIGH/LOW while the switch ID (`alarmArm<N>` or `alarmTrigger<N>`) is in the 'ON' state, inline with the ON/OFF/REPEAT pattern defined. The bound `RELAYx` and switch ID (`alarmArm<N>` or `alarmTrigger<N>`)  will then be in an 'OFF' state when 'OFF'.

mcspr:
> Just to clarify, switch ID alarmArm<N> and alarmTrigger<N> are two separate virtual switches that can be either ON or OFF. Each switch ID can be bound to a RELAYx.

Right, both need to be a valid relay (aka switch, [names are hard](https://en.cppreference.com/w/cpp/language/switch)) ID, virtual or real.

> Each one of the bound RELAYx outputs will then pulse the GPIO HIGH/LOW while the switch ID (alarmArm<N> or alarmTrigger<N>) is in the 'ON' state, inline with the ON/OFF/REPEAT pattern defined. The bound RELAYx and switch ID (alarmArm<N> or alarmTrigger<N>) will then be in an 'OFF' state when 'OFF'.

`alarmGPIO<N>` is used to output the pattern. Relay GPIO, if it happens to be a real relay, will do the same thing as it did before. The pattern will only happen if alarm is both armed and triggered, requiring both switches to be ON.

davebuk:
Sounds good. I guess I would end up with a configuration similar to:

```
#define RELAY2_TYPE                   RELAY_TYPE_INVERSE  //Bound Buzzer output
    
#define ALARMARM1_RELAY          1     //ALARMTRIGGER association.        
#define ALARMGPIO1_PIN          1    //GPIO pin (1 in my current setup).
#define ALARMTRIGGER1_RELAY          2   //Relay number to bind to. Similar to a button/relay.
#define ALARMPATTERN1          100,100,0  //ON, OFF, REPEAT pattern.
```
In my openHAB system, switching `ALARMTRIGGER1` ON while `ALARMARM1` is ON, the `ALARMGPIO1` pin is pulsed as per `ALARMPATTERN1`. Once my system is "ARMED" `ALARMTRIGGER1` would go OFF and stops the GPIO pulsing the buzzer. If the openHAB alarm system is triggered, `ALARMTRIGGER1` is returned back to ON. Once I have fully "DIS-ARMED" the alarm, `ALARMTRIGGER1`  and `ALARMARM1` are both returned to OFF.

mcspr:
Yes. Order matters though, I've added a check for armed status before triggering anything so the system wont go into "triggered" mode when not "armed". So another question is - should the "trigger" relay be flipped back by the alarm module to represent that it did nothing?

Note that I followed the original idea with relays to map the state, perhaps I am overthinking it and it can simply be a different action MQTT /API / CLI topic:
https://www.home-assistant.io/integrations/alarm_control_panel.mqtt/
Sending `armed`, `disarmed`, `triggered` etc. payload to `home/alarm/<N>/set`
*edit:* and ofc announcing same states with `home/alarm/<N>`

davebuk:
From the setup I currently have, `arming/armed/disarmed/triggered` etc are all dealt with by openHAB as virtual switches. For my idea, I just wanted the buzzer to pulse during the exit delay as a notification sound. It would also pulse during the entry delay time to notify someone the alarm has been “TRIGGERED” and they need to "DISARM" the system. A pulsed buzzer sounds nicer than the continuous tone. :-)

If it is easier for you to develop an almost stand alone, espurna based alarm system software module, I’ll fit my requirements in around your design. Personally, I don’t need to use the RPN rules using the `EVENT` sensor as a trigger, openHAB does this using its rules and mqtt commands. I was just looking to be able to turn ON and OFF the pulsed output for a given GPIO when the alarm is “ARMING” or “ARMED and TRIGGERED”.

mcspr:
So, is there a need for 2 switches then? It is just a matter of restricting things to go from DISARMED to TRIGGERED, avoiding intermediate ARMED state. Which can be avoided by using a single state command entrypoint.
*edit:* Or why even have switches at all, even one, and just have a separate topic(s) for all interactions.

I thought the goal here was to delegate as much as possible to the device itself, thus knowledge of alarm states and not being just a buzzer engine :). With RPN it can still be useful in case MQTT and / or WiFi break? Or in case of several different ARMED modes: "At home", "Night", "Away" etc.
But I guess "buzzer" part can be standalone to allow both external and internal interactions (e.g. also previously discussed `buzzer` topic with patterns as message payload).

---

*Note-to-self-or-someone-else-reading:* HASS doc for mqtt.alarm_control_panel (or base alarm_control_panel too) is not very descriptive of how things are expected to work, only thing to go on is state transitions from test suite:
(note `assert` strings with `STATE_ALARM...`, names are pretty self-explanatory)
https://github.com/home-assistant/home-assistant/blob/e6388e186ce97b48853f1f196193018cc4c4e109/tests/components/manual/test_alarm_control_panel.py

davebuk:
Sorry if I have sent you down a different path but yes, a single switch buzzer engine. :-) At the moment openHAB does the buzzing by sending ON/OFF commands to an mqtt controlled espurna relay definition. When the alarm is in a state that I want that output switched ON, it does a 1 second ON, 1 second OFF sequence. This is the part I was looking to remove from openHAB and have espurna produce internally.

Separately, a fully espurna driven alarm module is still a great idea. It could be a completely stand alone device and not even need to report mqtt etc (but it should). It could be placed in a remote location like a car, shed or boat and just sound a loud sounder or flash a strobe light depending on the state of directly connected magnetic trigger switches, PIR sensor or even change in LUX level. This is where the RPN rules engine would come into play.

mcspr:
Then I'm really tempted to just modify LED MQTT mode to accept pattern string (`<root>/led/<N>/set`). So, sending pattern would turn on the buzzer and sending 0 would turn it off. Or, if pattern expires, it will turn off all by itself. Led configuration already handles pin and inverse mode, just a matter of adapting relay changes for runtime config.

And Alarm stuff will follow HASS model.

mcspr:
*(email edit)*
One issue compared with relays is that you would not really know if led is active and doing something, but maybe it can optionally send some status info to `<root>/led/<N>` (since in normal mode it always is)

davebuk:
As long as when the LED output is set to ON, mqtt receives that ON/OFF status. Not the ON/OFF state of the pulse, just that the pulsed output is ON or OFF.

Can a `RELAY_BUZZ` and `LED_BUZZ` be bound to each other. When `RELAY_BUZZ` is turned ON, it sets `LED_BUZZ` on which is then tied to `LED_BUZZ_GPIO` and pulses that GPIO?

mcspr:
Yep, LED will now have the same "status" as relay does, so there is no real need to attach relay to LED to know it's status. Just hijacking manual mode seems like the easiest way of having a pattern for just that one thing, not involving other modes. 

Note that there are also `LED_MODE_FINDME_WIFI`, `LED_MODE_RELAY_WIFI` that change behaviour based on WiFi statuses... But I don't think there were any proposals to make other led modes timings configurable at all, but it feels like there also can be `ledPatternWifi`, `ledPatternRelay` etc.

davebuk:
Is there any update on the status of adding the 'GPIO buzzer engine' to the LED mode? I would love to be able to implement the option in my arm system.

mcspr:
I need to finish up with #2162 first and do the same thing with LEDs, I will merge changes from here after that.

davebuk:
Excellent, thanks.

PR #2109, opened 2020-01-14
Debug: store boot log

mcspr:
Support enabled in tree by default for testing purposes.
Buffer itself is disabled, need to mention it in the docs

Terminal command `debug.buffer` prints log buffer
`DEBUG_LOG_BUFFER_SUPPORT` to build support
`DEBUG_LOG_BUFFER_ENABLED` to enable (or `dbgBufEnabled` setting)
`DEBUG_LOG_BUFFER_SIZE` buffer size in bytes (`dbgBufSize` setting)


mcspr:
@namezero Assuming you never attached serial debugging tool to see boot log b/c it is not possible electric-wise or not easy to access at all, this branch will print the boot log with possible relay status change with `debug.log` command

Issue #376, opened 2018-01-07
GPIO16 issues


xoseperez:
Originally reported by: **Michael Murton (Bitbucket: [crazyivan359](https://bitbucket.org/crazyivan359), GitHub: [crazyivan359](https://github.com/crazyivan359))**

----------------------------------------

Hello Xose, let me start by saying what a great piece of software you have made here! I am in the process of switching all of my devices over from Tasmota to your ESPurna because it is so much more stable and easy to use.

I am having a bit of difficulty with one of my devices though. I have a Sonoff 4CH that I use to oversee a tortoise habitat. It uses 4 DS18B20 sensors and 3 DHT22 sensors. I was able to modify your code to present the Dallas sensors using their serial number instead of their magnitude type index (code modifications below if you are curious), and the DHT sensors to use DHT[type]_[gpio]. Those modifications are tested and working, however, one of the DHT sensors is on GPIO16. As I am sure you are aware, GPIO16 is an unusual pin in that it does not have an internal pull-up resistor, but in fact a pull-down resistor. I have dealt with this by using a hardware pull-up resistor, but your code uses INPUT_PULLUP when listening to the DHT's response, which I changed to INPUT. But alas I still get no readings from the sensor on GPIO16 and the device is unstable.

While I understand that this is outside the scope of your development, I was hoping you could tell me if there is something else in your code that would be blocking me from using GPIO16 in this manner, before I start debugging with serial and all.

*The following code snippets are all of the modifications I have made to use multiple sensors and report them using a device "address" instead of magnitude type index*

**sensors/DHTSensor.h**
```
#!arduino
        // String of device address
        String address(unsigned char index) {
            char buffer[8];
            snprintf(buffer, sizeof(buffer), "DHT%d_%d", _type, _gpio);
            return String(buffer);
        }
```
```
#!arduino
            //if (_gpio == 16) {
            #if defined(ITEAD_SONOFF_4CH_MURPHY)
                pinMode(_gpio, INPUT);
            //} else {
            #else
                pinMode(_gpio, INPUT_PULLUP);
            //}
            #endif
```
**sensors/DallasSensor.h**
```
#!arduino
        // String of slot # device address
        String address(unsigned char index) {
            _error = SENSOR_ERROR_OK;
            if (index < _count) {
                char buffer[40];
                uint8_t * address = _devices[index].address;
                snprintf(buffer, sizeof(buffer), "%02X%02X%02X%02X%02X%02X%02X%02X",
[...]

PR #2110, opened 2020-01-15
Fixed update of led bug and GPIO16 support

foxman69:
I bought ZSP-001 Device
https://templates.blakadder.com/ZSP-001.html

When I tried to configure it, I couldn't, because Espurna ignored GPIO16 on setup (which caused the led (LED2) not to function on that specific device).
I also saw that Xose noticed this issue before (https://github.com/xoseperez/espurna/issues/376) but didn't change anything about it.
Anyhow I have added support for the GPIO16.

What I also noticed is that when I choose "Follow Switch#" on LED configuration tab it didn't function at all, as far as I saw the only function that supposed to update the led status according to the relay status is `ledUpdate` and its not being called anywhere but from MQTT, so I have added a call to it after every relay status change. that solved the problem for me.

### DIFF
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -383,7 +383,7 @@ void _relayProcess(bool mode) {
     if (_relay_sync_locked && needs_unlock && changed) {
         _relaySyncUnlock();
     }
-
+	ledUpdate(true);
 }
 
 #if defined(ITEAD_SONOFF_IFAN02)
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -383,7 +383,6 @@ void _relayProcess(bool mode) {
     if (_relay_sync_locked && needs_unlock && changed) {
         _relaySyncUnlock();
     }
-	ledUpdate(true);
 }
 
 #if defined(ITEAD_SONOFF_IFAN02)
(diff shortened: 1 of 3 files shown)


mcspr:
Do you build with BROKER_SUPPORT disabled? See `StatusBroker` line just above in the processing. On the other side, led module will do ledUpdate() each time it receives a callback with RELAY topic

GPIO16 stuff is useful, but I am not seeing LED actually checking anything via `gpio...` methods.

foxman69:
I do compile with BROKER_SUPPORT disabled, probably that's why it didn't work for me.

About the gpio, you are right, it just took me time to make the led work, I got confused and thought its the gpio change that solved it, but the gpio change just helped me debug it in the terminal because without that change I was not able to control the 16th pin.

code/espurna/relay.ino
```diff
@@ -383,7 +383,7 @@ void _relayProcess(bool mode) {
     if (_relay_sync_locked && needs_unlock && changed) {
         _relaySyncUnlock();
     }
-
+	ledUpdate(true);
```
mcspr: I'd remove this.
`LED_SUPPORT` can hard-depend on `BROKER_SUPPORT` then? see dependencies.h & led mode (which can actually be disabled, removing `ledUpdate` symbol)

foxman69: Thats correct, just making it depend on BROKER would solve the problem👍.
its ok to remove this line...

mcspr (approved):

PR #2124, opened 2020-01-25
Thingspeak urn edit

sametflo:
A simple WebHook by modifying the Thingspeak service URL included in espurna. We can so direct those requests to the script of our choice (a PHP script in my case), and handle every notification we can imagine without having to activate any MQTT service or better in addition to the MQTT service.

### DIFF
--- a/code/espurna/thinkspeak.ino
+++ b/code/espurna/thinkspeak.ino
@@ -30,10 +30,6 @@ const char THINGSPEAK_REQUEST_TEMPLATE[] PROGMEM =
 bool _tspk_enabled = false;
 bool _tspk_clear = false;
 
-String _tspk_host;
-String _tspk_url;
-unsigned int _tspk_port;
-
 char * _tspk_queue[THINGSPEAK_FIELDS] = {NULL};
 String _tspk_data;
 
@@ -41,8 +37,25 @@ bool _tspk_flush = false;
 unsigned long _tspk_last_flush = 0;
 unsigned char _tspk_tries = THINGSPEAK_TRIES;
 
+class AsyncThingspeak : public AsyncClient
+{
+  public:
+    URL* address;
+    AsyncThingspeak(const String&);
+    ~AsyncThingspeak();
+};
+
+AsyncThingspeak::AsyncThingspeak(const String& _url) {
+  address = new URL(_url);
+}
+
+AsyncThingspeak::~AsyncThingspeak() {
+  delete(address);
+}
+
+AsyncThingspeak * _tspk_client;
+
 #if THINGSPEAK_USE_ASYNC
-AsyncClient * _tspk_client;
 bool _tspk_connecting = false;
 bool _tspk_connected = false;
 #endif
@@ -101,12 +114,7 @@ void _tspkConfigure() {
         _tspk_enabled = false;
         setSetting("tspkEnabled", 0);
     }
-    if (_tspk_enabled && !_tspk_client) _tspkInitClient();
-
-    URL _url(getSetting("tspkAddress", THINGSPEAK_ADDRESS));
-    _tspk_host = _url.host;
-    _tspk_url = _url.path;
-    _tspk_port = _url.port;
+    if (_tspk_enabled && !_tspk_client) _tspkInitClient(getSetting("tspkAddress", THINGSPEAK_ADDRESS));
 }
 
 #if THINGSPEAK_USE_ASYNC
@@ -121,9 +129,9 @@ tspk_state_t _tspk_client_state = tspk_state_t::NONE;
 unsigned long _tspk_client_ts = 0;
 constexpr const unsigned long THINGSPEAK_CLIENT_TIMEOUT = 5000;
 
-void _tspkInitClient() {
+void _tspkInitClient(const String& _url) {
 
-    _tspk_client = new AsyncClient();
+    _tspk_client = new AsyncThingspeak(_url);
 
     _tspk_client->onDisconnect([](void * s, AsyncClient * client) {
         DEBUG_MSG_P(PSTR("[THINGSPEAK] Disconnected\n"));
@@ -204,8 +212,9 @@ void _tspkInitClient() {
 
         _tspk_connected = true;
         _tspk_connecting = false;
+        AsyncThingspeak* _tspk_url = reinterpret_cast<AsyncThingspeak*>(client);
 
-    DEBUG_MSG_P(PSTR("[THINGSPEAK] Connected to %s:%u\n"), _tspk_host.c_str(), _tspk_port);
+    DEBUG_MSG_P(PSTR("[THINGSPEAK] Connected to %s:%u\n"), _tspk_url->address->host.c_str(), _tspk_url->address->port);
 
         #if THINGSPEAK_USE_SSL
             uint8_t fp[20] = {0};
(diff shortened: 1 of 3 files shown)


code/espurna/thinkspeak.ino
```diff
@@ -96,6 +103,10 @@ void _tspkConfigure() {
         setSetting("tspkEnabled", 0);
     }
     if (_tspk_enabled && !_tspk_client) _tspkInitClient();
+  
+    _tspk_host = getSetting("tspkHost", THINGSPEAK_HOST);
+    _tspk_url = getSetting("tspkUrl", THINGSPEAK_URL);
+    _tspk_port = getSetting("tspkPort", THINGSPEAK_PORT);
```
mcspr: Small issue with async client is that it is *possible* to have different host value in `_tspkPost()` and `_tspk_client->onConnect()`. `tspkPost` is expected to be called only once per request, so it would be safer to load things there?

mcspr (commented):
Is there a need for 3 variables? I know it is duplicating preprocessor vars, perhaps it could be a single `tspkUrl` combining all three?

We do have a small libs/URL.h class to parse ota url which extracts host,port and path, maybe such class can be used instead and preprocessor vars can be loaded into it.

code/espurna/thinkspeak.ino
sametflo: You're clearly right. I'm agree with it. I will do it asap. Thanks.

sametflo: You're clearly right. I'm agree with it. I pushed a new version.

code/espurna/thinkspeak.ino
```diff
@@ -29,6 +30,10 @@ const char THINGSPEAK_REQUEST_TEMPLATE[] PROGMEM =
 bool _tspk_enabled = false;
 bool _tspk_clear = false;
 
+String _tspk_host;
+String _tspk_url;
+unsigned int _tspk_port;
```
mcspr: Just copy from `URL` below? If am not mistaken, c++ creates default copy constructors which will allow to have:
```
// assuming there is new URL(const String& host, uint16_t port, const String& path) constructor
URL one {"host", 80, "/path"};
URL two {"other", 81, "/other_path"};
one = two;
```

code/espurna/thinkspeak.ino
```diff
@@ -96,6 +102,11 @@ void _tspkConfigure() {
         setSetting("tspkEnabled", 0);
     }
     if (_tspk_enabled && !_tspk_client) _tspkInitClient();
+
+    URL _url(getSetting("tspkAddress", THINGSPEAK_ADDRESS));
+    _tspk_host = _url.host;
+    _tspk_url = _url.path;
+    _tspk_port = _url.port;
```
mcspr: Nothing changed though? 

Just to clarify:
- SDK and network run in a separate light "thread"
- loop() / setup() run in a different thread, which can be paused at any time via `delay(...)` or `yield()`
- configuration callbacks run at the start of the `loop()`

Sync methods of thingspeak don't need a global URL to track things, since they start the network request and exit in the same function. Async schedules next `yield()` / `delay()` to execute network request in a system thread, thus we need a persistent variable to store host/port/path tuple. If we make a request and change configuration via configure callback, by the time system reaches network callbacks we can have a different value.

code/espurna/thinkspeak.ino
mcspr: ...or, just always using `_tspk_url` without copying.

code/espurna/thinkspeak.ino
sametflo: I understand that it is not atomic, but I use this URL for notifications to a simple PHP script. So if I lose one or more data, it doesn't matter for me. Moreover, I don't see in which case it could be annoying. However, I'm ready to make the changes you want, but I need you to be more explicit: tell me what you want me to change.

mcspr: What I was thinking is:
- parse URL in tspkFlush(), pass as argument to Post()
- sync function references arg, async copies to a global and uses that

sametflo: Hi Max, I pushed a new version. In this one, I did a AsyncThingspeak object that inherit from AsyncClient and contain the URL.

I must admit that c ++ is not my usual language, and I'm not as comfortable as in other languages. So, if you have advices to give me about this code, I am listening to all proposals.

However, I need help with the THINGSPEAK_USE_ASYNC directive: the _tspkConfigure() function which is in the general context, calls the _tspkInitClient() function which only exists in the THINGSPEAK_USE_ASYNC context. Is this a bug, or did I miss something ?

code/espurna/thinkspeak.ino
```diff
@@ -36,8 +37,25 @@ bool _tspk_flush = false;
 unsigned long _tspk_last_flush = 0;
 unsigned char _tspk_tries = THINGSPEAK_TRIES;
 
+class AsyncThingspeak : public AsyncClient
+{
+  public:
+    URL* address;
```
mcspr: Why use a pointer when it can be a class member?
Minor c++ note: if object has member `_a`, we can have `constructor(T a) : _a(a) {}` syntax.

code/espurna/thinkspeak.ino
```diff
@@ -95,7 +114,7 @@ void _tspkConfigure() {
         _tspk_enabled = false;
         setSetting("tspkEnabled", 0);
     }
-    if (_tspk_enabled && !_tspk_client) _tspkInitClient();
+    if (_tspk_enabled && !_tspk_client) _tspkInitClient(getSetting("tspkAddress", THINGSPEAK_ADDRESS));
```
mcspr, line 108: This can work, but then there should be a note somewhere that url can never change without reboot *or* on initial setup when tspk is still disabled?

code/espurna/thinkspeak.ino
```diff
@@ -193,8 +212,9 @@ void _tspkInitClient() {
 
         _tspk_connected = true;
         _tspk_connecting = false;
+        AsyncThingspeak* _tspk_url = reinterpret_cast<AsyncThingspeak*>(client);
```
mcspr: `tspk_client`?

code/espurna/thinkspeak.ino
mcspr: Yeah, sync client isn't specifically tested all the time, that is an oversight. `#if ...` preprocessor block can strip that code part.

code/espurna/thinkspeak.ino
sametflo: I wasn't use a class member simply because I had not found how to do it. Thanks for your help.

code/espurna/thinkspeak.ino
sametflo: Off course, it is clearer

code/espurna/thinkspeak.ino
sametflo: I haven't seen this note, but, isn't it logical to attach the URL and particularly the host to the client socket to ignore any changes during the request ?
However, I can confirm it, it works !

mcspr: I mean that the `tspkAddress` / `THINGSPEAK_ADDRESS` value is bound to it until we reboot, because client object is never destroyed thus `.address` is never updated again. *Unless*, we also do `_tspk_client.address = URL(getSetting(...))` somewhere. Like in `tspkPost()`, just before actual connection.

sametflo: I update address on post now. That way, there is no need to reboot.

mcspr (approved):
Thanks! I'll checkout the tree  to rebuild the web image.
Maybe will also see to the sync client

mcspr:
BTW I have changed it to support sync client too via #2140 and our SecureClient helpers (which are supposed to be magic on top of normal wificlientsecure).
Not tested though, only that it builds. And you may also note that the way we support custom CA certs is very particular.

PR #2129, opened 2020-01-29
Light: configure dimmer pins from settings

mcspr:
- `ltDimmerGPIO<index>` & `ltDimmerInv<index>` to construct / override existing settings. Add pin defaults, mapped as `GPIO_NONE`. `pwm_init()` receives resulting GPIO mapping based on settings loader. Not tested, but I am not seeing much difference atm (like situation with 0 channels)
- remove bogus `_light_channels` variable, use this name for channel array instead (...which is vector and has size(). functions always use that value anyway)

### DIFF
--- a/code/espurna/config/defaults.h
+++ b/code/espurna/config/defaults.h
@@ -666,6 +666,30 @@
 #define LIGHT_CHANNELS      0
 #endif
 
+#ifndef LIGHT_ENABLE_PIN
+#define LIGHT_ENABLE_PIN    GPIO_NONE
+#endif
+
+#ifndef LIGHT_CH1_PIN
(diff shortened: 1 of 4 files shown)


mcspr:
Comparing to v2 (again):
https://github.com/xoseperez/espurna/blob/cbd7064c5c99622204c04203de7e2e818871011c/code/espurna/migrate.ino#L283
https://github.com/xoseperez/espurna/blob/9f1429c441d2c241f979d8a763ef876460b6f9fd/code/espurna/light.ino#L1087
Not sure which logic modes were supposed to be added down the line, but I am not seeing use-case right now :/ (@xoseperez)

PR #2131, opened 2020-01-30
Added alert when websocket closes and trying to trigger it faster

foxman69:
The webpage does not recognize websocket disconnections, this can cause major problems when still viewing the webpage while it happens (Any action like turning on the switch will seem working on the webpage but actually nothing happens)

I've added an alert when `websocket.onclose` happens and to make it trigger faster (if no websocket action happens it can take forever) I've added also a ping pong function to make some noise on the websocket.


code/espurna/ws.ino
```diff
@@ -233,7 +233,9 @@ void _wsParse(AsyncWebSocketClient *client, uint8_t * payload, size_t length) {
     const char* action = root["action"];
     if (action) {
 
-        DEBUG_MSG_P(PSTR("[WEBSOCKET] Requested action: %s\n"), action);
+        if (strcmp(action, "ping") != 0) {
+            DEBUG_MSG_P(PSTR("[WEBSOCKET] Requested action: %s\n"), action);
+        }
```
mcspr: Why not
```cpp
+        if (strcmp(action, "ping") == 0) {
+            wsSend_P(client_id, PSTR("{\"pong\": 1}"));
+            return;
+        }
+
```
just once, right here?

code/html/custom.js
```diff
@@ -2136,6 +2143,15 @@ function connectToURL(url) {
...
         };
+		websock.onclose = function(evt) {
+			if(confirm("Connection lost with the device, click OK to refresh the page")) {
+				$("#layout").toggle(false);
+				location.reload();
+			}
+		}
+		websock.onopen = function(evt) {
+			ping_pong();
+		}
```
mcspr: Any reason this is not Interval?
```js
var ws_pingpong;
...
onclose() {
    clearInterval(ws_pingpong);
}
onopen() {
    ws_pingpong = setInterval(function() { sendAction("ping", {}); }, 5000);
}
```

code/html/custom.js
```diff
@@ -2136,6 +2143,15 @@ function connectToURL(url) {
                 processData(data);
             }
         };
+		websock.onclose = function(evt) {
+			if(confirm("Connection lost with the device, click OK to refresh the page")) {
```
mcspr: `window.`

code/html/custom.js
```diff
@@ -2136,6 +2143,15 @@ function connectToURL(url) {
...
+				$("#layout").toggle(false);
+				location.reload();
```
mcspr: `window.`

code/espurna/ws.ino
foxman69: The first one is because i didn't want to spam the debug log after each ping-pong (it only prints actions that are not ping...)

code/html/custom.js
foxman69: You are correct, setInterval is a better way to implement it, i will change it in the next commit

code/html/custom.js
foxman69: will fix in next commit

code/html/custom.js
foxman69: so as this...

code/espurna/ws.ino
mcspr: I mean, check action before print and return.

mcspr:
btw what happens on ota, when device reboots?
if we receiving `pong` in response to `ping`, should we use that as basis for protocol-level keepalive timeout?

code/espurna/ws.ino
foxman69: fixed it:)

foxman69:
As far as i seen during reboot and ota it works just fine, it takes a bit of time since the websocket closed till the `onclose` gets triggered, the ping pong function is just to make it happen a bit faster but maybe it need to get tested more deeply to make sure it does not interrupt with anything.
I've checked it on a few devices (Sonoff and Tuya) and it didn't cause any problems.

As far as i read online, this ping pong is the basics of keepalive in websockets, if you know a better way of doing it please suggest:)

mcspr:
So far I only found info for proxy use-cases, where intermediate closes the connection when we dont send anything for 55-60 seconds.

If the goal is to avoid that, sure that is fine. What I was asking about is whether we should have something like this:
- have setInterval run ping() function
- ping() sets setTimeout()
- when processData receives "pong", we abort the setTimeout()
- when we don't receive "pong", we treat the connection as closed

foxman69:
I'm not sure that putting a timeout on pong response is a good thing to do, because there are scenarios where the communication on the websocket is slow but the websocket itself is still connected.
I'm afraid doing so can cause a lot of false positives and disrupt the regular surf experience in the webui.

When i tested it, i tried 3 scenarios
1) Connect to the device while my phone is on wifi and then turn the wifi of (on my phone) and see what happens. (Result: the `onclose` get called immediately as it should and after the refresh it connects back regulary)
2) Doing the same as 1 but moving between two different APs (Result: same as 1)
3) Connect to the device and then plug it off the power while still viewing the webui (Result: Takes about 15 seconds in average till the `onclose` function gets called)

IMO the current implementation is fine because it don't cause any unneeded disruption to the user

mcspr (approved):
OK, we'll see then :)

otoh, only issue I can see with missing messages is webserver wsSend_P() may stumble upon the queuing issue. I will fix that some time in the future though.
*edit:* and in case tcp connection is slow and we see such long delays, it may do weird things it is true.

PR #2132, opened 2020-01-30
NTP: use sntp app from lwip on latest Cores, replace ntpclientlib

### LINKED ISSUE #1380: NtpClientLib issues / alternatives

### LINKED ISSUE #1517: Enhanced TimeZone support
Would it be possible to incorporate the following library or similar to increase timezone support, or have a flag in current DST for the Southern hemisphere?
https://github.com/JChristensen/Timezone
Thanks

### LINKED ISSUE #1629: Daylight saving time in effect a day too late.
**Is your feature request related to a problem? Please describe.**
DST gone in effect in the US yesterday, nevertheless when checking the time of my devices, they showed the previous time. Checking them now (the following day), the show the correct time.

**Describe the solution you'd like**
Internal clock should change that moment, DST goes in effect.

**Additional context**
Firmware 1.13.6-dev and 1.13.3, DST set to yes, zone US select and -5 offset of UTC for US East Coast.

mcspr:
subj

fix #1380 (yay)
fix #1517
fix #1629

TODO:
- timebroker. can it be used as scheduler replacement + rpn? experimenting with $tick1m, $tick1h.
slightly complicated because of restore-last-schedule since rpn is generic and scheduler part would need to iterate over $tick rules. Can use a different rpn context and ruleset / settings, maybe.
- fix lib_deps forcibly linking ntpclientlib + timelib. if not possible, drop 2.3.0 from supported altogether. if it does work, just switch implementations as needed. webui can have `ntplegacyVisible` hint to hide offending parts or we could also have a separate panel.
- localtime / gmtime make for some weird interactions
- timelib compat types are all over the place. uptime tries to be fancy, so elapsed... need to accept unsigned and not only time_t


mcspr:
<s>Also something of note that `ntpStartDelay` needs to be >= 15 seconds. 0 option also works though, bot nothing in-between 0 and 15</s>
Based on RFC, it has to be at least 15s. Having re-checked the code, it never enforces it. Random re-tuned a bit, since it does actually wait. Probably needs further reducing

Niek:
Nice job 👍🏻
Any improvements in build size after dropping ntpclientlib?

mcspr:
None, really. Re-enabling scheduler code, travis shows the same size for basic.h build.
+rpnlib -scheduler is ~10KB as noted in the comment
-scheduler is ~4KB, 1KB more than the comment notice
-legacy ntp code (adding tz, removing offsets) in html/js is 240 bytes total

mcspr:
In general `ntpLocal2UTC()` makes no sense if it wants to return timestamp. Timestamp itself is already UTC, where localtime modifies `tm` struct based on TZ when we use it.

Not really looking too portable, since there needs to be a code chunk using libc functions instead of timelib. I guess it can temporarily `setenv("TZ", "UTC0"); dowork(); setenv("TZ", "whatever");` if that is enough to apply the change to `localtime` result, so `minute()` / `hour()` / `day()` work. Scheduler references timestamps a lot, which needs to be refactored to use time struct

code/espurna/ntp.ino
```diff
@@ -2,264 +2,335 @@
...
-        offset *= 60;
-        setSetting("ntpOffset", offset);
+void _ntpConfigure() {
+    // Note: TZ_... provided by the Core are already wrapped with PSTR(...)
+    const auto cfg_tz = getSetting("ntpTZ", NTP_TIMEZONE);
+    const char* active_tz = getenv("TZ");
+    if (cfg_tz != active_tz) {
+        setenv("TZ", cfg_tz.c_str(), 1);
+        tzset();
+    }
```
mcspr, line 213: Since configTime re-inits sntp, set timezone through **[`tzset(3)`](http://man7.org/linux/man-pages/man3/tzset.3.html)**

code/espurna/ntp.ino
```diff
@@ -2,264 +2,335 @@
...
 
+void _ntpSetTimeOfDayCallback() {
+    _ntp_synced = true;
+    _ntp_last = time(nullptr);
     #if BROKER_SUPPORT
-        _ntpBroker();
+    static bool once = true;
+    if (once) {
+        // XXX: Nonos docs for some reason mention 100 micros as minimum time. Schedule next second in case this is 0
+        _ntp_broker_timer.once((60 - second(_ntp_last)) ?: 1, _ntpBrokerCallback);
```
mcspr: Legacy option always triggers scheduler when time is set. With this one, it always waits for minute transition.
Mostly to avoid running something on 58th second and 00th after that

code/espurna/ntp.ino
```diff
@@ -2,264 +2,343 @@
...
-String ntpDateTime() {
-    if (ntpSynced()) return ntpDateTime(now());
-    return String();
+void _ntpSetTimeOfDayCallback() {
+    _ntp_synced = true;
+    _ntp_last = time(nullptr);
+    #if BROKER_SUPPORT
+    static bool once = true;
+    if (once) {
+        _ntpBrokerSchedule(60 - second(_ntp_last));
```
mcspr: Using the same scheduling policy here as in _ntpBrokerCallback(). Mostly to avoid running something on 58th second and 00th after that, always force the timer to expire at 0 seconds.

code/espurna/ntp.ino
```diff
@@ -2,264 +2,343 @@
...
+    _ntpConfigure();
+
+    // make sure our logic does know about the actual server
+    // in case dhcp sends out ntp settings
+    static WiFiEventHandler on_sta = WiFi.onStationModeGotIP([](WiFiEventStationModeGotIP) {
+        const auto server = _ntpGetServer();
+        if (sntp_enabled() && (!_ntp_server.length() || (server != _ntp_server))) {
+            DEBUG_MSG_P(PSTR("[NTP] Updating `ntpServer` setting from DHCP: %s\n"), server.c_str());
+            _ntp_server = server;
+            setSetting("ntpServer", server);
```
mcspr, line 365: Dhcp routine will call sntp_init() whenever dhcp server sets ntp option:
https://github.com/xoseperez/espurna/pull/1828
I also wonder if this is the reason that sntp can never be optimised away

code/espurna/ntp.ino
```diff
@@ -268,13 +347,24 @@ void ntpSetup() {
...
+            _ntp_synced = true;
+            _ntpSetTimestamp(String(e->argv[1]).toInt());
+            terminalOK();
+        });
 
-    // Sets up NTP instance, installs ours sync provider
-    _ntpStart();
+        // TODO:
+        // terminalRegisterCommand(F("NTP.SYNC"), [](Embedis* e) { ... }
+        //
```
mcspr, line 393: We can't exactly speed-up sntp like NtpClientLib, so idk about this.
Changing `ntpServer` value also triggers sync

code/espurna/ntp.h
```diff
@@ -0,0 +1,52 @@
...
+*/
+
+#pragma once
+
+#include "broker.h"
+
+// TODO: need this prototype for .ino
+struct NtpCalendarWeekday;
+
+#if NTP_SUPPORT
```
mcspr, line 14: BTW some benefit of having `#if` guards in headers is that compilation will fail instead of linker when function is not implemented / supported

mcspr (commented):
Timelib helpers are working, scheduler days are in order.
Everything builds even with `lib_deps` referencing NtpClientLib, only a matter of reading platformio documentation more closely on how it parses dependencies.

bluexp1:
I am currently using 1.14.1 and ntp fails to get time from the internet, Is there any solution available to make it work with current version or I need to wait for 1.14.2 to get released?

mcspr:
download the `dev` branch and use the latest version? or nightly builds to use binaries?
I'd also suggest to use gitter or just open a new issue for any questions, closed issues / PRs are not easy to track for new discussions.

PR #2140, opened 2020-02-09
Thingspeak: cleanup

mcspr:
- fix _tspk_client shadowing in callback code
- fix sync client configuration, use correct settings chain
- refactor async client, add enum state variable instead of two bools
- refactor sync client, drop raw wificlient


code/espurna/thinkspeak.ino
```diff
@@ -256,55 +272,95 @@ void _tspkPost() {
...
-        nice_delay(100);
+    nice_delay(100);
 
-        String response = _tspk_client.readString();
-        int pos = response.indexOf("\r\n\r\n");
-        unsigned int code = (pos > 0) ? response.substring(pos + 4).toInt() : 0;
-        DEBUG_MSG_P(PSTR("[THINGSPEAK] Response value: %u\n"), code);
-        _tspk_client.stop();
+    const auto response = client->readString();
+    int pos = response.indexOf("\r\n\r\n");
```
mcspr, line 320: This whole part should be HTTPClient, since we have it readily available from `#include <ESP8266HTTPClient.h>`
At least we will know about the status code.

PR #2146, opened 2020-02-13
Use deveopment version of PlatformIO Core

ivankravets:
We would be thankful if you could use a development version of PlatformIO Core. in 99.999% it's very stable but there sometimes the cases which we didn't expect, such as https://github.com/platformio/platformio-core/issues/3377

So, if you will see any issues with PlatformIO Core-dev, please report us and we will fix ASAP. Thanks!

### DIFF
--- a/travis_install.sh
+++ b/travis_install.sh
@@ -9,6 +9,7 @@ npm_install() {
 
 pio_install() {
     pip install -U platformio
+    pio upgrade --dev
     pio platform update -p
 }
 


mcspr:
Thanks! Makes sense, will merge.

btw any idea about the 
```
In file included from /home/travis/.platformio/lib/AsyncTCP_ID1826/src/AsyncTCP.cpp:24:0:
/home/travis/.platformio/lib/AsyncTCP_ID1826/src/AsyncTCP.h:26:23: fatal error: sdkconfig.h: No such file or directory
```
?

For some reason installing ESPAsyncWebServer causes it to install AsyncTCP, which was working with 4.1.0 but now it tries to build it

ivankravets:
> In file included from /home/travis/.platformio/lib/AsyncTCP_ID1826/src/AsyncTCP.cpp

@mcspr this is our bug :( and has been fixed. We are going to release PIO Core 4.2.1 next week. See my comment https://community.platformio.org/t/pio-4-2-0-complile-error-for-asynctcp-id1826/12025/5?u=ivankravets

PR #2157, opened 2020-02-18
Config for Teckin SP23 & Maxcio W-UK007S

julianwb:
Additional devices:  I think I have it right, please LMK so I can fix if necessary, thanks

### DIFF
--- a/code/platformio_ota.ini
+++ b/code/platformio_ota.ini
@@ -246,6 +246,9 @@ extends = env:maxcio-wus002s
 [env:maxcio-wde004-ota]
 extends = env:maxcio-wde004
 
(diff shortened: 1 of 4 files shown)


code/platformio_ota.ini
```diff
@@ -246,6 +246,9 @@ extends = env:maxcio-wus002s
 [env:maxcio-wde004-ota]
 extends = env:maxcio-wde004
 
+[env:maxcio-wuk007s-ota]
+extends = env:maxcio-wuk007s
+
```
mcspr, line 251: btw note the comment at the top
but, it is still here until I fix the documentation

PR #2159, opened 2020-02-18
Extend 'pulse' to allow OFF when current below threshold

julianwb:
Speculative PR: this may or may not be acceptable, looking for comment really.

If the principle is accepted then I would be pleased to make any necessary changes for merge; if not, that's fine - I did it for me in  the first place!

### DIFF
--- a/code/espurna/relay.ino
+++ b/code/espurna/relay.ino
@@ -22,6 +22,10 @@ Copyright (C) 2016-2019 by Xose Pérez <xose dot perez at gmail dot com>
 
 #include "relay_config.h"
 
+#ifndef MIN_CURRENT_CHECK_COUNT 
+#define MIN_CURRENT_CHECK_COUNT    5
+#endif
+
 struct relay_t {
 
     // Default to dummy (virtual) relay configuration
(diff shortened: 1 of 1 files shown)


julianwb:
I realise that I have not extended the "reboot" persistence/restoration for the current setting - behaviour may be unexpected I suspect as the ON/OFF state may be persisted and restored, but not the limit or the "state".   This should be done too

mcspr:
Nice idea. I think there were some issues mentioning this use-case for this sort of thing, but other way around - when values are too high turn relay off. Not sure there was any issues regarding turning relay off when consumption drops.
I do not think pulse is a way to go though, but it is a creative approach :)

For persistence, take a look at the relay module methods at the top of the file. Rtcmem is a basic struct that is mapped to a specific region of RAM that is never erased on reboots. getSetting / setSetting are self-explanatory.

---

As mentioned in gitter, to use rpn here we would need variable to store the counter, so some kind of variable manipulation needs to happen first (as there is none atm, push, pop, peek at the very least)

e.g.
`0 pusha` -> $ra is now 0
`popa` -> $ra is now unset and 0 is on the stack
`peeka` -> $a is still set to 0, 0 is on the stack

Operator registration is very straightforward. But, I need to remember to implement this.

julianwb:
"I do not think pulse is a way to go though, but it is a creative approach :)" <- It is the lazy approach!   As I said I started with something that would work for me only .. this was simplest: though abusing the "pulse" command in particular changing the meaning of its parameter is decidedly suspect ....
A new command would be clearer .... cannot think of a name for it though!
Persistence: yep working on it presently thanks.

PR #2162, opened 2020-02-21
Button pin provider & runtime settings

mcspr:
Remove direct pin control from debounceevent, implement basic class to override digitalRead(), digitalWrite() and pinMode()
Rework debouncevent header, move locally to simplify changes (may be subject to change)
Next on TODO is to re-implement sonoff dual through this approach and remove hardware-specific changes, replacing them with a different "pin" variant. Hopefully this approach isn't completely wrong.

@Tofandel as proposed buttons changes. still no schema, but will add that after (also with proxies to getSetting for ws properties)
as was noted when changing lights code, I am not sure I want getSetting in ctor code since we can't guarantee that settings storage is working. Instead, provide general constructor and apply every setting externally.


code/espurna/config/types.h
```diff
@@ -22,37 +22,48 @@
...
+#define BUTTON_MODE_OFF               BUTTON_ACTION_OFF
+#define BUTTON_MODE_AP                BUTTON_ACTION_AP
+#define BUTTON_MODE_RESET             BUTTON_ACTION_RESET
+#define BUTTON_MODE_PULSE             BUTTON_ACTION_PULSE
+#define BUTTON_MODE_FACTORY           BUTTON_ACTION_FACTORY
+#define BUTTON_MODE_WPS               BUTTON_ACTION_WPS
+#define BUTTON_MODE_SMART_CONFIG      BUTTON_ACTION_SMART_CONFIG
+#define BUTTON_MODE_DIM_UP            BUTTON_ACTION_DIM_UP
+#define BUTTON_MODE_DIM_DOWN          BUTTON_ACTION_DIM_DOWN
+#define BUTTON_MODE_DISPLAY_ON        BUTTON_ACTION_DISPLAY_ON
```
mcspr, line 54: Deprecation path for these is a bit tricky. `-DBUTTON1_CLICK=BUTTON_MODE_DIM_UP` would show `error: ‘BUTTON_MODE_DIM_UP’ was not declared in this scope`, but that's it.

`#pragma GCC poison TOKEN` can restrict token usage in code, assuming poison is set even before custom header is included. One issue with that it won't work with -D... flags, `warning: poisoning existing macro "..."` is the best it could do.

davebuk:
Is it worthwhile adding a comment in the wiki that `BUTTON_MODE` is now `BUTTON_CONFIG`? I stumbled across this change the other day so I have changed all my button defines in my `custom.h` file.

mcspr:
I have changed wiki to mention CONFIG: https://github.com/xoseperez/espurna/wiki/Buttons-and-switches/_compare/d7ebe181744cc33de0b4b33e204682aa0be48639...91f9bd8eb9b8e10eea38b1eacb3c770869e6d436
Action <-> mode are missing though (unchanged part), that needs a fix there.

It should probably also mention the deprecation itself, something like "since version 1.14.2 BUTTON[1-8]_CONFIG is used instead of BUTTON[1-8]_MODE"

davebuk:
I've added it, please amend if not correct :-)

mcspr:
Thanks! Seems sorted now

PR #2170, opened 2020-03-04
LED Updates

### LINKED ISSUE #2081: Flip/Flop or slow pulse output while relay is in ON state.

mcspr:
includes & depends on #2162 , rebase will fix order
fix #2101 
fix #2081

@davebuk 3rd try. if you can test this, clone the tree directly (as always)

This should work with any state that just turns LED either ON or OFF (relay follow, findme, always on, always off).
`ledPatternN` setting, where `N` is LED ID, controls the pattern as `500,500,10 100,100,0`. Rule is read left-to-right. Like the previous PR, only difference is that patterns are now separated by ` ` (space)
LED manual mode (`ledMode => 0`) via MQTT will use the setting too.

Although, MQTT does not parse the pattern. Right now it checks for strings like `1`, `on`, `ON`, `toggle` etc. I guess this can be changed and it could try to parse the pattern when it can't parse the status string


mcspr:
> MQTT does not parse the pattern. Right

Now it does.

Changing any settings in WebUI or calling `reload` in terminal will reset the pattern to the one in settings. If there is none, it will turn off the led completely.

If there is any need in persistence, I think better option is to update `/config` endpoint to have `PATCH` method that can update existing key-value settings without erasing them first.

davebuk:
1.
I'm trying to understand what I need to `#define` to test this. I currently have two relays. The one I want to buzz is `RELAY2` which is connected to `GPIO 1`. I can set `LED1` (on board LED) to follow `RELAY2`.
```     
    // Relays
    #define RELAY1_PIN           5  //D1 Sounder relay
    #define RELAY1_TYPE          RELAY_TYPE_NORMAL
    #define RELAY2_PIN           1  //TX Buzzer relay
    #define RELAY2_TYPE          RELAY_TYPE_INVERSE

    // LED
    #define LED1_PIN             2  //D4
    #define LED1_PIN_INVERSE     1
```
What do I define to make `RELAY2` a buzzer? Do I define a second LED with the same GPIO as the relay and bind the LED to that relay?
``` 
    #define LED2_PIN          1 
    #define LED2_RELAY     2
```
Or just define the second LED2 and not have a `RELAY2`?

2.
Once configured, where do the pattern parameters get set? In the webUI? 

3.
To make the 'buzzer' output the pattern, do I send an string e.g. `500,500,10 100,100,0` using mqtt to `{root}/relay/1/set` or `{root}/led/1/set`, or do I just turn `LED2` ON?

mcspr:
> To make the 'buzzer' output the pattern, do I send an string e.g. 500,500,10 100,100,0 using mqtt to {root}/relay/1/set or {root}/led/1/set, or do I just turn LED2 ON?

The general idea is to use pattern instead of simply turning led ON.
That depends on the mode of course. `{root}/led/1/set` needs `ledMode1 => 0` aka `LED_MODE_MANUAL` (see types.h for the specific number). Relay needs `LED_MODE_FOLLOW` to work, and by default 2nd led is in manual mode.

See debug panel to `set`, `get` or `del` current settings key-values without rebuilding i.e. just enter `set ledPattern1 "500,500,0"` + `set ledMode1 2` + `reload` and it should work in relay mode. MQTT can receive message with the pattern string, which will implicitly turn the led on (as noted, temporarily, until the next settings reload event). MQTT receiving `1` should have the same effect as relay when pattern is set in settings.

davebuk:
Please excuse my ignorance, so do I get rid of `RELAY2` completely? Should I set GPIO 1 (previously configured as a relay) as LED2? I understand that an LED can be set to follow the state of a relay, but I don't understand how the LED can control a bound relay.

Once configured, I guess then in webUI I can then set `ledPattern1` and `ledMode1` depending on requirement. To activate the output I then turn ON LED2 using mqtt?

mcspr:
> Please excuse my ignorance, so do I get rid of RELAY2 completely? Should I set GPIO 1 (previously configured as a relay) as LED2? I understand that an LED can be set to follow the state of a relay, but I don't understand how the LED can control a bound relay.

LED pin should control the buzzer. Relay only controls the LED state, but if you keep RELAY2_PIN set it will turn it on briefly. You can add `DUMMY_RELAY_COUNT 1` instead of relay pin, it will create a single virtual switch that you can still use as `ledRelay1 1` / `LED2_RELAY 2`.

> Once configured, I guess then in webUI I can then set ledPattern1 and ledMode1 depending on requirement. To activate the output I then turn ON LED2 using mqtt?

Yep. When using manual mode, we need to use `led/#/set`. When using relay follow mode, we can only control the relay state through `relay/#/set`

davebuk:
Perfect. All working, great work.

I've added dummy relay and set LED1 in web to `FOLLOW ME` `switch #1`. Setting pattern as per instructions above allows `relay/1/set` to be turned ON/OFF via webUI or mqtt. This makes the buzzer beep to the pattern. Changing the numbers and setting `reload` saves the new values and changes the beep pattern.

I'll do some more testing over the next few days but basically it meets my original request perfectly. 👍

mcspr:
Thanks!
It would be also be interesting to know if I accidentally broke something with buttons, because of #2162 inclusion.
I also wonder if patterns like `500,500,10 250,250,10` should repeat from the start. For example, by prefixing the string with some flag or adding some additional param to the comma-separated list.

davebuk:
> Thanks!
> It would be also be interesting to know if I accidentally broke something with buttons, because of #2162 inclusion.

In my current configuration I do have a push button connected to `GPIO0`.     
```
    // Buttons
    // No buttons on the D1 MINI alone, but defining it without adding a button doen't create problems
    #define BUTTON1_PIN          0   // Connect a pushbutton between D3 and GND,
                                     // it's the same as using a Wemos one button shield
    #define BUTTON1_MODE         BUTTON_PUSHBUTTON | BUTTON_DEFAULT_HIGH
    #define BUTTON1_RELAY        1
```

Setting `LED#0` to follow `Switch#0` in webUI then setting `set ledPattern0 "1000,1000,0"` + `relaod` configures the on board LED. Selecting the push button ON activates the relay and the LED pulses 1sec ON/OFF and pressing the button again switches the LED and relay OFF. 

> I also wonder if patterns like `500,500,10 250,250,10` should repeat from the start. For example, by prefixing the string with some flag or adding some additional param to the comma-separated list.

Could you have a final `stop after n# times digit`? `500,500,10 250,250,10 0` would run the first 10 then the second 10 pulses until stopped by the OFF command. `500,500,10 250,250,10 1` would run the sequence once and stop by itself. `500,500,10 250,250,10 2` would run it twice and stop etc? 

For my alarm system, the arming or trigged state **could** go something like: `1000,500,7 500,500,10 300,300,11 0` so over roughly a 30 second period the pulse rate would quicken to warn you that you are running out of time to disarm before the loud sounder goes off. 

For reference `1000,500,7 500,500,10 300,300,0` works. I don't know what the limit is to sets of pulses?

mcspr:
<s>...edit...</s> I think I misread the first time. Yes, repeats would work like that. It can be anywhere in the string though, it would not mean 'restart after reading this', since we have already parsed all of the delays and merely store the condition to repeat things after pattern does not have anything more to do.

Limit for amount of sets is actually the available RAM. Main issue is that kv settings storage can only store up to 255 chars. MQTT can accept more, ~1000 at least.
Delays between ON OFF states can be up to 50000ms or around 24000ms when running 160MHz CPU setting.

PR #2172, opened 2020-03-07
Fix for long click.

ElderJoy:
Parameter length should be defined as uint16_t in _buttonMapReleased as it defined in _buttonMapEvent.

### DIFF
--- a/code/espurna/button_config.h
+++ b/code/espurna/button_config.h
@@ -134,7 +134,7 @@ constexpr const unsigned char _buttonDecodeEventAction(unsigned long actions, un
     );
 }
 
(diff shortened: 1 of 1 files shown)


ElderJoy:
With uint8_t length function _buttonMapReleased never returned BUTTON_EVENT_LNGCLICK. BUTTON_EVENT_CLICK returned instesd. I verified that with uint16_t length it returns BUTTON_EVENT_LNGCLICK when correspondent delay passed.

mcspr:
I will move debounceevent in the tree (#2162), so perhaps we should also change the callback signature from u16 to u32, as the timestamp is stored as unsigned long anyways.

ElderJoy:
If it will be compared with ul, sure it would be better if it will have the same type.

PR #2174, opened 2020-03-08
[Extend Dallas 1-Wire] Add support for read PIO-A of DS2406

rmcbc:
## Extend Dallas 1-Wire

Extend the devices supported with DALLAS 1-WIRE.
The DS2406 can be read, the system adds it as a digital device and only read PIO-A state.

## Electronics

1. Connect a 3.3Kohm resistor from pin 3.3V to IO2 (pad after 3.3V hole) on a Sonoff Basic R2.
2. Take 3 wires: GND, 3.3V and IO2 (I think is GPIO14)
3. Connect DS2406 pin 1 to GND, pin 2 to IO2 and pin 3 is the input port

**Just a Note:** if it's to receive a signal from a relay I put a 10K resistor to 3.3v as a pull up or to GND as a pull down depending if relay NC is GND or 3.3v and NO is a open signal)

## Devices

To have access to DALLAS devices I use in platformio:
`
default_envs = itead-sonoff-basic-r2-dallas
`

You can mix DS18x20 devices with this new DS2406.

Because only have 5 units of DS2406, I test all in the same Sonoff, it detects all and work without any problem.

### DIFF
--- a/code/espurna/sensors/DallasSensor.h
+++ b/code/espurna/sensors/DallasSensor.h
@@ -159,7 +159,13 @@ class DallasSensor : public BaseSensor {
 
                         uint8_t data[DS2406_STATE_BUF_LEN];
 
-                        _wire->reset();
+                        // Read scratchpad
+                        if (_wire->reset() == 0) {
+                            // Force a CRC check error
+                            _devices[index].data[0] = _devices[index].data[0] + 1;
+                            return;
+                        }
+                        
                         _wire->select(_devices[index].address);
 
                         data[0] = DS2406_CHANNEL_ACCESS;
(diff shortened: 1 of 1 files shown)


code/espurna/sensors/DallasSensor.h
```diff
@@ -120,37 +155,57 @@ class DallasSensor : public BaseSensor {
...
 
-                    _wire->select(_devices[index].address);
-                    _wire->write(DS_CMD_READ_SCRATCHPAD);
+                        uint8_t data[DS2406_STATE_BUF_LEN];
 
-                    uint8_t data[DS_DATA_SIZE];
-                    for (unsigned char i = 0; i < DS_DATA_SIZE; i++) {
-                        data[i] = _wire->read();
-                    }
+                        _wire->reset();
```
mcspr: I have noticed that the other sections does check for the 'expected' return value. Should it do the same here, and the one `reset()` below?

https://github.com/PaulStoffregen/OneWire/blob/master/OneWire.cpp#L164
> // Returns 1 if a device asserted a presence pulse, 0 otherwise.

Based on the commit history, hot-plugging is the reason for that branch:
90c51552c7358c0d0e86380f034256f865ec0a9b

rmcbc: @mcspr you were right, sorry for that.
Fixed, thanks.

PR #2182, opened 2020-03-16
Update IRremoteESP8266 to 2.7.4

mcspr:
Fix version specifier, use PIO registry instead of git
Need this for -DALLOW_DELAY_CALLS=false


mcspr:
Not tested.

Switch to the correct upstream version. Build did not break so hopefully things are fine at runtime too.

IR module has a very strange flow :/ It feels like it was written this way for pubsub API which works only in CONT context and expects us to block the whole time while code is sending (based on comment about MQTT_MAX_PACKET_SIZE:).
I wonder if sendRaw(..., count, ...) is a typo, but I don't exactly have an easy way to test it.

PR #2183, opened 2020-03-17
support for esp8266 serial, UART_TO_MQTT

orrpan:


code/espurna/config/hardware.h
```diff
@@ -4299,6 +4299,26 @@
...
+    // Info
+    #define MANUFACTURER                "ESP8266"
+    #define DEVICE                      "SERIAL"
+
+    // Serial to MQTT
+    #define UART_MQTT_SUPPORT       1
+    #define UART_MQTT_USE_SOFT      0
+    #define UART_MQTT_HW_PORT       Serial
+    #define UART_MQTT_BAUDRATE      115200
+    #define TERMINAL_SUPPORT        0
```
mcspr, line 4318: Looking at terminal.ino, it never calls Serial.read() unless debug_serial_support flag is on (which is off here).

mcspr:
You are using out-of-date branch btw, there is no need in migrate.ino and platformio.ini was changed considerably

orrpan:
True, might do a new PR instead. But it works really fine with `#define debug_serial_support 0`. Shouldn't it work? I'm getting responses on `ESPURNA-######/uartin`.
Testing it with three PWM fans connected to an arduino nano. https://github.com/orrpan/arduino_pwm_fan/blob/master/arduino_pwm_fan.ino

code/espurna/config/hardware.h
mcspr: (this comment was meant to point to the disabled TERMINAL_SUPPORT)

PR #2185, opened 2020-03-17
added support for Generic ESP01 boards with 512KiB flash

ziggurat29:
added board def and have tested with boards on-hand.  These are the first generation ESP01 boards with an 8 pin header and only 512 KiB flash.


mcspr (approved):
Sure. We still have 512kb flags and it still (somehow) works :) Even while Espressif sdk did remove any mention of it from the most recent versions. Not a concern yet though for Arduino base.

mcspr:
I was also curious about Update removal from the build, since there is no point in OTA in such conditions.
basic builds:
493120 - defaults
484928 - OTA_ARDUINOOTA_SUPPORT=0, OTA_CLIENT=OTA_CLIENT_NONE
481856 - #if0 onUpdate callbacks in web.ino + the above

ziggurat29:
> I was also curious about Update removal from the build, since there is no point in OTA in such conditions.

that makes sense to me.  feel free to change if you like, or perhaps I can try it out here since I have these boards on hand and update the PR; let me know if you want me to do that.
at the moment, I am working on gosund wp3 definition in a separate branch which will be a separate submission.

mcspr:
I have added the web part as OTA_WEB_SUPPORT, which should also cut the WebUI 'Upgrade' input from the Admin panel. esp01 hardware updated with the defines

PR #2191, opened 2020-03-20
board definition for the Gosund WP3 smart socket

ziggurat29:
this has been tested 'ex vitro' by monitoring the various GPIO levels for the peripherals when changing their states.
I'm quite confident it will work correctly when the module is re-installed into the plug, but my testing in that case is stymied by failures in OTA updates (which I will require once I solder the module back in, since the UART,RST,GPIO0 will no longer be accessible).  I am pretty sure the OTA update failures have nothing to do with this board definition, but is a more general problem.


ziggurat29:
I have confirmed my failed OTA update was merely due to insufficient supply power in my test jig, so I consider this branch to be working correctly at this point.

mcspr:
I have also updated the https://github.com/xoseperez/espurna/wiki/Hardware page

PR #2200, opened 2020-03-29
correct Gosund WP3 LED documentation and provide reasonable default actions

ziggurat29:
minor update that corrects the LED definitions (specifically, colors were reversed).
Also, updated LED default definitions to do something reasonable.  Similarly with the button.

resolve #2205

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -3248,8 +3248,8 @@
 //     GPIO2 via test points on the back of the module)
 // and these are wired to devices:
 // GPIO4: /BTN
-// GPIO12: /LED blue
-// GPIO13: /LED red
+// GPIO12: /LED red
+// GPIO13: /LED blue
 // GPIO14: RELAY
 // -----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


code/espurna/config/hardware.h
```diff
@@ -3262,17 +3262,31 @@
...
 
     // LEDs
+
+    // LED1 (red) indicates wifi activity
     #define LED1_PIN                    12
     #define LED1_PIN_INVERSE            1
+    #define LED1_MODE                   LED_MODE_WIFI
 
+    // LED2 (blue) indicates on/off state; you could use LED_MODE_FOLLOW_INVERSE
+    // so that the LED lights the button when 'off' so it can be found easily.
```
mcspr: Note of the FINDME mode, SHPX as an example.
I have not tracked color scheme all the way through all of the devices, but comparing to SPHX again - red is relay state, blue is wifi; coloring is other way around.

ziggurat29: OK! I reverse the red/ blue roles.

PR #2205, opened 2020-04-01
Add default relay for GOSUND_WP3 BUTTON1

hungnguyenm:
Following up on #2191, I wanted to define the default relay controlled by BUTTON1 similar to other devices. Without this default relay definition, pressing the button won't control the relay and rendering this switch to be remote controlled only.

Patched and tested on several of my WP3s.


ziggurat29:
lol; I did the same in #2200 (and added some stuff for the lights, though there is a suggestion I reverse the light colors)

PR #2213, opened 2020-04-06
Sensor / emon revamp (wip?)

mcspr:
@irmishappy (sorry for the delay)

Resolves ... (TODO: find all related issues)

- Update sensor definitions to support a generic way to store energy values
- Update sensor conversion code to deal with units and not magnitudes (https://github.com/xoseperez/espurna/commit/ee5384949dbae221896270a196e49d7572286c6c)
- Update sensor unit definitions to use a more generic way of defining unit IDs.
- Reset energy value based on index through external means

While this solves the energy conversion issues and we are finally seeing the real value, what I don't really like:
- KilowattHour and WattHour are separate enum tags, thus sort-of are different types altogether
- Conversion code in Energy object should probably use some generic 'ratio' calculation? (https://en.cppreference.com/w/cpp/numeric/ratio/ratio)
- We are still using runtime checks to do calculations and depend that sensor outputs only one specific value type.

I have found some alternative systems:
https://github.com/nholthaus/units/
Which implement unit as strongly typed class that can be used to enforce our runtime checks at compile time. Non-trivial part is that I am not really sure what structure the code must use to implement strongly typed getters. We can do something like:
```cpp
class BaseSensor {
    virtual units::energy::KilowattHour getEnergy(unsigned char index) { return 0; } 
    virtual units::temperature::Celcius getTemperature(unsigned char index) { return 0; } 
};
```
And we still need to check specific run-time property of the sensor to know what getter outputs correct data. We would get rid of conversion stuff though, it would magically do it all by itself.
(Note that I am not sure about the underlying type / value stuff and whether we can hit some limits there, units.h source is pretty cryptic :/)


mcspr:
Also bookmarking something from https://github.com/discover:
https://github.com/LLNL/units#alternatives

At least one of the options may be reasoned with a bit more easily:
https://github.com/mpusz/units
https://mpusz.github.io/units/framework/basic_concepts.html
What will not be really obvious is how to port c++20 / c++23 code to ours c++11 (or, potentially, c++14)

PR #2227, opened 2020-04-22
Fixes in ADC_VCC and ADC_TOUT.  Adding support of HDC1080 sensor.

vtochq:
Defining macro with different names ADC_VCC_VALUE and ADC_TOUT_VALUE for compatibility with macro ADC_MODE_VALUE.

Adding support for HDC1080 and 831R sensor.


code/espurna/config/sensors.h
```diff
@@ -1255,9 +1268,13 @@
 // ADC
 // -----------------------------------------------------------------------------
 
+// In esp.h ADC_VCC and ADC_TOUT is not macro, so define macro.
+#define ADC_VCC_VALUE                   255
+#define ADC_TOUT_VALUE                  33
```
mcspr: Why hard-code it here to proxy ADC_VCC & ADC_TOUT? enum value == int when used in C-style definition.
We can keep using Esp.h enum directly

code/espurna/ws.ino
```diff
@@ -358,9 +358,10 @@ void _wsUpdate(JsonObject& root) {
     root["uptime"] = getUptime();
     root["rssi"] = WiFi.RSSI();
     root["loadaverage"] = systemLoadAverage();
-    #if ADC_MODE_VALUE == ADC_VCC
+    if (ADC_MODE_VALUE == ADC_VCC)
         root["vcc"] = ESP.getVcc();
-    #endif
+    else
+        root["vcc"] = "N/A (TOUT) ";
```
mcspr: Please add brackets

mcspr (changes requested):
`#define HDC1080_SUPPORT 1` needs to be in test/build/sensor.h to catch any issues with build process
`code/espurna/config/arduino.h` needs updated comment

I am also assuming you had noticed humidity bug, will fix that asap :)

code/espurna/config/sensors.h
vtochq: But I think we can't define macro with value from enum.
We need this values on pre-compiler level cause it used in #if directives for define ADC_MODE_VALUE.
We also can try to #define ADC VCC and ADCT OUT but I'm afraid it's broke compatibility with esp.h.

vtochq:
> I am also assuming you had noticed humidity bug, will fix that asap :)

Sorry, I don't noticed humidity bug. What do you mean?

code/espurna/config/sensors.h
mcspr: We can define anything as token, it is just text.

I am not seeing any preprocessor checks with ADC_MODE_VALUE, ADC_VCC_VALUE, ADC_TOUT_VALUE. So what I mean is:
```cpp
// sensor.h
#ifndef ADC_MODE_VALUE
#define ADC_MODE_VALUE ADC_VCC
#endif
```
(and slightly below when we force MODE_VALUE to become TOUT with analog sensors
```cpp
// sensor.h
#undef ADC_MODE_VALUE
#define ADC_MODE_VALUE ADC_TOUT
```

vtochq: If we define ADC_MODE_VALUE as text "ADC_VCC" we can't compare it's value with ADC_VCC value defined in esp.h. We often use ADC_VCC in utils.ino, for example.

mcspr: We can't compare things in preprocessor, true, but the there is no issue with:
```cpp
if (ADC_MODE_VALUE == ADC_VCC) {
```
gcc expands MODE_VALUE to ADC_VCC essentially making this expression either `if (true) {` or `if (false) {`

vtochq: I agree, but SDK use it in utils.ino

// This method is called by the SDK to know where to connect the ADC
int __get_adc_mode() {
    return (int) (ADC_MODE_VALUE);
}

mcspr: But the function is called at runtime, preprocessor puts correct enum value here which is converted back to int.

Building analog sensor binary, objdump -d firmware.elf
```asm
402422cc <_Z14__get_adc_modev>:
402422cc:       122c                    movi.n  a2, 33 <<< here is 33 aka ADC_TOUT
402422ce:       f00d                    ret.n
```

vtochq: You right, sorry.
We need just rewrite #if to if in all ADC_ expressions.

mcspr (approved):

mcspr:
I noticed the sensor value constrain 0...100, we don't have it properly doing it in the dev. It does apply constrain on pressure instead of humidity though, which is wrong.

PR #2233, opened 2020-04-29
Publish button actions to broker

zafrirron:
In some use cases we want to register on button press event to trigger some actions

### DIFF
--- a/code/espurna/button.ino
+++ b/code/espurna/button.ino
@@ -691,6 +691,9 @@ void _buttonLoopGeneric() {
         auto event = _buttons[id].loop();
         if (event != button_event_t::None) {
             buttonEvent(id, event);
+            #if BROKER_SUPPORT
+                StatusBroker::Publish(MQTT_TOPIC_BUTTON, id, _buttonEventNumber(event));
+            #endif
         }
     }
 }


mcspr:
tbh I'd prefer to add a separate broker for buttons specifically, so we don't have to hijack 'status' integer to be 3 separate things depending on topic

zafrirron:
Ok,
I'm not so sure about the new broker model of separating brokers to multiple brokers types, a more conventional publish subscribe model with a topic registration, could make the registration, publishing easier and clean code better.

mcspr:
Do you mean something like the previous model? Or that we don't always have any use for the 'topic' itself?

Types are just a means to have an array of things shared between modules and have different arguments instead of string serialization<->deserialization inside of every broker callback.
https://github.com/xoseperez/espurna/blob/546c8f09b2a8754038fcd57e0176fa0198011432/code/espurna/broker.ino#L21
At least we don't have to match strings with some internal, when we are already running the code that knows the exact types of things:
https://github.com/xoseperez/espurna/blob/546c8f09b2a8754038fcd57e0176fa0198011432/code/espurna/relay.ino#L187
https://github.com/xoseperez/espurna/blob/546c8f09b2a8754038fcd57e0176fa0198011432/code/espurna/sensor.ino#L1213
https://github.com/xoseperez/espurna/blob/546c8f09b2a8754038fcd57e0176fa0198011432/code/espurna/light.ino#L643-L644
Although, we still not really following this rule with status broker, by not sending Status enumeration and percentages separately. But we still need topic (as string or otherwise) to separate different modules.

zafrirron:
I'm not sure I understand the need for different broker types (may have missed your explanation).
As I look at the broker its a simple internal message que for espurna,
I want to be able to subscribe to a topic (button, relay...) from any module (plugin engine in my use case) and to be notified, by callback when a messages is published on the topic.
having the ability to subscribe on a topic (requires broker engine implementation of a loop on each topic callbacks list) , eliminates the need of the callback function to test topic value if a module subscribes to generic broker type or handle different brokers.  
so:
1. module x subscribes (register): BrokerRegister(topic,callback)
2. module y publishes : BrokerPublish(topic,args....)
3. broker engine publish implementation:  BrokerPublish(topic,args....):  loop on callbacks of the topic 

I do agree that the downside is serialization/deserialization of the callback args, this however the responsibility of the callback function and as I see, mosts of the use cases could handle two strings (id,value) as args for the topic, in rare cases the value could be serialized.

mcspr:
I am not defending types, on the contrary, just happened to be done that way :) I was referring on how the existing model changes call sites. My only goal, improving over previous version, was to allow us to pass arbitrary things directly. See NTP or key-value brokers, for example, both do not fit the generic id-value model.

- We can have a base behaviour-defining broker class that extends publish and subscribe? That would also solve the broker.h describing every possible way we can call it, each module will define broker individually.
- Topic string matching on the upper level can be a nice improvement.

mcspr:
iydm I'll add the suggested broker code, keeping the existing syntax for now.

```cpp
// base for our class, with slightly nicer user-facing syntax
template <typename Base, typename TFunc>
struct TBroker : public TBrokerBase {};

// CRTP-stuff, see below
template <typename Base, typename R, typename ...TArgs>
struct TBroker<Base, R(TArgs...)> : public TBrokerBase {
    using TCallback = std::function<R(TArgs...)>;
    using TCallbacks = std::vector<TCallback>;

    static void Register(TCallback callback) {
        callbacks.push_back(callback);
    }

    static void Publish(TArgs... args) {
        for (auto& callback : callbacks) {
            callback(args...);
        }
    }
    protected:
        static TCallbacks callbacks;
};

// we no longer need to bind things as TArgs, just pass function signature like with std::function<...>
using StatusBrokerFunc = void(const String&, unsigned char, unsigned int);

// because we are stuck with static allocations (hopefully, for now), c++ allows us to inherit from 'ourselves' and put static variable into a unique class
// ref: https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern
template <typename T>
using TStatusBroker = TBroker<T, StatusBrokerFunc>;
struct StatusBroker : public TStatusBroker<StatusBroker> {};

...

// Later, in the 'status' module, since we are required to instantiate static variable
template<>
TStatusBroker<StatusBroker>::TCallbacks TStatusBroker<StatusBroker>::callbacks {};

```

*edit:* So, for buttons the func will be:
```cpp
using ButtonFunc = void(unsigned char id, button_event_t event);
```

mcspr:
Closing, #2253

Issue #2219, opened 2020-04-10
Power plug specific settings not respected


reaper7:
Today I updated my HYKKER_SMART_HOME_POWER_PLUG to latest dev 
but some defined for this device settings like SENSOR_ENERGY_UNITS and probably RATIOS are not respected.
So...device show more than ~330V instead ~236 and Energy units as J (instead defined kWh)

When I back to dev commit from Apr 2, 2020 then all works fine again.

is the reason for this? :
https://github.com/xoseperez/espurna/commit/cae50fa5445afac661e89eed86b5e2122da8a9d7
and will be fixed soon? 
```
Further sensor refactoring... soon.
```

mcspr:
> RATIOS

That is something I missed bc it was in the configuration part, but you should be able to use <s>runtime overrides in WebUI</s> settings overrides? Or see following key values <- flags association:
`pwrRatioC` <- `HLW8012_CURRENT_RATIO`
`pwrRatioV` <- `HLW8012_VOLTAGE_RATIO`
`pwrRatioP` <- `HLW8012_POWER_RATIO`

> SENSOR_ENERGY_UNITS

Is reset to default in settings when previous value is used, but should work when set as a flag? See current `eneUnits` setting value. I will probably revert that part though :/ Currently `dev` version tracks global enumeration of 'units', and previously it used magnitude-specific index, so numbers are different. But, it should probably keep using local indexes to allow us to change things in the future

reaper7:
Both can be solved by setting the correct values in the web interface
but I often do some experiments on devices and equally often I have to do Factory Reset via webinterface...
..so, it would be good to have fully working default settings for specific devices from hardware.h :)

I will temporarily use dev from Apr 2, 2020

---

<b>OT</b>. I don't want to open another issue but
I noticed a problem with the time display.
I have defined server and time zone in custom.h
```
  #define NTP_SUPPORT                 1
  #define NTP_SERVER                  "homegateway.lan"
  #define NTP_TIMEZONE                TZ_Europe_Warsaw
```
the time refresh process is running correctly, 
timezone is also correctly decoded because I see in webinterface:
```
CET-1CEST,M3.5.0,M10.5.0/3
```
but on webinterface I still see time as UTC (18h instead 20h)

mcspr:
That was meant to point out to a missing code! Not dismissing the issue :) I'll bundle some other changes in the mean time regarding other configuration options. This will be included:
```patch
[...]

PR #2240, opened 2020-05-01
Hlw8012 early config

mcspr:
- fix ratios getting lost when not using generic hlw resistor values. move initialization to sensor itself (@reaper7)
- emon methods now follow default setters when not overriden
(todo: mark levels above as override would be a nice touch too, to actually hint about what is happening)
as noted in the #2219, we can't call such indexed methods in child directly when it implements one of the method types and must use baseemon cast to resolve them properly.

### DIFF
--- a/code/espurna/sensor.cpp
+++ b/code/espurna/sensor.cpp
@@ -1491,9 +1491,6 @@ void _sensorLoad() {
         sensor->setSEL(getSetting("snsHlw8012SelGPIO", HLW8012_SEL_PIN));
         sensor->setCF(getSetting("snsHlw8012CfGPIO", HLW8012_CF_PIN));
         sensor->setCF1(getSetting("snsHlw8012Cf1GPIO", HLW8012_CF1_PIN));
-        sensor->setCurrentRatio(HLW8012_CURRENT_RATIO);
-        sensor->setVoltageRatio(HLW8012_VOLTAGE_RATIO);
-        sensor->setPowerRatio(HLW8012_POWER_RATIO);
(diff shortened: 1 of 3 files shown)


reaper7:
After upgrade device with this commit and factory reset, hlw8012 shows good value, 
so it seems that default ratios are set correctly.

Now I have two additional questions:
- should these values not be shown by terminal? 
get pwrRatioC (and rest ratios) still show null value

- **web interface -> sensors -> Reset calibration**
restores what factory defaults?
Certainly not the ones that are default for the selected device
because after Reset calibration I get strange readings again... e.g 339V instead real 234V

Only after **web interface -> admin -> Factory Reset**
again receives correct results.

Unfortunately I can't verify which calibration values are currently being used in each of these two cases because, as I wrote above, terminal commands like **get pwrRatioXX** return null values

mcspr:
> web interface -> sensors -> Reset calibration

It deletes pwrRatioX keys and calls calculateDefaultRatios, again, restoring resistor values. Sigh. See resetRatios() in HLW8012Sensor header, we also need that call changed with the addition of default ratios part.

> should these values not be shown by terminal?
get pwrRatioC (and rest ratios) still show null value

`get` just uses kv storage, it does not interact with the sensor code in any way. although I might add a query doing just that if value is null. Effectively, hw.h just puts the ratio value in the sensor class code, we don't actually set pwrRatio 'thing'

reaper7:
> I might add a query doing just that if value is null

that would be helpful in analyzing errors/problems and adding new devices with this (and others) sensor

PR #2241, opened 2020-05-03
Load ratios after boot + show pwr defaults with `get`

mcspr:
@reaper7 how about this then. (3rd time 's the charm?) This got lost from the previous per-device config.

https://github.com/xoseperez/espurna/commit/c266163b48eb6c70af759de55c0bda3fdaf5e8c6 Now we remember ratios that are generated from resistor values + ours ratio flags and use them as defaults when calling resetRatios(). Also, use these as settings defaults, so we don't end up never restoring original values.

https://github.com/xoseperez/espurna/commit/eab86c2bdce5f3b933df854941ca3fc1bb5c10f2 Small tweak to the `get` command, we can retrieve defaults from the sensor itself. This is a bit tricky though, so we end up using model similar to the WebUI key processor and try to match some known prefixes.

In addition:
- settings::internal::serialize draft, since I initially tried to hook things into the getSetting itself, but that did not quite work out :) This is still useful for the WebUI, we need to serialize things anyway. Not eveything is converted yet.
- DEBUG_MSG now can use %f formatting (not with PIO, yet, I need to unflag the float_printf)

### DIFF
--- a/code/espurna/sensors/BaseSensor.h
+++ b/code/espurna/sensors/BaseSensor.h
@@ -39,11 +39,11 @@ class BaseSensor {
         // Descriptive name of the sensor
         virtual String description() = 0;
 
+        // Descriptive name of the slot # index
+        virtual String description(unsigned char index) = 0;
+
         // Address of the sensor (it could be the GPIO or I2C address)
         virtual String address(unsigned char index) = 0;
-
-        // Descriptive name of the slot # index
-        virtual String slot(unsigned char index) = 0;
 
         // Type of sensor
         virtual unsigned char type() { return sensor::type::Base; }
@@ -83,6 +83,9 @@ class BaseSensor {
 
         // Number of available slots
         unsigned char count() { return _count; }
+
+        // Convert slot # index to a magnitude # index
+        virtual unsigned char local(unsigned char slot) { return 0; }
 
         // Hook for event callback
         void onEvent(TSensorCallback fn) { _callback = fn; };
--- a/code/espurna/alexa.cpp
+++ b/code/espurna/alexa.cpp
@@ -82,7 +82,7 @@ void _alexaBrokerCallback(const String& topic, unsigned char id, unsigned int va
 // -----------------------------------------------------------------------------
 
 bool alexaEnabled() {
-    return getSetting<bool>("alexaEnabled", 1 == ALEXA_ENABLED);
+    return getSetting("alexaEnabled", 1 == ALEXA_ENABLED);
 }
 
 void alexaLoop() {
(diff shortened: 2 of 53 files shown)


mcspr:
(TODO: other sensors need to have correct defaults)

reaper7:
Nice!
```
[090569] [WEBSOCKET] Requested action: dbgcmd
[090571] > pwrRatioC => 25740.00 (default)
[090571] +OK
[098444] [WEBSOCKET] Requested action: dbgcmd
[098448] > pwrRatioV => 282060.00 (default)
[098448] +OK
[104859] [WEBSOCKET] Requested action: dbgcmd
[104866] > pwrRatioP => 3414290.00 (default)
[104866] +OK
```
after Reset calibration or/and after Factory reset, get pwrRatioX shows good default values from hw.h,

after changing by for e.g. Expected Voltage, get pwrRatioV show new value without "(default)" word

PERFECT!

P.S. testing is a bit difficult now, they broke something with mDNS on esp8266 dev branch
and all my devices disappear after a while from OTA :(

code/espurna/sensors/BaseSensor.h
```diff
@@ -39,12 +39,12 @@ class BaseSensor {
...
 
+        // Descriptive name of the slot # index
+        virtual String description(unsigned char index) = 0;
+
         // Address of the sensor (it could be the GPIO or I2C address)
         virtual String address(unsigned char index) = 0;
 
-        // Descriptive name of the slot # index
-        virtual String slot(unsigned char index) = 0;
-
```
mcspr, line 47: - we will name-clash with `sensor_magnitude_t::slot` otherwise
- slot aka index is already used as 'description' in the webui part
- I would've added some active deprecation notice if c++ part was more flexible. virtual method can be `final`, but we can't add custom message. attribute(( would not work as expected. static_assert is always active.
searching for solution also points to adding documentation instead of issuing compiler warnings...

code/espurna/sensors/BaseSensor.h
```diff
@@ -84,6 +84,9 @@ class BaseSensor {
         // Number of available slots
         unsigned char count() { return _count; }
 
+        // Convert slot # index to a magnitude # index
+        virtual unsigned char local(unsigned char slot) { return 0; }
```
mcspr, line 88: naming might be better.
`counts(unsigned char slot)?` (or just `count`), since we mostly repeat what `sensor_magnitude_t::counts()` does. this will accompany already existing `count()`

reaper7: about names/nomenclature, it's hard for me to relate, you overwhelm the whole project, you know better :D

I uploaded your recent changes:
https://github.com/mcspr/espurna/commit/d8196c3a89924dca846c7dc8d9287e7526390b25
works well.

mcspr: thanks : ) just recording some reasoning / warning

PR #2243, opened 2020-05-05
hardware: Add support for HUGOAI smart socket plug.

estebanz01:
Following the same pattern as my previous PR, I added support to the HUGOAI AWP02LN smart sockets, which have one button and one led, which in my case is blue.


code/platformio.ini
```diff
@@ -1006,3 +1006,7 @@ src_build_flags = -DKINGART_CURTAIN_SWITCH
 [env:aoycocr-x5p]
 extends = env:esp8266-1m-base
 src_build_flags = -DAOYCOCR_X5P
+
+[env:hugoai-awp02l-n]
+extends = env:esp8266-1m-base
+src_build_flags = -HUGOAI_AWP02L_N
```
pezinek: shouldn't this be prefixed with -D ?
```suggestion
src_build_flags = -DHUGOAI_AWP02L_N
```

mcspr:
Thanks!

Issue #2015, opened 2019-11-26
Review Embedis storage library usage


mcspr:
Something to explore:
- [are channels](https://github.com/thingSoC/embedis/blob/430cb3ae1a92b7e0f2ca14e77269f17e16098061/src/Embedis.cpp#L509-L526) useful at all (logging toggle?)
- could some of the terminal methods benefit from ["hardware" abstraction](https://github.com/thingSoC/embedis/blob/430cb3ae1a92b7e0f2ca14e77269f17e16098061/src/Embedis.cpp#L745-L754). could hardware abstration be used for some new functionality
- is it viable to use multiple dictionaries (e.g. separate internal and user settings, simplify backup)
- simplify backup mechanism by manually using kvs_work (that needs a small change from private: to protected: in the Embedis class)
- would something like [nvs_flash](https://github.com/espressif/ESP8266_RTOS_SDK/tree/master/components/nvs_flash) be more appropriate for settings storage. It will also sort-of remove the need for EEPROM rotation and support typed storage.

---
Channels and hardware take around 1.5KB of space regardless of their use, but are easily stripped via ifndefs. I'd rather have those as app code locally.

We also do have some assumptions about Embedis structures (setttingsSize()), which perhaps should be reworked to also use kvs_work and implicit kvs_reset, so that the real size it consistent with how far Embedis could read the data

---
SPIFFS / LittleFS could be useful alternative, but that would imply settings being in one place for easy serialization (for example, see https://github.com/letscontrolit/ESPEasy/blob/68cae5519fe4043d109d99170fc0cd85c2b071fa/src/ESPEasyStorage.ino#L232-L300)
Partial writes would be really inefficient due to the FS thing and non-trivial metadata / size restrictions. The most sound approach is to keep Embedis data structure but replace EEPROM dict with some generic array storage and periodic sync with the FS. I'd look at LittleFS first and only implement it for the latest Cores.

PR #2247, opened 2020-05-12
Terminal: change command-line parser

### LINKED ISSUE #2245: Newline handling can cause Embedis::process() to get stuck
**Bug description**

When using `nc -t $IP 23` we can sometimes cause Embedis to stop processing any new commands.

Adding `-` at the beginning of the command will trigger the issue. Some tests were added upstream https://github.com/thingSoC/embedis/commit/31629f12237dc8d90ff02a8d4e6109a71388702c that use hyphen before command, and it *seems* that those expect `\r\n` or just `\r` (based on the comments). Running test suite locally shows the problem, e.g.:
```diff
diff --git a/tests/protocol.cpp b/tests/protocol.cpp
index b66543a..3b29d80 100644
--- a/tests/protocol.cpp
+++ b/tests/protocol.cpp
@@ -143,6 +143,8 @@ TEST(Protocol, Binary)

     a = {"rom", "ram"};

+    EXPECT_EMBEDIS_ARRAY("-DICTIONARIES\n", a);
+    EXPECT_EMBEDIS_ARRAY("-DICTIONARIES\r\n", a);
     EXPECT_EMBEDIS_ARRAY("DICTIONARIES", a);
     EXPECT_EMBEDIS_OK("*2\r\n$6\r\nSELECT\r\n$3\r\nRAM\r\n");
     EXPECT_EMBEDIS_ARRAY("DICTIONARIES", a);
```
Note that it looks like Embedis is expected to parse the stream even without any newline char(s). What I am not sure about is why `-` makes such a difference.

**Steps to reproduce**

Embedis processing gets stuck on specific inputs, but the most common test is to add `-` at the start of the command, switching it (intentionally, or otherwise) into binary mode. MVCE:
- `nc -t $ESPURNA_IP 23`
- type `-heap`, which translates to `-heap\n` and now we are expecting to see heap info
- nothing happens... until we restart the connection and add `-C` to the args list, using crlf for line breaks. previous buffer will finally be discarded and the next command will work.

**Expected behavior**

`heap` outputs data, we don't get stuck :)

**Possible solutions**

- we can implicitly inject the correct newline char into the stream when using telnet. Not sure, but I'd expect this broken with serial too, for example we can use `--eol LF` as `pio device monitor` argument.
- fixing parser is another option, but I need a bit more time to understand the reasoning behind the code that it uses right now. we don't use a lot of features of the embedis library overall, binary mode is a bit of a mystery.
- using some basic 'library' that will handle argv creation and dispatching commands instead of Embedis
- rolling our own parser is yet another option, where I'd probably just try to use something generated instead of hand-written. 

**Screenshots**
**Device information**
* *ESPurna version* All versions, no changes to Embedis for a long time.
* *Flash mode* \-
* *Device brand, model and version* \-

**Tools used**
* *Desktop operating system* Win10. `nc` via WSL terminal, `pio device monitor` via cmd.exe
* *Browser & version* \-
* *IDE & version* \-
* *Compiler & version (if not embedded in IDE)* \-

**Additional context**
\-

mcspr:
- switch to a custom parser, modified sdssplitargs from redis / sds. resolve #2245 
- update terminalRegisterCommand signature, pass only bare minimum
- never create Embedis terminal instance and thus never set up it's default commands (ref #2015 )
- port Embedis datastore `del` and `set` commands.
- clean-up `help` & `commands`
- allow our custom test suite to run command-line tests

This is direct port of the functionality, nothing really new besides CommandContext. But, now we only depend on Embedis to do settings. Which I will also replace in the next PR, since we need more access to the inner workings.

TODO:
- check how std::function<...> for CommandFunc changes things
- text responses for datastore can be improved
- see about wiki changes


mcspr:
> check how std::function<...> for CommandFunc changes things

For the worse. +3.5KB of data due to 'reasons'. This can be alleviated by using vector instead of map.

> see about wiki changes

Also a note about ctx.output. Generic implementation uses debugSend, but now we can also do http callback <s>like this:</s> *edit:* For example, full implementation of the callback:

<details>

<summary>Old example code (PR was updated with a more *correct* implementation)</summary>

```cpp
webRequestRegister([](AsyncWebServerRequest* request) {
    const auto url = request->url();
    if (!url.equals("/cmd")) return false;

    webLog(request);

    auto* cmd_param = request->getParam("line");
    if (!cmd_param) return false;

    auto cmd = cmd_param->value();
    if (!cmd.endsWith("\r\n") && !cmd.endsWith("\n")) {
    cmd += '\n';
    }

    auto* response = request->beginResponseStream("text/plain");
    response->addHeader("Connection", "close");

    StreamAdapter<const char*> stream(*response, cmd.c_str(), cmd.c_str() + cmd.length() + 1);
    terminal::Terminal handler(stream);

    if (handler.processLine() != terminal::Terminal::Result::Command) {
        response->setCode(500);
    }

    request->send(response);
    return true;
});
```
Where StreamAdapter class is:
```cpp
// TODO: maybe we can somehow generalize Print part into a separate class
template <typename T>
struct StreamAdapter : public Stream {
    StreamAdapter(Print& writer, T&& begin, T&& end) :
        _writer(writer),
        _current(std::forward<T>(begin)),
        _begin(std::forward<T>(begin)),
        _end(std::forward<T>(end))
    {}

    int available() override {
        return (_end - _current);
    }

    int peek() override {
        if (available() && (_end != (1 + _current))) {
            return *(1 + _current);
        }
        return -1;
    }

    int read() override {
        if (_end != _current) {
            return *(_current++);
        }
        return -1;
    }

    void flush() override {
        _writer.flush();
    }

    size_t write(const uint8_t* buffer, size_t size) override {
        return _writer.write(buffer, size);
    }

    size_t write(uint8_t ch) override {
        return _writer.write(ch);
    }

    protected:

    Print& _writer;

    T _current;
    T const _begin;
    T const _end;
};
```
(ofc, we also need to make sure we have linebreak)
*edit:* We do end up allocating twice, but imo it should be worth it if http / mqtt callbacks are added after this

</details>

mcspr:
`eeprom.dump` works, yey
obviously, bandwidth dependent, and that we receive '\n' ('println' or 'printf("...\n")')
it is a very hack'y way to do things though

mcspr:
Since this PR ended up researching Stream / Print integration, I will mention
WriteBufferingStream from https://github.com/bblanchon/ArduinoStreamUtils/
(https://arduinojson.org/v6/api/json/serializejson/#performance)

Which might be a good idea to add to have similar behaviour with Core's WiFiClient. In theory, using buffering stream + pubsubclient would allow us to have mqtt terminal topic (up to MQTT 3.1.1 payload limits, see standard docs)

mcspr:
Also, showcase of the patch replacing unordered_map with a plain `vector<pair<String, CommandFunc>>`, saving 2368 bytes (567504 vs. 565136, observed with git env and test/build/basic.h)
```diff
diff --git a/code/espurna/terminal_commands.cpp b/code/espurna/terminal_commands.cpp
index c6fbfe4d..9f7a61d2 100644
--- a/code/espurna/terminal_commands.cpp
+++ b/code/espurna/terminal_commands.cpp
@@ -17,13 +17,11 @@ Heavily inspired by the Embedis design:

 namespace terminal {

-std::unordered_map<String, Terminal::CommandFunc,
-    parsing::LowercaseFnv1Hash<String>,
-    parsing::LowercaseEquals<String>> Terminal::commands;
+std::vector<Terminal::CommandPair> Terminal::commands;

 void Terminal::addCommand(const String& name, CommandFunc func) {
        if (!func) return;
-       commands.emplace(std::make_pair(name, func));
+       commands.emplace_back(std::make_pair(name, func));
 }

 size_t Terminal::commandsSize() {
@@ -61,7 +59,9 @@ Terminal::Result Terminal::processLine() {
                        auto cmdline = parsing::parse_commandline(buffer.data());
             buffer.clear();
             if (cmdline.argc >= 1) {
-                 auto command = commands.find(cmdline.argv[0]);
+                auto command = std::find_if(commands.begin(), commands.end(), [&cmdline](const CommandPair& pair) {
+                    return pair.first.equalsIgnoreCase(cmdline.argv[0]);
+                });
                 if (command == commands.end()) return Result::CommandNotFound;
                 (*command).second(CommandContext{std::move(cmdline.argv), cmdline.argc, stream});
                 return Result::Command;
diff --git a/code/espurna/terminal_commands.h b/code/espurna/terminal_commands.h
index 928a68a9..01159a69 100644
--- a/code/espurna/terminal_commands.h
+++ b/code/espurna/terminal_commands.h
@@ -13,6 +13,7 @@ Copyright (C) 2020 by Maxim Prokhorov <prokhorov dot max at outlook dot com>
 #include "terminal_parsing.h"

 #include <unordered_map>
+#include <utility>
 #include <functional>
 #include <vector>

@@ -40,6 +41,7 @@ struct Terminal {
     };

     using CommandFunc = void(*)(const CommandContext&);
+    using CommandPair = std::pair<String, CommandFunc>;
     using ProcessFunc = bool(*)(Result);

     // stream      - see `stream` description below
@@ -81,9 +83,7 @@ struct Terminal {
     // TODO: every command is shared, instance should probably also have an
     //       option to add 'private' commands list?
     // Note: we can save ~2.5KB by using std::vector<std::pair<String, CommandFunc>>
-    static std::unordered_map<String, CommandFunc,
-        parsing::LowercaseFnv1Hash<String>,
-        parsing::LowercaseEquals<String>> commands;
+    static std::vector<CommandPair> commands;

 };
 }
~
```

PR #2250, opened 2020-05-13
Kingart curtain switch UI support

echauvet:
This is a preliminary commit to extend KA work by @AlbertWeterings. Using AT+START commands, we know the real position of the curtain. Small MCU bugs are bypassed as commented in the code.
The UI is animated  and displays :
- A basic animation of the curtain opening (4 types available) - this reflects the real position of the curtain if the switch is correctly calibrated ( the MCU adds 2 seconds to the initial calibration)
- The 3 buttons are displayed and animated (blue/red colors) as the curtain switch. (this is achieved with AT+START commands and move status detection)
- A slider to set an absolute position. Its position is refreshed whenever the curtain stops moving.

image

A configuration panel as been created to choose:
- The curtain type. I will use it later for mqtt to integrate a basic UI layout of the curtain
image
- Once rebooted/powered on, the curtain initial position could be set
image

Next steps are code cleaning, integrate any reviews from contributors and improve the MQTT stack with  RedNode. For now, I can drive my curtain as a Google curtain, or a switch or a lamp bulb (for brightness setting). I'm preparing some demo flow with Google legacy curtain, and the hybrid solution.

Both core base and latest-base are working - with latest-base, a two steps upgrade is required.
I'm cheking the work done for OTA in RedNode to check if some code is ready for 2 steps upgrade.

All Tests has been done using a curtain switch, espurna UI, google home and RedNode UI and all seems to animate together without bugs.


code/html/custom.js
```diff
@@ -1691,6 +1734,49 @@ function processData(data) {
...
+                case '2': //One side right to left
+                    $("#curtainGetPicture").css('background', 'linear-gradient(270deg, black ' + value.curtainGet + '%, #a0d6ff ' + value.curtainGet + '%)');
+                break;
+                case '3': //Two sides
+                    $("#curtainGetPicture").css('background', 'linear-gradient(90deg, black ' + value.curtainGet/2 + '%, #a0d6ff ' + value.curtainGet/2 + '% ' + (100 - value.curtainGet/2) + '%, black ' + (100 - value.curtainGet/2) + '%)');
+                break;
+            }
+            $("#curtainSet").val(value.curtainSet); //Update sliders
+            if(value.curtainMoving == "Idle") { //When idle, all buttons are off (blue)
+                $("button.curtain-button").css('background', 'rgb(66, 184, 221)');
```
mcspr, line 1791: Consider adding a vars for both rgb(...) as value explanation instead of comment?

Similar for background, inline func definition with an arg, perhaps?
```js
var gradient = function(degrees, pair1, pair2) { $("#curtainGetPicture").css('background', ...); };
// for case '3', just check if `pair2 !== undefined`
```

code/espurna/curtain_kingart.cpp
```diff
@@ -24,21 +29,157 @@ Copyright (C) 2019 by Albert Weterings
...
+int _curtain_position_set           = CURTAIN_POSITION_UNKNOWN; //Last position asked to be set (not the real position, the real query - updated when the curtain stops moving)
+bool _curtain_waiting_ack           = false; //Avoid too fast MQTT commands
+bool _curtain_ignore_next_position  = false; //Avoid a bug (see (*1)
+bool _curtain_initial_position_set  = false; //Flag to detect if we manage to set the curtain back to its position before power off or reboot
+
+// Calculated behaviour depending on KA switch, MQTT and UI actions
+bool _curtain_moving                = true;
+
+//Enable more traces, true as a default and stopped when curtain is setup.
+bool _debug_flag = true;
```
mcspr: `_curtain` prefix?

code/espurna/curtain_kingart.cpp
```diff
@@ -24,21 +29,157 @@ Copyright (C) 2019 by Albert Weterings
...
 }
-  
-void _KACurtainReceiveUART() {
+
+//------------------------------------------------------------------------------
+//Send a formatted message to MCU
+void _KACurtainSet(int button, int position = CURTAIN_POSITION_UNKNOWN) {
+
+   if(_curtain_waiting_ack) {
+        DEBUG_MSG_P(PSTR("[KAUART] MCU BUSY : Request ignored!\n"));
```
mcspr: I noticed that we not always respect debug_flag.
Maybe add a file-level function / macro at the top that wraps DEBUG_MSG_P with debug_flag? e.g.
```cpp
#define KINGART_DEBUG_MSG_P(...) do { if (_curtain_debug_flag) { DEBUG_MSG_P(__VA_ARGS__); } } while(0)
```

code/espurna/curtain_kingart.cpp
```diff
@@ -87,108 +235,248 @@ To configure the device:
...
+                _KAStopMoving();
+       }
+    } else //Not paused, not moving, and we received an AT+UPDATE -> This means that we are moving
+        _KASetMoving();
 
-        _KACurtainNewData = false; 
+    //Update last position and transmit to MQTT (GUI is at the end)
+    if(_curtain_position != CURTAIN_POSITION_UNKNOWN && _curtain_last_position != _curtain_position) {
+        _curtain_last_position = _curtain_position;
+        if(_curtain_initial_position_set) //Update initial position - TDOD : maybe only when move is finished to avoid to write too frequently
```
mcspr: re TODO, you can always use rtcmem to store intermediate value and commit when it settles
see lights for example, where ticker is reset each time the value changes. setsetting only happens when values don't change for a while

code/espurna/curtain_kingart.cpp
```diff
@@ -87,108 +235,248 @@ To configure the device:
...
+//------------------------------------------------------------------------------
+bool _curtainWebSocketOnKeyCheck(const char * key, JsonVariant& value) {
+    if (strncmp(key, "curtain", strlen("curtain")) == 0) return true;
+    return false;
+}
+
+//------------------------------------------------------------------------------
+void _curtainWebSocketUpdate(JsonObject& root) {
+    JsonObject& state = root.createNestedObject("curtainState");
+    state["curtainGet"] = _curtain_last_position;
```
mcspr: No need to prefix with `curtain` when we already know we are in `curtainState`?

code/espurna/curtain_kingart.cpp
```diff
@@ -24,21 +29,157 @@ Copyright (C) 2019 by Albert Weterings
...
+        _curtain_position_set = _curtain_last_position;
+
+    if(_curtain_initial_position_set == false) {//The curtain stopped moving for the first time, set the position back to 
+        int init_position = getSetting("curtainInitialBehaviour", 0);
+        DEBUG_MSG_P(PSTR("[CURTAIN] curtainInitialBehaviour : %d, curtainInitialPosition : %d\n"), init_position, getSetting("curtainInitialPosition", 100));
+        if (init_position == 1)
+             _KACurtainSet(CURTAIN_BUTTON_CLOSE);
+        else if (init_position == 2) 
+             _KACurtainSet(CURTAIN_BUTTON_OPEN);
+        else if (init_position == 3) {
```
mcspr: I think we are missing explanation of curtain behaviour numbers?

code/espurna/curtain_kingart.cpp
```diff
@@ -87,108 +235,248 @@ To configure the device:
...
+
+//------------------------------------------------------------------------------
+void _curtainWebSocketUpdate(JsonObject& root) {
+    JsonObject& state = root.createNestedObject("curtainState");
+    state["curtainGet"] = _curtain_last_position;
+    if(_curtain_position_set == CURTAIN_POSITION_UNKNOWN)
+        _curtain_position_set = _curtain_last_position;
+    state["curtainSet"] = _curtain_position_set;
+    state["curtainButton"] = _curtain_last_button;
+    state["curtainMoving"] = _curtain_moving ? "Moving" : "Idle";
```
mcspr: `bool` directly? there are only two possible 'states'
same for js, no need to compare string.

code/espurna/curtain_kingart.cpp
```diff
@@ -87,108 +235,248 @@ To configure the device:
...
+        } else if(data.containsKey("button")){
+            _curtain_last_button = data["button"].as<int>();
+             _KACurtainSet(_curtain_last_button);
+        }
+    } else if (strcmp(action, "dbgcmd") == 0) { //Directly send our buffer to the KA serial
+            if(data["command"] == "debug") {
+                _debug_flag = true;
+            } else
+                _KACurtainSend(data["command"]);
+    }
```
mcspr: This should be handled by terminal commands?
`terminalRegisterCommand(F("KAUART.DEBUG"), ...)`
`terminalRegisterCommand(F("KAUART.SEND"), ...)`?

code/espurna/curtain_kingart.cpp
```diff
@@ -24,21 +29,157 @@ Copyright (C) 2019 by Albert Weterings
...
+//------------------------------------------------------------------------------
+//Stop moving will set the real curtain position to the GUI/MQTT
+void _KAStopMoving() {
+     _curtain_moving = false;
+     if( _curtain_position != CURTAIN_POSITION_UNKNOWN)
+        _curtain_position_set = _curtain_position;
+    else if( _curtain_last_position != CURTAIN_POSITION_UNKNOWN)
+        _curtain_position_set = _curtain_last_position;
+
+    if(_curtain_initial_position_set == false) {//The curtain stopped moving for the first time, set the position back to
```
mcspr:
```suggestion
    if (!_curtain_initial_position_set) { //The curtain stopped moving for the first time, set the position back to 
```

code/espurna/curtain_kingart.cpp
```diff
@@ -87,108 +235,248 @@ To configure the device:
...
+        _curtain_position_set = _curtain_last_position;
+    state["curtainSet"] = _curtain_position_set;
+    state["curtainButton"] = _curtain_last_button;
+    state["curtainMoving"] = _curtain_moving ? "Moving" : "Idle";
+    state["curtainType"] = getSetting("curtainType", "0");
+}
+
+//------------------------------------------------------------------------------
+void _curtainWebSocketStatus(JsonObject& root) {
+    root["curtainVisible"] = 1;
```
mcspr: Already set in onvisible

code/espurna/curtain_kingart.cpp
```diff
@@ -87,108 +235,248 @@ To configure the device:
...
+// -----------------------------------------------------------------------------
+// WEB Support
+// -----------------------------------------------------------------------------
+
+#if WEB_SUPPORT
+
+//------------------------------------------------------------------------------
+void _curtainWebSocketOnConnected(JsonObject& root) {
+    root["curtainType"] = getSetting("curtainType", "0");
+    root["curtainInitialBehaviour"] = getSetting("curtainInitialBehaviour", "0");
```
mcspr: `curtainInitialBehaviour` -> `curtainBoot`?
"0" is only explained in web

mcspr (changes requested):
Thanks!

- If you don't mind going over code comments once more, please verify that those still apply.
- We should explain magic numbers in code (and, if possible, in log too). I've added some helpers for getSetting() 2nd arg, we can map any type from string. see `namespace settings { namespace internal` in relays / buttons / sensors.
- We don't enforce any specific style / formatting... But one thing I am not a fan of is `if (cond) expr;`, unless those are one-liners like `if (something) return;`. Having brackets is preferable.

code/espurna/curtain_kingart.cpp
echauvet: Done

code/espurna/curtain_kingart.cpp
echauvet: Done

code/espurna/curtain_kingart.cpp
echauvet: Done

code/espurna/curtain_kingart.cpp
echauvet: Done

code/espurna/curtain_kingart.cpp
echauvet: Right

code/espurna/curtain_kingart.cpp
echauvet: curtainBt/curtainBtPos

code/espurna/curtain_kingart.cpp
echauvet: I need more time to understand this.

code/espurna/curtain_kingart.cpp
echauvet: Done

code/html/custom.js
echauvet: It seems to increase the footprint, no?

code/espurna/curtain_kingart.cpp
echauvet: Let see this later - Terminal command for this device?

code/espurna/curtain_kingart.cpp
echauvet: Done

code/espurna/scheduler.cpp
```diff
@@ -62,7 +71,7 @@ void _schWebSocketOnConnected(JsonObject &root){
         switch_.add(getSetting({"schSwitch", i}, 0));
         action.add(getSetting({"schAction", i}, 0));
         type.add(getSetting({"schType", i}, SCHEDULER_TYPE_SWITCH));
-        hour.add(getSetting({"schHour", i}, 0));
+        hour.add(getSetting({"schHour", i}, 12));
```
mcspr: Can you explain why change to 12?

code/espurna/scheduler.cpp
```diff
@@ -102,15 +111,15 @@ void _schConfigure() {
 
                 bool sch_enabled = getSetting({"schEnabled", i}, false);
                 int sch_action = getSetting({"schAction", i}, 0);
-                int sch_hour = getSetting({"schHour", i}, 0);
+                int sch_hour = getSetting({"schHour", i}, 12);
```
mcspr: Can you explain why change to 12?

code/espurna/scheduler.cpp
```diff
@@ -230,7 +246,7 @@ void _schCheck(int relay, int daybefore) {
         String sch_weekdays = getSetting({"schWDs", i}, SCHEDULER_WEEKDAYS);
         if (_schIsThisWeekday(sch_utc ? calendar_weekday.utc_wday : calendar_weekday.local_wday, sch_weekdays)) {
 
-            int sch_hour = getSetting({"schHour", i}, 0);
+            int sch_hour = getSetting({"schHour", i}, 12);
```
mcspr: Can you explain why change to 12?

code/html/custom.js
mcspr: The idea was to shorten the string where the action happens. We can easily see that values are actually set

code/espurna/curtain_kingart.cpp
```diff
@@ -87,112 +243,266 @@ To configure the device:
...
+        } else if(data.containsKey("button")){
+            _curtain_last_button = data["button"].as<int>();
+             _KACurtainSet(_curtain_last_button);
+        }
+    } else if (strcmp(action, "dbgcmd") == 0) { //Directly send our buffer to the KA serial
+            if(data["command"] == "debug") {
+                _curtain_debug_flag = true;
+            } else
+                _KACurtainSend(data["command"]);
+    }
```
mcspr: re Terminal, you **only** handle WebUI this way, while command handles every input from Telnet / http / webui / etc.
(and which also tries to handle the command that you send it, you just don't see the error message there)

Commands are akin to the main() func, we get argc and argv[],so argv[0] == command name and argc > 1 means that we have some params.

code/espurna/scheduler.cpp
```diff
@@ -311,7 +337,7 @@ void schSetup() {
 
         static bool restore_once = true;
         if (restore_once) {
-            for (unsigned char i = 0; i < relayCount(); i++) {
+            for (unsigned char i = 0; i < schedulableCount(); i++) {
```
mcspr, line 345: Doesn't this step on schCheck id resolution? If we have 1 relay and 1 curtain, for example.

mcspr (commented):
Let's think a bit more about the scheduler stuff? I am not sure restoring works like that, but I'd expect normal operation mode does. This can be a separate patch, not to complicate things further.

If you are ok with kingart .cpp stuff, I'll reformat it for the merge.

code/espurna/scheduler.cpp
echauvet: The default value in the custom.js UI is set to 12h, if not edited, the final stored value will be 0h even if you use save button. After reboot it will be 0h.
I added schType in change listening too. It seems that light schedule suffer from a bug which stores a light schedule as switch schedule.

code/espurna/scheduler.cpp
echauvet: I've done a big answer and as a result:
image
 So the quick answer is that I tested all with a dummy switch in hardware.h and you can choose between both types and it works:
image

Switch and curtain schedule task worked.
This is how the light schedule bug has been detected (it stays as a switch schedule instead of light schedule)

code/espurna/scheduler.cpp
echauvet: Same as previous answer

code/espurna/scheduler.cpp
echauvet: Same as previous answer

code/espurna/scheduler.cpp
mcspr: Instead of changing default, we should change UI to display the correct value.
This looks like bug from #1986, 12 should not have been there

echauvet:
How do we do ? I start a correction, you start it ? let me know.   De : Max ProkhorovEnvoyé le :samedi 16 mai 2020 22:14À : xoseperez/espurnaCc : Eric Chauvet; AuthorObjet :Re: [xoseperez/espurna] Kingart curtain switch UI support (#2250) @mcspr commented on this pull request.In code/espurna/scheduler.cpp:> @@ -62,7 +71,7 @@ void _schWebSocketOnConnected(JsonObject &root){         switch_.add(getSetting({"schSwitch", i}, 0));         action.add(getSetting({"schAction", i}, 0));         type.add(getSetting({"schType", i}, SCHEDULER_TYPE_SWITCH));-        hour.add(getSetting({"schHour", i}, 0));+        hour.add(getSetting({"schHour", i}, 12));Instead of changing default, we should change UI to display the correct value.This looks like bug from #1986, 12 should not have been there—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

code/espurna/curtain_kingart.cpp
echauvet: Removed it from now

AlbertWeterings:
Great work you are doing Eric, your work makes the Curtain switch usable for many people that are not using it only over MQTT. I did not have the knowledge and time to build the web interface so I took off after MQTT was finished and my 9 shutters where moving the way I wanted them to move.

In your code I fount this:
    // Need to send confirmation to the N76E003AT20 that message is received
    // ECH : TODO Check this is the case every time
    _KACurtainSend("AT+SEND=ok");

I can confirm it is mandatory to confirm to the N76E003AT20 that every message send by it is received. When I started working on it I found that after power up the switch it takes about 30sec before the N76E003AT20 starts sending messages (just not log enough to flash the ESP) so at first I had jumpers on the data lines to disconnect both chips from each other and be able to flash ESPURNA. After silencing the N76E003AT20 flashing is possible. Now a day's we have to OTA flash in two stages this is possible as long the N76E003AT20 stay's quiet after a power cycle flashing is not possible anymore as the N76E003AT20 is corrupting the flash process.

I have been logging communication between both chips with the original software for quite some time to fully understand how it works.

mcspr:
> Now a day's we have to OTA flash in two stages this is possible as long the N76E003AT20 stay's quiet after a power cycle flashing is not possible anymore as the N76E003AT20 is corrupting the flash process.

re. some discussions in the Core repository and general observations about reboot.
Perhaps we need to do Serial.end() before doing OTA? Chip persist some of the configuration between sw reboots, including pins state / uart / etc.

AlbertWeterings:
b.t.w. Flashing over serial also fails as the N76E003AT20 starts sending data during flashing. This can be solved by holding it's pin 4 or on the rear side of the board RES1 to GND this keeps the N76E003AT20 in reset state.

mcspr (approved):
Merging as-is b/c curtain part is done, delaying scheduler changes (?) for later.

code/espurna/scheduler.cpp
mcspr: In this specific block, schCheck will recursively check every 'rule' that uses specific relayID (index from old relayCount()) and goes back a week day-by-day to check if rule applies in the past.
Although, we don't count light channels, so it's still sort-of broken because of non-matching ID / indexes.

echauvet: Understood! In fact, we need specific processing for curtain.

echauvet:
@AlbertWeterings the solution I found is to calibrate the curtain to a very long period for intial flash so the MCU does not send messages until the closed position is not reached. Do you confirm that this is the AT+UPDATE message "setclose":0 "switch":"off" sent at the end of the first move which breaks the update? If so, just calibrate curtain opening to > 60s - this will be enough to keep MCU quiet.

AlbertWeterings:
@echauvet This is a nice work around at the moment you are blocked from flashing after something went wrong. I think good to document. I preferer to silence the MCU in the software by confirming the last message is understood. I did a quick test, you can preform a OTA 2 stage update without problems when ESP is rebooting the MCU doesn't. Only if someone is touching the buttons the MCU starts repeating it's last message and flashing is not possible. Disabling the serial port during OTA flash procedure as @mcspr referred to is the best sollution.

Issue #2232, opened 2020-04-28
More information on 'platformio_override.ini' usage.


davebuk:
I was looking through the files and noticed a reference to `platformio_override.ini`. I guess it allows an easier way to build custom setups without effecting the main repo files? I was looking for more info on its use.

I already have a `custom.h` file but have been using the `ITEAD_SONOFF_BASIC` environment as a starting point and adjusting its definitions in the custom.h file. I also set `-DUSE_CUSTOM_H` to the default build flags and `platform = ${common.platform_latest}` so all builds use the latest version.

How does the `platformio_override.ini` work with platformIO? Would I create the file similar to the details below and define in `custom.h` `#elif defined(MY_SONOFF_BUILD)`

platformio_override.ini
```
[env:esp8266-1m-latest-base]
platform = ${common.platform_latest}
board = ${common.board_1m}
board_build.ldscript = ${common.ldscript_1m}

[env:my_sonoff_build]
extends = env:esp8266-1m-latest-base
src_build_flags = -DUSE_CUSTOM_H
```

mcspr:
- `[env:esp8266-1m-latest-base]` is redundant, 'extends =' will read the base file (see platformio_ota.ini that handles old -ota envs)
-  `env:` name isn't used to provide hardware flag, you need to set it yourself
`src_build_flags = -DMY_SONOFF_BUILD -DUSE_CUSTOM_H`
- also note that values will not be merged, you are always overriding the existing value. for example, to re-use `src_build_flags` from other env, you would need to have something like `src_build_flags = ${env:itead-sonoff-basic.src_build_flags} -DSOME_OTHER_FLAG`

davebuk:
_EDIT_

So for my custom Sonoff Basics I could have:

`platformio_override.ini`
```
[env:my-sonoff-build01]    // My custom Sonoff config
extends = env:esp8266-1m-latest-base
src_build_flags = -DMY_SONOFF_BUILD01 -DUSE_CUSTOM_H

[env:my-sonoff-build02]    // Another custom Sonoff config
extends = env:esp8266-1m-latest-base
src_build_flags = -DMY_SONOFF_BUILD02 -DUSE_CUSTOM_H

[env:my-sonoff-build03]    // Yet another custom Sonoff config
extends = env:esp8266-1m-latest-base
src_build_flags = -DMY_SONOFF_BUILD03 -DUSE_CUSTOM_H
```

Then in `custom.h` I would have:
```
[...]

PR #2257, opened 2020-05-19
Add '.example' files.

davebuk:
Ref #2232 add, add platformio_override.ini.example and custom.h.example files.


code/espurna/config/custom.h.example
```diff
@@ -0,0 +1,95 @@
...
+// See 'code/espurna/config/general.h' for default settings.
+//
+// See: https://github.com/xoseperez/espurna/wiki/Software-features#enabling-features
+// and 'code/platformio_override.ini.example' for more details.
+// ------------------------------------------------------------------------------
+
+
+// Example of a Sonoff Basic board with some options disabled to reduce firmware size.
+
+#if defined(MY_SONOFF_BUILD01)    // This matches the 'src_build_flags = -DMY_SONOFF_BUILD01'
```
mcspr: *eaasl, maybe I misunderstand what 'matches' means here*

"This section will be used when `src_build_flags` contains -DMY_SONOFF_BUILD01"?

code/platformio_override.ini.example
```diff
@@ -0,0 +1,37 @@
...
+# own file: 'platformio_override.ini'
+# 'platformio_override.ini' allows users to create their own environments to add, remove or change
+# firmware configurations. 
+#
+# See: https://github.com/xoseperez/espurna/wiki/PlatformIO#customize-build-settings
+# for more details.
+# -------------------------------------------------------------------------------------------------------------
+
+// Example for a standard Sonoff Basic R2 or R3 but enabling DHT support on GPIO 2.
+// It can be built from VSCode/PlatformIO by selecting the 'my-sonoff-build-dht'
```
mcspr: "selecting ... build selection" -> "selecting the 'Build: my-sonoff-build-dht'"?

code/platformio_override.ini.example
```diff
@@ -0,0 +1,37 @@
...
+# Either copy and paste this file, then rename removing the .example or create your
+# own file: 'platformio_override.ini'
+# 'platformio_override.ini' allows users to create their own environments to add, remove or change
+# firmware configurations. 
+#
+# See: https://github.com/xoseperez/espurna/wiki/PlatformIO#customize-build-settings
+# for more details.
+# -------------------------------------------------------------------------------------------------------------
+
+// Example for a standard Sonoff Basic R2 or R3 but enabling DHT support on GPIO 2.
```
mcspr: afaik, there are only 2 allowed comment prefixes - `;` and `#`. I don't think `//` will work

Implementation detail:
https://docs.python.org/3/library/configparser.html#customizing-parser-behaviour
> comment_prefixes, default value: ('#', ';')

https://github.com/platformio/platformio-core/blob/3ef96cb215024709f463d5b1b5574ad66bca3ebf/platformio/project/config.py#L70 for value lists
https://github.com/platformio/platformio-core/blob/3ef96cb215024709f463d5b1b5574ad66bca3ebf/platformio/project/config.py#L94 for other stuff

code/espurna/config/custom.h.example
davebuk: Yes, your version makes more sense. I was trying to say the the name in brackets `( MY_SONOFF_BUILD01)` has to be same as the `src_build_flags` name in platformio.override.ini

code/platformio_override.ini.example
davebuk: I have got my comment prefixes muddled between the `.h` and `.ini` files. :-)

code/platformio_override.ini.example
davebuk: Again yes, that makes more sense.

davebuk (commented):
Two files updated.

Issue #2231, opened 2020-04-28
Mixed Sensors BH1750 + BMP180 : Calibration


kris969:
Hi,

Thanks for your great job, I'm using Espurna for many years ;)
I'm not sure it's an issue, maybe I was wrong somewhere...

I have generated a LOLIN 8266 node, based on espurna 1.14.2-dev, using relative compile option in platformio.ini for the two sensors as mentioned in the title, it works fine. I receive from it : light, pressure and temperature values as expected.

My surprise is about the Web GUI for this node : 
In the sensor configuration tab, I was expected to find three entries (Temperature, pressure and Lux) for calibration, only the one for temperature is available here.
So here is my question: 
Is there additional compile option to set, or is it an issue?

Let me know if you need additional information.

Regards
Kris

mcspr:
Do you mean the missing  ..._CORRECTION flag setting or something related to the sensor calibration?
*edit: I missed the flag description, value can be negative*

kris969:
Cause pictures are more clear than lot of words, I propose you this:
Here is the status tab, where sensors appears:
![Espurna-2020-04-28 09-11-24](https://user-images.githubusercontent.com/11741168/80458572-49a25680-8931-11ea-9034-cc6b24334f81.png)

here is the sensors configuration tab:
![Espurna-2020-04-28 09-12-06](https://user-images.githubusercontent.com/11741168/80458679-6f2f6000-8931-11ea-9121-1488813c3ebc.png)

Only information relative to temperature is available.

mcspr:
Referencing existing code, we have similar <s>absolute</s>one-point 'correction' for lux (through `luxCorrection` settings key). It is not in the WebUI, though, as seen above.
Nothing for pressure atm and no other methods like 2-point calibration, but I guess that can be easily implement.

davebuk:
Which file/files would need to be modified? Is it just a matter of using the code that allows temperature calibration in the webui and adding the necessary lines of code for LUX and PRESSURE?

mcspr:
I think I misunderstood the original issue. Re-reading, I think the summary is:

#### Missing `luxCorrection` input in WebUI, like this one:
https://github.com/xoseperez/espurna/blob/18dea8902d32f57eebb99cad67d7c9a57a10ab98/code/html/index.html#L1677 (also see surrounding form)
[...]

PR #2265, opened 2020-05-29
Streamline correction settings

mcspr:
fix? #2231 

- group correction settings for known types, add webui detection, relax constraints for values
- de-hardcode webui magnitude and error types


davebuk:
Did you want me to try your code and build over the weekend or shall I wait until you have merged into the main xoseperez dev code?

mcspr:
If you can try it out like this, please do. It would be good to verify nothing else in the sensors breaks because of this change.

davebuk:
Looking good!
These are from a BME280.
**BEFORE**
bmx01
**AFTER**
bmx02

The order is which Pressure and Humidity are shown between the status page and sensor page are different but you wouldn't normally be looking at both pages at the same time.

MQTT values are correct and seen by openHAB.

I can try a Lux later today if you like?

davebuk:
Second device with BH1750 sensor.
**BEORE**
lux01
**AFTER POSITIVE CORRECTION**
lux02
**AFTER NEGATIVE CORRECTION**
lux03

mcspr:
Thanks!

I did not notice the order. Corrections page is generated outside of sensor class, so numerical sorting comes into play (Temperature, Humidity, Pressure). Sensor though uses a special logic to exclude humidity when sensor happens to be BM**P**280, so naturally it is the last one.

PR #2268, opened 2020-06-01
RPN: check relay status before triggering

mcspr:
Since we can't be sure rule won't trigger repeatedly. 

https://gitter.im/tinkerman-cat/espurna?at=5ed456e52c49c45f5ab4e326
@Grimmy347


Grimmy347:
Worked perfect for me.  Thank you!

PR #2269, opened 2020-06-02
Implement support for prodino wifi esp

### LINKED ISSUE #1311: add support for prodino wifi esp
I've a prodino wifi esp. I'm interested in get it supported.

Since the device have onboard an MCP23S08 to manage input and output, I have to wait for virtual gpio support.

Regards

dpeddi:
As promised, here is a pull request to implement basic support for prodino wifi esp (fix #1311)

Some of the code for the Expander is borrowed by the original board developer (I left them in the copyright statement)

The relay works great.

The opto works as well, but since I connected them to some button, I have some bounce.

I think you can accept this request, and later we could move the expander poll routine in DebounceEvent


code/espurna/button.cpp
```diff
@@ -21,6 +21,9 @@ Copyright (C) 2016-2019 by Xose Pérez <xose dot perez at gmail dot com>
 #include "relay.h"
 #include "light.h"
 #include "ws.h"
+#ifdef MCP23S17_SUPPORT
+#include "mcp23s17.h"
+#endif
```
mcspr: No need to guard headers?

code/espurna/mcp23s17.cpp
```diff
@@ -0,0 +1,222 @@
...
+Copyright (C) 2016-2017 by Xose Pérez <xose dot perez at gmail dot com>
+Copyright (C) 2016 Plamen Kovandjiev <p.kovandiev@kmpelectronics.eu> & Dimitar Antonov <d.antonov@kmpelectronics.eu>
+
+*/
+
+#include "mcp23s17.h"
+
+#if MCP23S17_SUPPORT
+
+#define CS 0x0F
```
mcspr: Can this be changed to a setting?
Pls also use decimal

code/espurna/mcp23s17.cpp
```diff
@@ -0,0 +1,222 @@
...
+ * @brief Relay pins.
+ */
+const uint8_t RELAY_PINS[DUMMY_RELAY_COUNT] =
+{ REL1PIN, REL2PIN, REL3PIN, REL4PIN };
+
+/**
+ * @brief Input pins.
+ */
+const int OPTOIN_PINS[OPTOIN_COUNT] =
+{ IN1PIN, IN2PIN, IN3PIN, IN4PIN };
```
mcspr: I had noticed these are defined in hardware header.
- please prefix with module name
- can we provide default values?

code/espurna/mcp23s17.cpp
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @param relayNumber Number of relay from 0 to RELAY_COUNT - 1. 0 - Relay1, 1 - Relay2 ...
+ * @param state New state of relay, true - On, false = Off.
+ *
+ * @return void
+ */
+void MCP23S17SetRelayState(uint8_t relayNumber, bool state)
+{
+	// Check if relayNumber is out of range - return.
+	if (relayNumber > RELAY_COUNT - 1)
```
mcspr: `relayCount()`?

code/espurna/mcp23s17.cpp
```diff
@@ -0,0 +1,222 @@
...
+ * @brief Get opto in state.
+ *
+ * @param optoInNumber OptoIn number from 0 to OPTOIN_COUNT - 1
+ *
+ * @return bool true - opto in is On, false is Off. If number is out of range - return false.
+ */
+bool MCP23S17GetOptoInState(uint8_t optoInNumber)
+{
+	// Check if optoInNumber is out of range - return false.
+	if (optoInNumber > OPTOIN_COUNT - 1)
```
mcspr: `mcp23s17OptoinCount()`?

code/espurna/mcp23s17.cpp
```diff
@@ -0,0 +1,222 @@
...
+#define INTF    0x07
+#define INTCAP  0x08
+#define GPIO    0x09
+#define OLAT    0x0A 
+
+/**
+ * @brief Relay pins.
+ */
+const uint8_t RELAY_PINS[DUMMY_RELAY_COUNT] =
+{ REL1PIN, REL2PIN, REL3PIN, REL4PIN };
```
mcspr: Note that relay count is not always available / can change at runtime. Maybe...
- define max value, gpio_none default
- use runtime vector, push_back on setup

code/espurna/mcp23s17.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "espurna.h"
+
+#if MCP23S17_SUPPORT
+
+#include <SPI.h>
+
+// Inputs and outputs count.
+#define RELAY_COUNT  4
+#define OPTOIN_COUNT 4
```
mcspr: Pls use prefixes.
Should these be ifndef guarded so we can override values?

code/espurna/config/hardware.h
```diff
@@ -4689,6 +4689,68 @@
...
+	#define IN3PIN 0x02
+	#define IN4PIN 0x03
+
+    // Buttons
+    #define BUTTON1_RELAY           1
+    #define BUTTON2_RELAY           2
+    #define BUTTON3_RELAY           2
+    #define BUTTON4_RELAY           1
+
+    #define BUTTON_EVENTS_SOURCE    BUTTON_EVENTS_SOURCE == BUTTON_EVENTS_SOURCE_MCP23S17
```
mcspr:
```suggestion
    #define BUTTON_EVENTS_SOURCE    BUTTON_EVENTS_SOURCE_MCP23S17
```

code/espurna/button.cpp
```diff
@@ -612,6 +616,18 @@ void _buttonLoopSonoffDual() {
 
 }
 
+#ifdef MCP23S17_SUPPORT
```
mcspr: ifdef -> if
Please add _support default 0 to the general.h

code/espurna/config/hardware.h
```diff
@@ -4689,6 +4689,68 @@
...
+    #define BUTTON1_RELAY       1
+    #define BUTTON2_RELAY       2
+
+    // Relays
+    #define RELAY1_PIN          12
+    #define RELAY2_PIN          13
+
+    #define RELAY1_TYPE         RELAY_TYPE_NORMAL
+    #define RELAY2_TYPE         RELAY_TYPE_NORMAL
+*/
```
mcspr: Drop the comment?

code/espurna/mcp23s17.cpp
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @param pinNumber Pin number for set.
+ * @param mode direction mode. 0 - INPUT, 1 - OUTPUT.
+ *
+ * @return void
+ */
+void MCP23S17InitGPIO()
+{
+	// Relays.
+	for (uint8_t i = 0; i < DUMMY_RELAY_COUNT; i++)
```
mcspr: `relayCount()`?

mcspr (changes requested):
- we don't have any formatting tool (yet, clang-format is weird....), but please follow the suggested style guideline. tabs -> spaces, at least in the existing source that you have modified
- `code/test/build` config should be added so we can check if this code builds. see rfbridge as an example
- MCP23S17-specific configuration should have prefixes

code/espurna/button.cpp
dpeddi: what do you mean with guarded?

code/espurna/mcp23s17.cpp
dpeddi: _"please prefix with module name"_ 
done

_can we provide default values?_ 
no idea on how to do that in the correct way

code/espurna/mcp23s17.cpp
dpeddi: changed as DUMMY_RELAY_COUNT

(i don't want to import relay.h)

code/espurna/mcp23s17.cpp
dpeddi: i din't plan to implement such function

code/espurna/mcp23s17.h
dpeddi: done/moved into hardware.h

code/espurna/config/hardware.h
dpeddi: mistake... fixed

code/espurna/button.cpp
dpeddi: done

code/espurna/config/hardware.h
dpeddi: done

code/espurna/mcp23s17.cpp
dpeddi: i didn't plan to include relay.h here

dpeddi:
I applied most changes you suggested, but i'm not able to do it better.

code/espurna/mcp23s17.cpp
mcspr: 'guards' I was mentioning earlier are constructs like:
```cpp
#ifndef FOO
#define FOO "bar"
#endif
```

`#if SOMETHING_SUPPORT` can be used to exclude code, it will always be 0 even without `#ifndef` block above. `#ifdef SOMETHING_SUPPORT`, however, means that we used either `-DSOMETHING_SUPPORT=0` or have a `#define SOMETHING_SUPPORT` (even without any value attached to it)

code/espurna/mcp23s17.cpp
mcspr: Ah, yes, no need for relay-relay here. This counts as a module setting, my mistake.

mcspr:
Configuring still confuses me. Does MCP need to know about the purpose of the buttons / relays / etc.? Should OPTOIN, RELAY etc. be removed / renamed, instead we could somehow describe PINS as IN or OUT? DUMMY_RELAY_COUNT is even weirder, since these are actual GPIOs. Right now we depend on a lot of assumptions?

I had noticed the debouncing, but I don't think I intended to do it that way overall. [libs/BasePin.h](https://github.com/xoseperez/espurna/blob/dev/code/espurna/libs/BasePin.h) (also see [basic GPIO](https://github.com/xoseperez/espurna/blob/dev/code/espurna/gpio.h)) class was *supposed* to allow a hardware like this to be used directly with DebounceEvent class. 
Meaning, PROVIDER stuff could still use basic GPIO code, only replacing GPIO behavior to call SPI instead of the hardware stuff.

Buttons already use basepin. Relays don't, but I can look into it instead.

dpeddi:
>Configuring still confuses me. 

>Does MCP need to know about the purpose of the buttons / relays / etc.? 
the MCP is an expander... what i've done is probably still related to the hardware i own (the prodino wifi)

>Should OPTOIN, RELAY etc. be removed / renamed,
Probably yes in the MCP.... could be IN/OUT
> instead we could somehow describe PINS as IN or OUT? 

>DUMMY_RELAY_COUNT is even weirder, since these are actual GPIOs. 
correct... original implementation from [kmpelectronics](https://github.com/kmpelectronics/Arduino/blob/master/ProDinoWiFiEsp/src/PRODINoESP8266/src/KMPDinoWiFiESP.cpp) uses RELAY_COUNT. I've renamed in DUMMY_RELAY_COUNT since it was what i've seen in espurna.

>Right now we depend on a lot of assumptions?

>I had noticed the debouncing, but I don't think I intended to do it that way overall. libs/BasePin.h (also see basic GPIO) class was supposed to allow a hardware like this to be used directly with DebounceEvent class.
>Meaning, PROVIDER stuff could still use basic GPIO code, only replacing GPIO behavior to call SPI instead of the hardware stuff.

Yes, on my board i've connected some button to the optoin, so i need debounce... but probably it should be configurable from the webif

>Buttons already use basepin. Relays don't, but I can look into it instead.

As said before, i'm available to improve the code, but I can't do better... If you are interested by, feel free to improve it, or help me with lots of hints :-)

mcspr:
Maybe, idk when though :)

The basic idea with buttons is:
```cpp
struct McpGpioPin final : public BasePin {
    // "default" constructor based on BasePin(unsigned char)
    using BasePin::BasePin;

    int digitalRead() override {
        return MCP23S17GetOptoInState(pin)
    }
    void digitalWrite(uint8_t val) override {
        MCP23S17SetRelayState(pin, val);
    }
    void pinMode(uint8_t mode) override {
        MCP23S17SetDirection(pin, OUTPUT);
    }
};
```
Then, just create McpGpioPin instead of GpioPin in the setup code. When GENERIC button loop happens, DebounceEvent will poll those functions for data. I wonder how fast that would be though, SPI vs. essentially 0-cost register reads.

dpeddi:
and what about the constructor EventEmitter::EventEmitter?
There are lot of code related to gpio... Should it be disabled with #ifndef MCP_SUPPORT?

pin will still be Type::Pin? What would be it's value range?

I think this can be extended for other extender.. I've seen people interested on pcf as well.

mcspr:
I'd start with just replacing GpioPin creating with MCP, when it is enabled. BasePin methods are 'virtual', https://en.cppreference.com/w/cpp/language/virtual. So,
```cpp
GpioPin pin(5);
BasePin* as_base = &pin;
as_base->digitalWrite(1);
```
will call GpioPin::digitalWrite()

mcspr:
Also, re
> What would be it's value range?
>
> I think this can be extended for other extender.. I've seen people interested on pcf as well.

`gpioValid()` check at the top of the loop is bogus , I'd guess `!= GPIO_NONE` should suffice. Range, based on config, is just 0...8? (or ...16?).

dpeddi:
> The basic idea with buttons is:
> 
> ```c++
> struct McpGpioPin final : public BasePin {
>     int digitalRead() override {
>         return MCP23S17GetOptoInState(pin)
>     }
>     void digitalWrite(uint8_t val) override {
>         MCP23S17SetRelayState(pin, val);
>     }
>     void pinMode(uint8_t mode) override {
>         MCP23S17SetDirection(pin, OUTPUT);
>     }
> };
> ```
Struct or class like gpio.h?

mcspr:
Either. Check out https://en.cppreference.com/w/cpp/language/access and https://en.cppreference.com/w/cpp/language/class

C++ struct is a class with hidden public: at the top
(or class is a struct with private: at the top)

dpeddi:
can-t figure out this error:
```
espurna\button.cpp: In function 'void buttonSetup()':
espurna\button.cpp:753:49: error: no matching function for call to 'make_shared(const unsigned char&)'
                 std::make_shared<McpGpioPin>(pin), config,
```

mcspr:
Oops, my bad. There needs to be  `using BasePin::BasePin;` in the body of the class to proxy the original constructor.
(or do the same thing GpioPin does, where constructor calls the parent directly e.g. `CustomPin(unsigned char pin) : BasePin(pin) {}`)

dpeddi:
I have been able to build with override. Tests follows

mcspr:
Did it work ❓

dpeddi:
This period is quite intense and I have no time to test yet

dpeddi:
I've flashed the reworked version, but i get not webif. It seems it crash somewere. I have to do some troubleshooting

mcspr:
Can you share a snippet of the code related to the web?

Most common reason for ws / webserver callbacks to crash - trying to use APIs that use delay / yield to 'wait' for something, because we can't switch contexts from system callback (which is where we are in all of the networking stuff) back to the system, it is hard-wired to crash.

Or maybe it's just a resources problem, idk. platformio ships exception decoder via device monitor or you can also try using https://github.com/mcspr/esparduinoexceptiondecoder with c/p of the trace (note the `-s` switch to only handle the stack)
*edit:* note that trace decoding will work best with -git or -latest core versions toolchain, default 2.3.0 has some weird problems incorrectly parsing the .elf file

dpeddi:
A question... maybe gpiovalid should be a virtual function?

mcspr:
But are you checking validity via BasePin pointer at any point? If there is no need to do that, might as well have static / member only for the class on question

dpeddi:
_But are you checking validity via BasePin pointer at any point? If there is no need to do that, might as well have static / member only for the class on question_
No it was just a consideration..

Here is the work in progress branch.
https://github.com/dpeddi-iot/espurna/commits/dev-compliant-wip

The code build (at least for my board) but as said before it crash somewhere and espurna starts in recovery mode.

I suspect that something goes wrong with the button polling routine.

mcspr:
Not sure. Only apparent problem is that without gpioValid() check it creates 32 buttons. I also see 32 "Pressed" events right on boot, but that is without any hardware present.

dpeddi:
It works
```
[018189] [WEBSOCKET] #1 connected, ip: 192.168.10.101, url: /ws
[023964] [BUTTON] Button #2 event 1 (Pressed)
[024465] [BUTTON] Button #2 event 2 (Click)
[024466] [RELAY] #2 scheduled ON in 0 ms
[024478] [RELAY] #2 set to ON
[024479] [RELAY] [MCP23S17] Set relay 2 to ON
[025150] [BUTTON] Button #3 event 1 (Pressed)
[025480] [RELAY] Setting relay mask: 0b100
[025654] [BUTTON] Button #3 event 2 (Click)
[025655] [RELAY] #3 scheduled ON in 0 ms
[025667] [RELAY] #3 set to ON
[025667] [RELAY] [MCP23S17] Set relay 3 to ON
[026139] [BUTTON] Button #0 event 1 (Pressed)
[026649] [BUTTON] Button #0 event 2 (Click)
[026650] [RELAY] #0 scheduled ON in 0 ms
[026662] [RELAY] #0 set to ON
[026663] [RELAY] [MCP23S17] Set relay 0 to ON
[027094] [BUTTON] Button #1 event 1 (Pressed)
[027595] [BUTTON] Button #1 event 2 (Click)
[027596] [RELAY] #1 scheduled ON in 0 ms
[027608] [RELAY] #1 set to ON
[027609] [RELAY] [MCP23S17] Set relay 1 to ON
[028610] [RELAY] Setting relay mask: 0b1111
```
need few code cleanup. Please look at my fork

dpeddi:
probably there are a couple of small thing to optimize but we should be near to the end

mcspr:
Cool!

re. previous comment, indeed the default setting for buttons is to expect HIGH signal by default, thus the Pressed event.

It feels like it does not matter right now, as Prodino MCP has fixed configuration and no other options but to use specific ones, but:
- OPTIN seems like just IN, Prodino inputs are optical but it is not MCP thing?
- DUMMY_RELAY_COUNT should not be needed (even so, in the MCP code). But, it is necessary right now, but perhaps relays code can be rewritten in such a way that we could also use RELAY1_PIN, RELAY2_PIN with MCP numbers.
- <s>MCP23S08_..._COUNT should somehow be auto-generated from MCP23S08_..._PIN. / Or isn't even needed as all configuration is external. As I was thinking, perhaps it can be [configuration template a-la buttons setup function](https://github.com/xoseperez/espurna/blob/774a73d1d42c0ff9b0d6305e82100ea1ee9726aa/code/espurna/button.cpp#L688-L711). This is mostly b/c of the fact that MCP pins are GPIOs (per http://ww1.microchip.com/downloads/en/DeviceDoc/MCP23008-MCP23S08-Data-Sheet-20001919F.pdf), </s> As 4 IN and 4 OUT is Prodino-specific thing and configuration is external, we should no longer need \_COUNT flags but depend on RELAY and BUTTON configs which will pinMode() all by itself.

I'll try to gather my thoughts about the relays and can even push something right here for you to test ~ this weekend.

code/espurna/mcp23s08.cpp
```diff
@@ -0,0 +1,238 @@
...
+bool MCP23S08GetOptoInState(uint8_t optoInNumber)
+{
+
+    // Check if optoInNumber is out of range - return false.
+    if (optoInNumber > MCP23S08_OPTOIN_COUNT - 1)
+    {
+        return false;
+    }
+
+    return !MCP23S08GetPin(optoInNumber);
```
mcspr: btw here is the reason for button config, button_default_high can actually work with this inversed

code/espurna/mcp23s08.cpp
```diff
@@ -0,0 +1,238 @@
...
+    return ::MCP23S08GetOptoInState(this->pin);
+}
+
+void MCP23S08Setup()
+{
+    DEBUG_MSG_P(PSTR("[MCP23S08] Initialize SPI bus\n"));
+    // Expander settings.
+    SPI.begin();
+    SPI.setHwCs(MCP23S08_CS);
+    SPI.setFrequency(1000000);
```
mcspr, line 65: Slightly confused about this.
Per http://ww1.microchip.com/downloads/en/DeviceDoc/MCP23008-MCP23S08-Data-Sheet-20001919F.pdf
SPI bus frequency is said to be 10MHz, but here it is 1MHz. Either SPI class does not care or there is some weird thing SPI does that 'sets' frequency for both.

mcspr: Also regarding SPI, I think we can just call begin(). Check out https://github.com/esp8266/Arduino/blob/cc1cc0b2ce4825057f6a8bb0c6b1b82e17be313e/libraries/SPI/SPI.cpp#L65-L87

setHwCs accepts boolean, so pin 15 cannot be changed
I wonder if digitalWrite() calls are also not needed and handled by the SPI library, as it *seems* those are meant to signal SPI busy / not busy, which SPI.cpp kind of does through some weird register calls.

code/espurna/mcp23s08.cpp
dpeddi: I know,

I reused most of the code from the prodinowifi vendor with minimal change.

https://github.com/kmpelectronics/Arduino/tree/master/ProDinoWiFiEsp

Should we diverge so much from their code?

mcspr: We just want getpin & setpin, since we implement button and relay abstractions ourselves.
Note the https://github.com/dpeddi-iot/espurna/pull/1 where it uses getpin directly as digitalRead().

dpeddi:
I've tried your pull request... 

In file included from espurna\libs/DebounceEvent.h:37:0,
                 from espurna\DebounceEvent.cpp:33:
espurna\libs/BasePin.h: In member function 'virtual BasePin::operator bool()':
espurna\libs/BasePin.h:20:16: error: 'GPIO_NONE' was not declared in this scope
         return GPIO_NONE != pin;
                ^
*** [.pio\build\prodino-wifi\src\DebounceEvent.cpp.o] Error 1


Indeed in BasePin.h is missing the include to types.h and/or espurna.h

mcspr:
My bad, the issue is because I was building using `env ESPURNA_BUILD_SINGLE_SOURCE=1 pio run ...`, our [version of unity build](https://en.wikipedia.org/wiki/Single_Compilation_Unit).
We should not include espurna header, but types.h can be included without any issues?

dpeddi:
Already added... build in progress... sound strange to me that your pull was failing!

code/espurna/relay.cpp
```diff
@@ -1457,8 +1473,30 @@ void _relaySetupAdhoc() {
...
 
     _relays.reserve(relays);
-    for (unsigned char id = 0; id < relays; ++id) {
-        _relays.emplace_back(id);
+
+    #if (RELAY_PROVIDER == RELAY_PROVIDER_RELAY)
+        using gpio_type = GpioPin;
+    #elif (RELAY_PROVIDER == RELAY_PROVIDER_MCP23S08)
+        using gpio_type = McpGpioPin;
+    #endif
```
mcspr: Regarding travis issue, it also needs to match RELAY_PROVIDER_LIGHT for every check I added.
So,  `#if (RELAY_PROVIDER == RELAY_PROVIDER_RELAY) || (RELAY_PROVIDER == RELAY_PROVIDER_LIGHT)`

dpeddi:
I've tested on the device and the code works, but can't understand the  quality check and travis errors

mcspr:
Thanks!

I'll do the codacy clean-up then. What it wants is essentially to avoid things like having `void function(DummyPin pin) { ... };` and calling it like `function(15)` instead of `function(DummyPin(15))`. Sadly, it wants it in general, not when it is actually used like that.
*edit:* Pin warning is valid though, and related to the preprocessor check I mentioned above.

Travis failure is unrelated, Nofuss library is unable to build because of recent changes to Arduino Core git

mcspr (approved):

PR #2276, opened 2020-06-09
Add support for AG-L4 v3

andrej-peterka:
I used AG-L4 as base, but changed the pin numbers for buttons and lights.


mcspr:
Thanks! I missed this completely.
Images also added to the AG-L4 page in the wiki

PR #2278, opened 2020-06-16
Please Update Espurna For Sonoff Basic R3 DIY

F4TF1SH:
hi there

i only notice the new version of sonoff basic R3
this sonoff can change the firmware via OTA (over the air) with you app ewelink and the mode DIY (DO IT YOURSELF), you dont need FTDI usb to serial 

so i dont know is compatible another version but i only send message

sorry for mi english

regards


mcspr:
Sonoff basic image is shared between revisions, special builds are there for when using external sensors b/c r1 and r2 have different GPIOs available.

DIY mode by itself has some quirks though. But, maybe we need to at least mention it in the Wiki.

PR #2282, opened 2020-06-20
Rework settings

### LINKED ISSUE #1795: A few things about crash dumps
Hi. a few thing about crash dumps: a little enhancement, one idea and a few questions

- when command crash is run in terminal.ino crashDump() and crashClear() is executed in crash.ino; if there is no crash  the simulated rom gets writed (and commited) anyway; it is not necessary.
Ok, I think that 

```
void crashClear() {
    uint32_t crash_time = 0xFFFFFFFF;
    EEPROMr.put(SAVE_CRASH_EEPROM_OFFSET + SAVE_CRASH_CRASH_TIME, crash_time);
    EEPROMr.commit();
}
```

can be replaced by something like:

```
void crashClear() {
    uint32_t crash_time;
    EEPROMr.get(SAVE_CRASH_EEPROM_OFFSET + SAVE_CRASH_CRASH_TIME, crash_time);
    if (crash_time == 0xFFFFFFFF) //already cleared, there is nothing to do
           return;
    crash_time = 0xFFFFFFFF;
    EEPROMr.put(SAVE_CRASH_EEPROM_OFFSET + SAVE_CRASH_CRASH_TIME, crash_time);
    EEPROMr.commit();
}
```
- crashDump can be modified for conform the format of EspArduinoExpceptionDecoder? https://github.com/me21/EspArduinoExceptionDecoder. There are two field missing (ctx and offset; I don't know what are they for but Decoder requires that) and a few differences is format ("Exception(n):",  "sp: xxxxxxxx end: xxxxxx offset: xxxxxxxx", these addresses without the prefix 0x). Or probably add a new command, say crashDecoder() (in terminal.ino under command crashDecoder) and crashDumpDecoder() in crash.ino for doing that.

- is it possible that crashDump() is doing something wrong? I used dumps in  EspArduinoExpceptionDecoder and the results were wrong;  you can see it in https://github.com/xoseperez/espurna/issues/1752#issuecomment-504721151 (for example, it resolves addresses to non-existent code lines).   Ok, possible the problem is xtensa toolchain, not crashDump...

### LINKED ISSUE #2015: Review Embedis storage library usage

mcspr:
- Drop Embedis, rework internal reverse reverse engineering of the format and just do it directly. fix #2015 
- Fix (again) crash function write direction, free up ~256bytes for settings. Now it lives within [14:256), not [256:whatever) like it was before. fix #1795
- Update crashDump to use Print as output. Kind-of wonky b/c of 2.3.0, but would be interesting to compare the Print generic 'printf_P' footprint. We almost never use malloc here to print things, buffer is much larger than Core's
- Rework crashDump to use upstream stackdump format, fix #1795
- Fix settings storage going over the 'left' limit and over the 0.
- Fix settings incorrectly calculating key size and going over the limit
- Fix settings incorrectly calculating kv size by introducing padding after writing data with 'set'
- Tangentially, fix main headers include order to avoid collisions and allow templated code to be properly instantiated.

Notice:
'KeyValue' class is implemented as a template based on a generic 'Storage' class, implementing read() / write() / commit(). It is easily replacable with virtual implementation (it is also in the history as https://github.com/xoseperez/espurna/pull/2282/commits/4390b9e6b6c07d5e982a406adff42b4eab1b0571). Notably, there is no actual difference between clever lambda dictionary approach from the original library and vtable base class, since we go through the same 2 pointers.
As to why - It is highly annoying to have such an overhead for read / write, when we already implement it through a EEPROM class. I will try to actually test implementations, although it is an interesting theory if 'virtual' implementation in the header can be de-virtualized and match the templated one. Size wise they are already pretty close.

TODO: link with issues


mcspr:
Partly breaking, because of CRASH_SAVE changes. Rolling back a version produces bogus results, as flash contents are garbage at that point.

mcspr:
Another sort-of breaking thing.

For example, we have key "a" with empty value. After this PR, resulting structure is:
```
0x00 0x00 0x00 0x00 0x61 0x00 0x01
              value            key
```
However, original Embedis does not have the 'empty' part of the value and simply uses 2 len bytes as a marker.
```
0x00 0x00 0x61 0x00 0x01
    value            key
```

iirc this was done pretty early, so I forgot to properly mention it. Original PR code was mistakenly assuming that padding at the end was part of the kv structure, so additional bytes were added to the estimation method and the cursor search to ensure len can only be empty when it is immediately followed by two 0 bytes. Knock-on-wood, this 'empty value' situation seems to only happen manually, both WebUI and terminal allow empty strings. WebUI though only allows this when old value was not empty, as trying to set empty flag default results in nothing 

*edit:*
<details>
<summary>For posterity sake, here's the POC diff to make old Embedis storage work seamlessly. Remove empty-value special treatment, fix estimate and make _raw_erase overwrite stuff with 0xff instead of keeping it as-is (as it seems to be another difference in implementations)</summary>

```diff
diff --git a/code/espurna/settings_embedis.h b/code/espurna/settings_embedis.h
index fe047d91..95654580 100644
--- a/code/espurna/settings_embedis.h
+++ b/code/espurna/settings_embedis.h
@@ -31,7 +31,7 @@ struct ValueResult {
     String value;
 };
 
-// We can't save empty keys but can save empty values as 0x00 0x00 0x00 0x00
+// We can't save empty keys but can save empty values as just 2 length bytes
 // total sum is:
 // - 2 bytes gap at the end (will be re-used by the next value length byte)
 // - 4 bytes to store length of 2 values (stored as big-endian)
@@ -41,10 +41,7 @@ inline size_t estimate(const String& key, const String& value) {
         return 0;
     }
 
-    const auto key_len = key.length();
-    const auto value_len = value.length();
-
-    return (4 + key_len + ((value_len > 0) ? value_len : 2));
+    return (4 + key.length() + value.length());
 }
 
 // Note:  KeyValueStore is templated to avoid having to provide RawStorageBase via virtual inheritance.
@@ -398,9 +395,6 @@ class KeyValueStore {
                 while (value_len--) {
                     (--writer).write(value[value_len]);
                 }
-            } else {
-                (--writer).write(0);
-                (--writer).write(0);
             }
 
             // we also need to pad the space *after* the value
@@ -546,7 +540,9 @@ class KeyValueStore {
 
     void _raw_erase(size_t start_pos, Cursor& to_erase) {
         // we either end up to the left or to the right of the boundary
+
         if (start_pos < to_erase.begin) {
+            // shift storage to the right, overwriting over the now empty space
             auto from = Cursor::fromEnd(_storage, start_pos, to_erase.begin);
             auto to = Cursor::fromEnd(_storage, start_pos + to_erase.size(), to_erase.end);
 
@@ -555,10 +551,11 @@ class KeyValueStore {
                 from.write(0xff);
             };
         } else {
-            // just null the length bytes, since we at the last key
+            // overwrite the now empty space with 0xff
             to_erase.resetEnd();
-            (--to_erase).write(0);
-            (--to_erase).write(0);
+            while (--to_erase) {
+                to_erase.write(0xff);
+            }
         }
 
         _storage.commit();
@@ -605,7 +602,7 @@ class KeyValueStore {
                 uint8_t len2 = _cursor.read();
                 --_cursor;
                 if ((0 == len) && (0 == len2)) {
-                    len = 2;
+                    len = 0;
                     _state = State::EmptyValue;
                 } else if ((0xff == len) && (0xff == len2)) {
                     _state = State::End;
@@ -626,12 +623,7 @@ class KeyValueStore {
                         --_cursor;
                     }
                     break;
-                // ...and only read 0's
                 case State::EmptyValue:
-                    while (_cursor && (_cursor.read() == 0) && --left) {
-                        --_cursor;
-                    }
-                    break;
                 default:
                     break;
                 }
@@ -643,7 +635,7 @@ class KeyValueStore {
 
                 // set the resulting cursor as [pos:len+2)
                 out.result = true;
-                out.length = (_state == State::EmptyValue) ? 0 : len;
+                out.length = len;
                 out.cursor.reset(
                     _cursor.begin + _cursor.position,
                     _cursor.begin + _cursor.position + len + 2
```

</details>

PR #2306, opened 2020-07-19
Update pH unit decimals to 3

ruimarinho:
This PR brings back pH unit decimals to 3 (as opposed to 0). In the future it might may sense to make these configurable as different probes will have different resolutions. Right now, the Atlas Scientific pH Probes come with 0.001 resolution.

### DIFF
--- a/code/espurna/sensor.cpp
+++ b/code/espurna/sensor.cpp
@@ -574,6 +574,8 @@ unsigned char _sensorUnitDecimals(sensor::Unit unit) {
             return 1;
         case sensor::Unit::UltravioletIndex:
             return 3;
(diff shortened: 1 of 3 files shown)


mcspr:
Sensor can have `signed char decimals(sensor::Unit unit) override { switch (unit) { ... } }`.

Kind of the same issue as UVI, it's not technically a unit. I wonder if decimals / etc. configuration should work with pairs of magnitude + unit.
*edit:* or, provide 'dimension-less' unit with some custom resolution, since None effectively does not have any

ruimarinho:
I like the idea of a dimension-less unit and it would be up to the sensor to return the resolution override.

PR #2313, opened 2020-08-02
sns/VL53L1X: improve sensor setup and validation

ruimarinho:
Small improvements to VL53L1X although I am getting more and more inclined to switch to the ultra lite driver version given the difference in accuracy.

Normalizes periods to milliseconds and adds some basic config validation.

### DIFF
--- a/code/espurna/sensor.cpp
+++ b/code/espurna/sensor.cpp
@@ -2034,9 +2034,17 @@ void _sensorLoad() {
 
     #if VL53L1X_SUPPORT
     {
+        #if (VL53L1X_INTER_MEASUREMENT_PERIOD <= (VL53L1X_MEASUREMENT_TIMING_BUDGET + 4))
+            #error "[VL53L1X] Intermeasurement period must be greater than the timing budget + 4ms"
+        #endif
+
+        #if (VL53L1X_MEASUREMENT_TIMING_BUDGET < 20 || VL53L1X_MEASUREMENT_TIMING_BUDGET > 500)
+            #error "[VL53L1X] Timing budget is limited to [20, 500]"
+        #endif
+
         VL53L1XSensor * sensor = new VL53L1XSensor();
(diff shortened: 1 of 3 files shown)


code/espurna/sensor.cpp
```diff
@@ -2034,9 +2034,17 @@ void _sensorLoad() {
 
     #if VL53L1X_SUPPORT
     {
+        #if (VL53L1X_INTER_MEASUREMENT_PERIOD <= (VL53L1X_MEASUREMENT_TIMING_BUDGET + 4))
+            #error "[VL53L1X] Intermeasurement period must be greater than the timing budget + 4ms"
+        #endif
+
+        #if (VL53L1X_MEASUREMENT_TIMING_BUDGET < 20 || VL53L1X_MEASUREMENT_TIMING_BUDGET > 500)
+            #error "[VL53L1X] Timing budget is limited to [20, 500]"
+        #endif
```
mcspr, line 2043: idk if this is necessary here

- looking at the library source, you could just check setMeasurementTimingBudget() boolean result by proxying it via the sensor method and setting internal error
- this could also happen with setInterMeasurementPeriod(), never allowing startContiniuous in the begin()

thus, replacing `#error` with debug_msg_p() call

code/espurna/sensors/VL53L1XSensor.h
```diff
@@ -36,13 +36,13 @@ class VL53L1XSensor : public I2CSensor<> {
           _vl53l1x->setDistanceMode(mode);
         }
 
-        void setMeasurementTimingBudget(uint32_t budget_us) {
-          _vl53l1x->setMeasurementTimingBudget(budget_us);
+        void setMeasurementTimingBudget(uint16_t budget_ms) {
+          _vl53l1x->setMeasurementTimingBudget(budget_ms * 1000);
```
mcspr, line 40: Just to note, this implicitly means:
```cpp
_vl53l1x->setMeasurementTimingBudget(static_cast<uint32_t>(static_cast<int>(budget_ms) * 1000)));
```

PR #2319, opened 2020-08-12
implement support for st7735 display

dpeddi:
initial support for st7735 display


dpeddi:
image
image
image
image

code/espurna/thermostat.h
```diff
@@ -8,14 +8,27 @@ Copyright (C) 2017 by Dmitry Blinov <dblinov76 at gmail dot com>
 
 #pragma once
 
+#ifndef THERMOSTAT_H
+#define THERMOSTAT_H
```
mcspr: `#pragma once` already does this btw

code/espurna/thermostat.h
```diff
@@ -8,14 +8,27 @@ Copyright (C) 2017 by Dmitry Blinov <dblinov76 at gmail dot com>
...
 #include <float.h>
 
 #if THERMOSTAT_DISPLAY_SUPPORT
+#if THERMOSTAT_DISPLAY_ST7735_SUPPORT
+#define TOMTHUMB_USE_EXTENDED 1
+#include <gfxfont.h>
+#include <Roboto_Thin9pt8b.h>
+#include "Adafruit_GFX.h"
+#include <Adafruit_ST7735.h>
+#else
```
mcspr: We don't actually want these in the .h, but in the .cpp where the objects are. Dependency resolution is still wonky as it is, this will cause some spurious dependency-updated-lets-rebuild builds for no reason

code/platformio.ini
```diff
@@ -154,8 +154,14 @@ lib_deps =
...
     https://github.com/mcleng/MAX6675-Library#2.0.1
     https://github.com/ThingPulse/esp8266-oled-ssd1306#3398c97
+    Adafruit GFX Library@1.7.5
+    Adafruit ST7735 and ST7789 Library@1.5.15
     Adafruit SI1145 Library@~1.1.1
     https://github.com/BoschSensortec/BSEC-Arduino-library.git#c5503e0
+lib_ignore =
+    SD
+    Adafruit ILI9341
+    Adafruit seesaw Library
```
mcspr, line 167: This does not do anything btw, `[common]` block must be explicitly used as `lib_ignore = ${common.lib_ignore}`. See below, in the `[env]`

Having libraries added is still pain, both b/c of the added compilation time and needless increase in cmdline length for linker and compilation (although, as noted in gitter, only linker breaks for some reason) :/
'Generic' solution is to add another env: entry that depends on display libs or some very clever usage of lib_ignore like we do with -latest and -git builds that ignores NtpClientLib and TimeLib

code/espurna/main.cpp
```diff
@@ -61,6 +61,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #include "web.h"
 #include "ws.h"
 #include "mcp23s08.h"
+#include "thermostat.h"
```
mcspr: It is above, see line 57?

code/espurna/thermostat.cpp
```diff
@@ -488,36 +489,75 @@ void resetBurnCounters() {
...
 #define LOCAL_TEMP_UPDATE_INTERVAL 60000
 #define LOCAL_HUM_UPDATE_INTERVAL  61000
 
+#if THERMOSTAT_DISPLAY_ST7735_SUPPORT
+/*#define cs   D8
+#define dc   D3
+#define rst  D4*/
+#define cs   D8
+#define dc   D4
+#define rst  -1
```
mcspr: D# are board-specific definitions, maybe a custom setting in general.h?
`#define DISPLAY_ST7735_{RST,CS,DC} ...` 

Similar, for the ssd init, as it is silently using tx and rx  :/

code/espurna/thermostat.cpp
```diff
@@ -533,20 +573,27 @@ bool _temp_range_need_update = true;
 
 //------------------------------------------------------------------------------
 void drawIco(int16_t x, int16_t y, const char *ico, bool on = true) {
-  display.drawIco16x16(x, y, ico, !on);
+#if THERMOSTAT_DISPLAY_ST7735_SUPPORT
```
mcspr, line 580: What about the approach used in rfbridge - instead of splitting functions in half, just make a separate drawIco() that only works when ST7735 is enabled?
(same applies to all the things below)

code/espurna/Roboto_Thin9pt8b.h
```diff
@@ -0,0 +1,432 @@
+const uint8_t Roboto_Thin9pt8bBitmaps[] PROGMEM = {
```
mcspr: I think this file should go in `static/`

code/espurna/thermostat.cpp
```diff
@@ -533,20 +573,27 @@ bool _temp_range_need_update = true;
...
+  display.drawIco16x16(x, y, ico, !on); //FIXME
+#endif
   _display_need_refresh = true;
 }
 
 //------------------------------------------------------------------------------
 void display_wifi_status(bool on) {
   _display_wifi_status = on;
   drawIco(0, 0, wifi_on_bits, on);
+  drawIco(34, 0, server_bits, on);
```
mcspr: `display_server_status`?

code/espurna/thermostat.cpp
```diff
@@ -533,20 +573,27 @@ bool _temp_range_need_update = true;
...
 }
 
 //------------------------------------------------------------------------------
 void display_mqtt_status(bool on) {
   _display_mqtt_status = on;
   drawIco(17, 0, mqtt_bits, on);
+  drawIco(51, 0, remote_temp_bits, on);
```
mcspr: `display_remote_temp_status`?

code/espurna/thermostat.cpp
```diff
@@ -587,12 +662,28 @@ void display_remote_temp() {
...
+#if THERMOSTAT_DISPLAY_ST7735_SUPPORT
+  display.fillRect(0, 32, ST7735_TFTHEIGHT_160, 16, ST77XX_BLACK);
+  display.setTextColor(ST77XX_WHITE);
+  display.setTextWrap(true);
+  //display.setFont(&TomThumb);
+  display.setFont(&Roboto_Thin9pt8b);
+  
+  display.setCursor(0, 32 + 13);
+  String local_temp_title = String("Local      t");
+  display.print(local_temp_title);
```
mcspr: In small case like this, just `display.print(F("Local      t"))` would do

code/espurna/thermostat.cpp
dpeddi: There should me more customization... eg: rotation.. 
There are lot of different similar display that may need some different specifications...
Maybe better in platformio_override?

Will try with general...

code/espurna/Roboto_Thin9pt8b.h
dpeddi: Will move

code/espurna/main.cpp
dpeddi: Didn't noticed it

code/espurna/thermostat.cpp
dpeddi: Will try

code/espurna/thermostat.cpp
dpeddi: Oh sorry... Test code just for the photo :-)

code/espurna/thermostat.h
dpeddi: fixed

code/espurna/thermostat.h
dpeddi: should i move 
#include <SSD1306.h> // alias for `#include "SSD1306Wire.h"`
as well?

code/platformio.ini
dpeddi: Since the hardware is custom both with ssd1306 and in st7735 display...
What about adding some examples in platformio_override.ini?

code/espurna/thermostat.cpp
dpeddi: i hope to have understood correctly

code/espurna/thermostat.cpp
mcspr: general header should have things regardless, as reference / defaults. ifndef check allows us to override things from custom.h / platformio_override

code/platformio.ini
mcspr: Meaning:
- new `display_lib_deps = ...`
- example with `lib_deps = ${common.display_lib_deps} ${common.lib_deps}`, `src_build_flags = -DTHERMOSTAT_SUPPORT=1 -DDISPLAY_SUPPORT=1`

?

code/espurna/thermostat.cpp
mcspr: I meant, we don't need to create a temporary variable, just doing `print(F(...))` will automatically create a temporary `String()` object that will be used by the `print()` method.
While the original code did this exact thing, it's not really needed.

dpeddi: fixed, I think

code/platformio.ini
dpeddi: should be fixed...

dpeddi:
the original developer used ThingPulse/esp8266-oled-ssd1306, if someone who have ssd1306 would move to adafruit/Adafruit_SSD1306 probably there would be less ifdef.

Next step may be to create a display "class" to be referenced  from espurna itself for ip address/ssid and/or relay status..

code/espurna/thermostat.cpp
dpeddi: done as you suggested!

mcspr:
> the original developer used ThingPulse/esp8266-oled-ssd1306, if someone who have ssd1306 would move to adafruit/Adafruit_SSD1306 probably there would be less ifdef.

idk about that. perhaps, availability and / or esp8266 quirks were handled better originally, but still it feels like Adafruit option is mostly targeted at other Arduino boards.
also check out https://github.com/lexus2k/ssd1306, despite the name it handles both of these displays.

> Next step may be to create a display "class" to be referenced from espurna itself for ip address/ssid and/or relay status..

maybe. depends on what exactly relays will do. e.g. do we create sort-of 'window' / region and allow external thing to fully control it, or keep some specific spaces for 'info' and only allow data input

let me think about the example / pio.ini code a bit, I'll merge the PR after that

PR #2321, opened 2020-08-13
relay + api refactoring

mcspr:
- share relay dummypin instance between relay objects, replace unique_ptr with basic pointer (as we never destroy things, everything is brought up on boot and is essentially static)
- reduce overall size of the Api (web_api_t) structure, store a single required unsigned char id inside of the Api object itself.
- drop std::function for the same reason, current implementation only needs a single u8 ID in all cases
- tweak internal functions to expect a certain path size, drop strange comparisons with sprintf return value
- (kind of a hack) add manual calls to vector<Api>::reserve() as we go over the current capacity and vector increases it's size times 2. e.g. for 9 relays, we would allocate space for 32 Api objects as vector size goes from 16 to 32, after we add 18 Api objects with relay + pulse
- (breaking) json calls on a separate path, don't waste time encoding a single entity as json object when we can encode more things
- rfbridge API learn will return the current status instead of a plain OK

### DIFF
--- a/code/espurna/api.h
+++ b/code/espurna/api.h
@@ -22,14 +22,71 @@ String apiKey();
 
 #if WEB_SUPPORT && API_SUPPORT
 
-#include <functional>
+#include <vector>
 
-using api_get_callback_f = std::function<void(char * buffer, size_t size)>;
-using api_put_callback_f = std::function<void(const char * payload)> ;
+constexpr unsigned char ApiUnusedArg = 0u;
 
-void apiRegister(const String& key, api_get_callback_f getFn, api_put_callback_f putFn = nullptr);
+struct ApiBuffer {
+    constexpr static size_t size = API_BUFFER_SIZE;
+    char data[size];
+
+    void erase() {
+        std::fill(data, data + size, '\0');
+    }
+};
+
+struct Api {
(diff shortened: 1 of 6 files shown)


code/espurna/api.h
```diff
@@ -22,14 +22,71 @@ String apiKey();
...
+};
+
+struct Api {
+    using BasicHandler = void(*)(const Api& api, ApiBuffer& buffer);
+    using JsonHandler = void(*)(const Api& api, JsonObject& root);
+
+    enum class Type {
+        Basic,
+        Json
+    };
```
mcspr, line 45: Initial attempt also included Type::Indexed
This is just 'nice-to-have' atm, since it only benefits relays and light channels, as sensor magnitudes are all named differently most of the time. However, this should be an easy change:
- adding ApiMatch check before comparing paths, when Api object includes `#` (just like `mqttMagnitude()`)
- setting `Api::arg` to the value extracted from the path, thus using ApiUnusedArg placeholder.
- (or) 3rd function type that includes index arg instead of re-using `Api::arg`

However, now we need to check if the index is valid, because the index value could be anything. Api callback *could* return status, same as callbacks above it, whether the request was handled or not. Then, when failing, it can just fall down to the 404 case

PR #2322, opened 2020-08-14
Correct manufacturer on HASSIO discovery + build error for IR TX not in raw mode

alextircovnicu:
Fixed the device model and manufacturer in the discovery topic. 
Fixed build error for IR TX not in raw mode.

### DIFF
--- a/code/espurna/ir.cpp
+++ b/code/espurna/ir.cpp
@@ -247,7 +247,7 @@ void _irTXLoop() {
         #if IR_USE_RAW
             _ir_sender.sendRaw(_ir_raw, _ir_repeat_size, _ir_freq);
         #else
-            _ir_sender.send(_ir_type, _ir_code, _ir_bits);
+            _ir_sender.send((decode_type_t)_ir_type, _ir_code, _ir_bits);
         #endif
 
         // Update repeat count
--- a/code/espurna/ir.cpp
+++ b/code/espurna/ir.cpp
@@ -67,9 +67,9 @@ Raw messages:
         uint8_t _ir_repeat_size = 0;   // size of repeat array
         uint16_t * _ir_raw;            // array for sending codes and repeat codes
     #else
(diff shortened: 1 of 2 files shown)


code/espurna/ir.cpp
```diff
@@ -247,7 +247,7 @@ void _irTXLoop() {
         #if IR_USE_RAW
             _ir_sender.sendRaw(_ir_raw, _ir_repeat_size, _ir_freq);
         #else
-            _ir_sender.send(_ir_type, _ir_code, _ir_bits);
+            _ir_sender.send((decode_type_t)_ir_type, _ir_code, _ir_bits);
```
mcspr: Having noticed that... _ir_type should be `int`, since we deal with enum.
Or, just decode_type_t, idk why it is unsigned :/

alextircovnicu: I am not sure I understand you (my English is a little bit rusty :) ). 
The _IRsend::send_ method expects a _decode_type_t_ enum so __ir_type_ as _int_ I don't think is possible. 
To have it as _decode_type_t_  ... well I guess that would be the elegant solution. We then need a helper method to convert the value received from MQTT (line 211) to _decode_type_t_.

mcspr: Hm. So:
- `uint8_t _ir_type` -> `decode_type_t _ir_type`
- `_ir_type = static_cast<decode_type_t>(data.toInt());`
- `static_cast<int>(_ir_type)` for debug print

(I missed the fact that _ir_type never gets changed, I though the reason behind uint8_t it was it added / subtracted / masked somehow)

alextircovnicu: Can you "fix' this pull request ? I tried to edit it but instead of removing the second commit I managed to add a third one :( .

mcspr: Git is weird like that. Done. Thanks!

PR #2332, opened 2020-08-21
web: prometheus metrics support

### LINKED ISSUE #2323: Create a metrics page for prometheus scraping
**Is your feature request related to a problem? Please describe.**
I would like to be able to have prometheus scrape metrics of my IOT devices such as outlets' power usage.

**Describe the solution you'd like**
Present a page that can be enabled that would output all stats available for prometheus to scrape.

**Describe alternatives you've considered**
Right now I have a node server that aggregates all the api calls that are available for reading power, voltage, total kw since power on, etc and formats them into a prometheus scrape-able format.

Since I am already running the infrastructure for prometheus, I would like to avoid having another datasource that I have to manage (like influxdb).

**Additional context**
Example dashboards that can be built if metrics can be ingested easily:

image

mcspr:
resolve #2323 

Looks like:
```
relay0 0
relay1 0
relay2 0
relay3 0
temperature0 10.000
temperature1 9.000
humidity 35.000
pressure 997.000

```

I'll need to adjust sensor value printing, decimal places are hard-coded to 3 and need to be handled by the sensors module (somehow), as those methods aren't public right now.
<s>HTTP auth won't work, as it seems. I can't find any config option to accept user:pass pair for the metrics target.</s>

```yaml
scrape_configs:
  - job_name: 'espurna'
    metrics_path: '/api/metrics'
    params:
      apikey: ['apikeyapikey']
    static_configs:
      - targets: ['espurna-blabla.lan:80']
```


mcspr:
Using decimals from sensor / defaults:
```
relay0 0
temperature0 30.5
temperature1 29.5
humidity 56
pressure 1005.50
```
(i.e. for my DummySensor, temperature resolution is 0.1. Same for pressure, where it is 0.01. Humidity is a percentage, thus no decimals at all. Any specific sensor will override this in case value could fit more digits)

Path is `/api/metrics?apikey=...`, config example updated above.

PR #2335, opened 2020-08-23
rfb: rework both efm8bb1 & rcswitch implementations

mcspr:
- build flags RF_... are renamed to RFB_...
- drop RFB_DIRECT boolean in favor of RFB_PROVIDER=..., adds RFB_PROVIDER_EFM8BB1 and RFB_PROVIDER_RCSWITCH
- rework rfbridge parser. drop serial buffering and combined processing of rcswitch / rfb data in favor of state-full parsing
- rework rcswitch integration, support variable length payload via bitlength and allow to output 64bit data via rc-switch fork from open-mqtt-gateway.
- apply bit-length comparison for the old rcswitch codes, bump cfg version to 5 to allow 1.14.1 config updates (broken on dev, since cfg setting was never updated on clean installations 😞)
- optional relay support for both, independent of mqtt
- rework settings scanning routine when trying to match rf payload with the relay ID, scan settings only once instead of relayCount()<sup>2</sup> times if we want to find something for relayCount()-1'th relay
- add broker callbacks list with proto id + stringified code. default rfbridge does not support more than one protocol, so send out ID 0 as default (which matches with what 0xa6 will show, supposedly as it does not work for me :?)
- send out  0xa6 and 0xb0 sniffer data from https://github.com/Portisch/RF-Bridge-EFM8BB1/ to the MQTT and internally via broker callback
- update receiver callback 'locking' to make sure specific relay ID is tracked instead of all at once
- update build tests and deprecation notices

re @Sag06, please test :)
https://gitter.im/tinkerman-cat/espurna?at=5f400f3049148b41c963527e
CI here seem to have died for some reason, I need to fix that before nightlies are up again. But, I noticed you mentioned esphome, so you *should* have all required stuff already since we also use PlatformIO build tools.

Either build works
- rfbridge-direct env
- or, using esp8266-1m-base and [`ESPURNA_FLAGS="-DITEAD_SONOFF_RFBRIDGE -DRFB_PROVIDER=RFB_PROVIDER_RCSWITCH"`](https://github.com/xoseperez/espurna/wiki/PlatformIO)

Mind the size requirements


mcspr:
re: https://github.com/xoseperez/espurna/commit/82d5de0b64265585faaedc365354393373f0eab9

Merge vs. Squash 'issue' submitted as ticket to Github, at this point it is definitely some kind weird interaction with their servers.
A lot of useless commits in the history now :(

PR #2353, opened 2020-09-15
Hardware: Add support for Fcmila E27 7W RGB+W light bulb

user176176:
Device in question: https://www.aliexpress.com/item/32925895199.html
Named after print on the PCB.


mcspr:
Thanks!

PR #2355, opened 2020-09-16
Add E1.31 Support

CalcProgrammer1:
Implements changes requested in #1922 

Adds configurable E1.31 support for controlling Espurna lighting devices with E1.31 lighting software such as Vixen 3, OpenRGB, etc.  Web-based configuration page provides the ability to enable E1.31 support globally and enable E1.31 control for each lighting channel.  You can set the universe for the device and the E1.31 channel for each Espurna lighting channel.  Can be disabled with E131_SUPPORT flag.  Requires light support.

I have addressed the comments from my first push of this branch.


code/espurna/e131.cpp
```diff
@@ -0,0 +1,161 @@
...
+    _e131_light_0_enabled       = getSetting("e131Light0Enabled", false);
+    _e131_light_0_channel       = getSetting("e131Light0Channel", 1);
+    _e131_light_1_enabled       = getSetting("e131Light1Enabled", false);
+    _e131_light_1_channel       = getSetting("e131Light1Channel", 2);
+    _e131_light_2_enabled       = getSetting("e131Light2Enabled", false);
+    _e131_light_2_channel       = getSetting("e131Light2Channel", 3);
+    _e131_light_3_enabled       = getSetting("e131Light3Enabled", false);
+    _e131_light_3_channel       = getSetting("e131Light3Channel", 4);
+    _e131_light_4_enabled       = getSetting("e131Light4Enabled", false);
+    _e131_light_4_channel       = getSetting("e131Light4Channel", 5);
```
mcspr, line 159: Note of `lightChannels()`. Right now configuration... seems straightforward, but bulky :/
As noticed in the other comment, my92xx uses simple string to map as `1,2,3,4,5` to map espurna channel 0 to 1, channel 1 to 2 and etc. We also won't need 10 config keys at the same time. (and we need a bit different parser, as there are more channels in e131 that go above a single digit)

'no mapping' is another question?
- `1,,3,4,5`
- `1,-,-,4,5`
- `0:1,2:2` (left of ':' is our channel, right is e131)

code/espurna/e131.cpp
```diff
@@ -0,0 +1,161 @@
...
+    root["e131Light3Enabled"]   = getSetting("e131Light3Enabled", false);
+    root["e131Light3Channel"]   = getSetting("e131Light3Channel", 4);
+    root["e131Light4Enabled"]   = getSetting("e131Light4Enabled", false);
+    root["e131Light4Channel"]   = getSetting("e131Light4Channel", 5);
+}
+
+#endif // WEB_SUPPORT
+
+void _e131Loop() {
+    if (_e131_enabled != 1) return;
```
mcspr: *nit, since this is not C*
`if (!_e131_enabled) return;`

code/espurna/e131.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    //* Initializing multicast mode must be done when the WiFi is connected, so
+    //* set a flag to track when WiFi is connected and disconnected
+    if (WiFi.status() == WL_CONNECTED) {
```
mcspr: Check out jw.subscribe callbacks, there is `MESSAGE_CONNECTED` and `MESSAGE_DISCONNECTED` for station events. Both happen in `loop()`, see wifi.cpp for example.

code/espurna/config/general.h
```diff
@@ -1388,6 +1388,14 @@
 #define DOMOTICZ_OUT_TOPIC      "domoticz/out"  // Default publication topic
 #endif
 
+// -----------------------------------------------------------------------------
+// E1.31
+// -----------------------------------------------------------------------------
+
+#ifndef E131_SUPPORT
+#define E131_SUPPORT            1               // Enable E1.31 by default
```
mcspr, line 1396: What is the .bin size difference between `0` and `1`?

code/espurna/e131.cpp
```diff
@@ -0,0 +1,161 @@
...
+    root["e131Light0Enabled"]   = getSetting("e131Light0Enabled", false);
+    root["e131Light0Channel"]   = getSetting("e131Light0Channel", 1);
+    root["e131Light1Enabled"]   = getSetting("e131Light1Enabled", false);
+    root["e131Light1Channel"]   = getSetting("e131Light1Channel", 2);
+    root["e131Light2Enabled"]   = getSetting("e131Light2Enabled", false);
+    root["e131Light2Channel"]   = getSetting("e131Light2Channel", 3);
+    root["e131Light3Enabled"]   = getSetting("e131Light3Enabled", false);
+    root["e131Light3Channel"]   = getSetting("e131Light3Channel", 4);
+    root["e131Light4Enabled"]   = getSetting("e131Light4Enabled", false);
+    root["e131Light4Channel"]   = getSetting("e131Light4Channel", 5);
```
mcspr, line 73: *(again about the cfg)*
We also will send less keys to the WebUI.

But, it will be more complicated from the .js script side :/
Another alternative to map string is to use `e131Ch#` and simply loop all channels like Domoticz, Thingspeak map internal IDs to external.

code/espurna/e131.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+        lightUpdate(false, false, false);
+    }
+}
+
+bool e131Enabled() {
+    return _e131_enabled;
+}
+
+void e131Setup() {
```
mcspr, line 145: Should the config be purely though settings / UI, no use for build flags to pre-configure the defauls?

mcspr (commented):
Can you point to any simple client to test this out?

CalcProgrammer1:
> 
> 
> Can you point to any simple client to test this out?

My OpenRGB project can control E1.31.  You can get the latest release on GitLab:

https://gitlab.com/CalcProgrammer1/OpenRGB

You will need to create an e131.txt file in the same folder as the OpenRGB application.  That file should contain something like:

```
e131_device_start
Espurna Bulb 1
num_leds=1
start_channel=1
start_universe=1
rgb_order=RGB
type=SINGLE
e131_device_end

e131_device_start
Espurna Bulb 2
num_leds=1
start_channel=4
start_universe=1
rgb_order=RGB
type=SINGLE
e131_device_end
```

This is the configuration I was testing with, two Espurna bulbs on universe 1.  The first bulb was configured with channels 1/2/3, the second with channels 4/5/6.

mcspr:
Thanks! Up and running.

I'll check out if Core's generic UDP server might work, too, since the only thing we want is to receive that giant struct and cast it. Might save a bit of .bin size, since mDNS already uses it.

code/espurna/config/general.h
mcspr: Answering myself, 5232 bytes

`-DE131_SUPPORT=1`
595360
`-DE131_SUPPORT=0`
590128

code/espurna/e131.cpp
CalcProgrammer1: Would it be possible to keep the same web UI and pack the settings from it into a string?  I basically just copy pasted the web functionality from other pages as I'm not really sure how it works, nor am I well versed in javascript.

mcspr: Right now this sends key-values as-is, so every key is 'mapped' to the same 'input name=...' elem, this is done at the end of `processData()` func of custom.js

What I meant about `e131Ch#` approach is to do something like:
```cpp
String base("e131Ch");
for (unsigned index = 0; index < lightChannels(); ++index) {
    String key(base + index);
    root[key] = getSetting(key); // without 'default' 2nd param, always returns String
}
```
Treating empty string as disabled and non-empty as e1.31 channel number, while always creating N-channels items. Same thing in the configuration.

Then, we either create html block programmatically like dczRelayTemplate, or always have html blocks in the index.html but hide them by default via `display: none` css style and add additional js code such as:
```js
// processData() near the top...

if (key.startsWith("e131Ch")) {
    $("input[name=" + key + "]").show();
}

// ...rest of processData()
```

code/espurna/e131.cpp
CalcProgrammer1: Updated to use a WiFi callback to get connection status

code/espurna/e131.cpp
CalcProgrammer1: Updated

PR #2361, opened 2020-09-24
BME680 fixes

mcspr:
- make sure we call run() before checking for errors, we will miss status update otherwise and lock the sensor class in the error state that it cannot recover from
- set error flag only when either bsec or bme680 report an error. ensure we show any issues reported by the library via the debug log
(as a theoretical TODO - add custom error messages through virtual method?)
- minor fix to the sensor.h when using 1m,2m boards


ruimarinho:
Nice one @mcspr 👍

PR #2367, opened 2020-09-25
add missing lights #include for rpn rules

ngilles:
When building a light with rpn rules support build fails because lightUpdate() etc can't be found. Looks like #include "light.h" was missing...


mcspr:
Thanks!

Cross-dependencies are weird :/ perhaps one of code/test/build/light\_...h can include rpn_rules_support=1, but this is getting ridiculous... 
*edit:* or, the rpn module itself could skip the ifndef in the .cpp at a cost of a slightly longer build

PR #2370, opened 2020-09-27
ota: detect Tasmota magic numbers when booting

mcspr:
Some magic to allow seamless Tasmota OTA upgrades
- inject dummy data sequence that is expected to hold current version info
- check rtcmem markers and purge settings, since we don't want accidentally reading something as a kv
- sometimes we cannot boot b/c of certain SDK params, purge last 16KiB as well

ref. `SetOption78 1`
https://tasmota.github.io/docs/Commands/#setoptions

The original idea from Esphome:
https://github.com/esphome/esphome/blob/0e59243b83913fc724d0229514a84b6ea14717cc/esphome/core/esphal.cpp#L275-L287
Does not seem to be working atm. Will need to re-check with the esphome builder, c/p did not work and I tried building with gcc-10

Specific things we want to do is to have 'version' byte between 2 'marker' bytes:
https://github.com/arendst/Tasmota/blob/217addc2bb2cf46e7633c93e87954b245cb96556/tasmota/settings.ino#L218-L262       https://github.com/arendst/Tasmota/blob/0dfa38df89c8f2a1e582d53d79243881645be0b8/tasmota/i18n.h#L780-L782

---

Hopefully settings reset helps reliably and does not break things, since I crash 99% of the time after doing:
```
backlog ssid1 SSID; password1 PASSWORD
```
And OTA'ing the binary. Right after softAP is brought up, everything halts and device resets after some time
*(might be interesting to check with the https://github.com/esp8266/Arduino/pull/7010, since this is a pretty reliable source of WDTs...)*

### DIFF
--- a/code/espurna/main.cpp
+++ b/code/espurna/main.cpp
@@ -153,14 +153,10 @@ void setup() {
     info(true);
 
     wifiSetup();
-    #if OTA_ARDUINOOTA_SUPPORT
-        arduinoOtaSetup();
-    #endif
+    otaSetup();
+
     #if TELNET_SUPPORT
         telnetSetup();
-    #endif
-    #if OTA_CLIENT != OTA_CLIENT_NONE
-        otaClientSetup();
     #endif
 
     // -------------------------------------------------------------------------
--- a/code/espurna/ota.cpp
+++ b/code/espurna/ota.cpp
@@ -7,7 +7,11 @@ OTA MODULE COMMON FUNCTIONS
 #include "ota.h"
 #include "system.h"
 #include "terminal.h"
+#include "rtcmem.h"
+#include "utils.h"
 #include "ws.h"
+
+#include <atomic>
 
 void otaPrintError() {
     if (Update.hasError()) {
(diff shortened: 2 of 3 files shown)


mcspr:
@alextircovnicu btw I had noticed the sp111 wiki page mentioning setoption78 for tasmota

could you check out these changes *without* doing it? OTA now should be able to mimic tasmota .bin and also will reset any existing flash settings when encountered

nvm Codacy checks here, it is expected to 'soft-crash' after Tasmota OTA and normally reboot after erasing any persistent settings

alextircovnicu:
> @alextircovnicu btw I had noticed the sp111 wiki page mentioning setoption78 for tasmota
> 
> could you check out these changes _without_ doing it? OTA now should be able to mimic tasmota .bin and also will reset any existing flash settings when encountered
> 
> nvm Codacy checks here, it is expected to 'soft-crash' after Tasmota OTA and normally reboot after erasing any persistent settings

Sure thing. Tomorrow I will get some new SP111 sockets (for the dishwasher and coffee filter :) ) and I will test it. I don't want to test with the ones that I already flashed - they are now embedded in the bathroom furniture and it is hard for me to get them in case physical action is needed.

alextircovnicu:
@mcspr, I confirm, this works like a charm. You can now switch from TASMOTA to ESPURNA without setting option 78 to 1.

PR #2379, opened 2020-10-12
btn: generate pressed & released events from switch changes

### LINKED ISSUE #2377: Switch 'Released' and 'Pressed' events
Dear All,
When I press a button, I have multiple "PRESSED" events and when I release it, I have nothing.
How can I do to get one event "PRESSED" and one event "RELEASED" instead ?

My hardware config is :
**#define BUTTON1_PRESS         BUTTON_ACTION_ON
#define BUTTON1_RELEASED      BUTTON_ACTION_OFF**

When I change in button.cpp  :
**case debounce_event::types::EventReleased: {
            // return _buttonMapReleased(
            //     event_emitter->getEventCount(),
            //     event_emitter->getEventLength(),
            //     event_delays.lngclick,
            //     event_delays.lnglngclick
            // );
            return button_event_t::Released;
        }**

I have alternatively "PRESSED" and "RELEASED" events.

Best Regards

mcspr:
- Add ::Released button event to be used specifically with switches. It is only generated when there are no click actions configured
- Always use default state for switches instead of ignoring it and using current pin reading as value. This allows us to boot with the switch held in the opposite state and trigger the correct event immediately
(meaning, switches now depend on the BUTTON_DEFAULT_HIGH presence)

Configuration adds BUTTONX_RELEASE build flag and btnRlseX action settings

@ksirob35
I don't think I broke things, but will need to re-check before commiting this proper


mcspr:
Alternative option - generate 'pressed' and 'released' from debounceevent directly, avoiding 'changed' altogether. With the same logic about the 'click', when there's no action there it will fallback to 'pressed' and 'released' actions

*edit:* which actually seems like a better option. Following change https://github.com/xoseperez/espurna/pull/2379/commits/cacc05db3b311b5d409b128996ee301817aea2c6 <s>stolen</s> made from https://github.com/a-tom-s/debounceevent/commit/911020fa58e0e28a4ea7f3336c1563c2ce95b448 and #571

mcspr:
Also, note-to-self: recover click event, we still want the wall-switch behaviour (and because switches did that for a long time, without any changes to the code)

Abstraction stuff gets a little weird here, since both changes do the opposite of that. Default state may be ignored akin to AceButton's approach, by having three-state 'status' as unknown on boot. Perhaps, as an additional bit for the button config to read the current pin value (PinValue::Current?).

sheinz:
Hi!
This PR brakes sonoff mini external switch function.
In the logs there are 'pressed' and 'released' events but no relay action.
Probably switch needs different configuration but with the default one it does not work after this PR.

sheinz:
After adding configuration for the switch:
```
    #define BUTTON2_PRESS       BUTTON_ACTION_ON
    #define BUTTON2_RELEASE     BUTTON_ACTION_OFF
```
Sonoff mini external switch works.
I actually like the new behavior more.
Thanks!

mcspr:
@sheinz tnx! will add in PR adding 'boot' default state

PR #2387, opened 2020-10-25
Add sleep support for Nova Fitness SDS011 sensor

ruimarinho:
The Nova Fitness SDS011 sensor has a limited lifespan time of 8000 hours, or roughly 1 year of continuous operation. This lifespan can be largely extended if the sensor is configured to sleep for a certain amount of time (the so called 'working cycles'), wake up to take a measurement, and then going back to sleep again.

This PR changes the default behaviour of this integration to benefit from these working cycles as users might inadvertently be 'killing' their sensors by an overly aggressive continuous reading strategy.

### DIFF
--- a/code/espurna/config/sensors.h
+++ b/code/espurna/config/sensors.h
@@ -1001,6 +1001,19 @@
 #define SDS011_TX_PIN                    12
 #endif
 
+#ifndef SDS011_REPORTING_MODE
+#define SDS011_REPORTING_MODE            1   // Mode 0 = query, 1 = active. Default mode is 'active' where data is
+#endif                                       // read from software serial instead of querying the sensor.
+
+#ifndef SDS011_CUSTOM_WORKING_PERIOD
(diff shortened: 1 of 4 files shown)


mcspr:
Nice! Defaults are indeed rough
(but it is kind-of possible that the sensor can be pre-configured using PC software tool? since the settings persist based on the .pdf info)

I wonder though, would it be enough to just add small func doing the serial writing + reading response by ourselves instead of using the lib? ref. https://cdn.sparkfun.com/assets/parts/1/2/2/7/5/Laser_Dust_Sensor_Control_Protocol_V1.3.pdf in the `3 Command, 5) Set working period`
Payload format from our side is always fixed, response seems to be too. Although, query reading mode is an interesting addition to be based off of the sensor read interval time

Also, don't we want the setting to be using minutes directly? Regardless of the implementation, library or custom code. Sensor does not have sub-minute setting anyway, just another possible bug when we will try to use numbers like 300 in uint8_t and overflow things by accident

Issue #1663, opened 2019-03-28
Migrating to Sonoff Tosmato bicks devices


dominikandreas:
**Bug description**
Using OTA to migrate to a Sonoff Tasmota minimal firmware bricks device

**Steps to reproduce**
Upload sonoff tasmota firmare (tested using web-interface, telnet, with and without prior firmware reset).

**Expected behavior**
Not bricking the device

**Device information**
```
[877379] [MAIN] ESPURNA 1.13.4 (2aa5c59)
[877380] [MAIN] xose.perez@gmail.com
[877381] [MAIN] http://tinkerman.cat

[877382] [MAIN] CPU chip ID: 0x49D786
[877385] [MAIN] CPU frequency: 80 MHz
[877387] [MAIN] SDK version: 1.5.3(aec24ac9)
[877392] [MAIN] Core version: 2.3.0
[877394] [MAIN] Core revision: 159542381
[877399] 
[877400] [MAIN] Flash chip ID: 0x1440A1
[877403] [MAIN] Flash speed: 40000000 Hz
[877408] [MAIN] Flash mode: DOUT
[877410] 
[877411] [MAIN] Flash size (CHIP)   :  1048576 bytes /  256 sectors (   0 to  255)
[877420] [MAIN] Flash size (SDK)    :  1048576 bytes /  256 sectors (   0 to  255)
[877424] [MAIN] Reserved            :     4096 bytes /    1 sectors (   0 to    0)
[877433] [MAIN] Firmware size       :   491520 bytes /  120 sectors (   1 to  120)
[877440] [MAIN] Max OTA size        :   532480 bytes /  130 sectors ( 121 to  250)
[877447] [MAIN] EEPROM size         :     4096 bytes /    1 sectors ( 251 to  251)
[877456] [MAIN] Reserved            :    16384 bytes /    4 sectors ( 252 to  255)
[877462] 
[877463] [MAIN] EEPROM sectors: 251, 250
[877465] [MAIN] EEPROM current: 250
[877468] 
[877470] [MAIN] EEPROM:  4096 bytes initially |  1523 bytes used (37%) |  2573 bytes free (62%)
[877479] [MAIN] Heap  : 35088 bytes initially | 19584 bytes used (55%) | 15504 bytes free (44%)
[877487] [MAIN] Stack :  4096 bytes initially |  1792 bytes used (43%) |  2304 bytes free (56%)
[877495] 
[877497] [MAIN] Boot version: 4
[...]

PR #2391, opened 2020-11-01
hardware: add support for Gosund P1 Power Strip

alextircovnicu:


mcspr:
Thanks!

I was wondering about the Tasmota implementation doing multiple analogRead calls and <s>taking an average</s> shifting the sum N times:
https://github.com/arendst/Tasmota/blob/3b432a58a3f0de1c1affedf9f894c2f0f4e1b8e1/tasmota/xsns_02_analog.ino#L234-L248
Buttons in the interface do it twice, rules example from the [templates page](https://templates.blakadder.com/gosund_P1.html) does it 32 times every 0.25seconds. If what we do right now works, might not be a problem :) My guess the precision might be slightly off, sometimes (or never, hopefully).

(*and, technically, Tasmota implementation is also missing out on the SDK function doing this specific thing... system_adc_read_fast, but it might not be really suited for using wifi at the same time*)

alextircovnicu:
Sorry for being so late - you know, it is hard with two kids abroad while Europe is more or less in lock-down. 
So, to the topic: Like I previously mentioned I had Tasmota running for a week, configured as explained on the [template page](https://templates.blakadder.com/gosund_P1.html).
Every 10 buttons presses it was behaving odd: turning ON two relays at once, turning ON  another relay while I was turning off one. Also the button click it was not natural. I had to keep the button pressed more than a usual "button click" . Seeing now the implementation I guess all these behaviors make sense. 

I think your implementation is the way to go here. Yes, for analog readings, I guess, getting more samples and shifting the sum may be a more logical approach. For this resistor ladder, you already know the expected values (which are about 20 ohms smaller than Tasmota uses). I have activated the analog sensor, to see the button values and the readings are pretty constant (never noticed a deviation bigger than 7 ohms).  

I will update the Wiki in the upcoming days.

P.S. Tuya convert never worked for me with this Gosund devices and, which I think it is a little bit strange, the Tasmota minimal firmware (uploaded after espurna) always bricked my devices - I had use the uart programmer again on them.

mcspr:
I meant to point out the buttons are seem to be configured similar to ours, through the GPIO config as analog button, although it is only available in the not-yet-released Tasmota 9.x version. So the script is kind of deprecated.
But, I have no idea what to do with the adcparam or whatever is used to configure them :)
About the minimal fw... that will probably never work, if this issue still applies:
https://github.com/xoseperez/espurna/issues/1663

Analog sensor does more than one analogRead btw, ref. ANALOG_SAMPLES defaulting to 10 times every ~1ms which adds each value then divides. If it is ok for ladder though, this can be just one read (meaning... two, including the debounce check that the value is still in range)

Stay safe! No worries.

PR #2394, opened 2020-11-05
Merge with dev & update to a docker build environment

wyattearp:
I was hoping for a docker build and it looks like it never got moved in. All credit to @copyrights for their work, I'm just coming by to try and get things wired back into place and current with `dev`

This should allow you to create a docker image like in #1754, which seems to have gone stale and I can't see why the code quality doesn't pass (hopefully this PR will tell me and I can fix it up).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,14 +2,19 @@ FROM debian:buster-slim
 
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update && apt-get upgrade -y && \
-    apt-get install -y python2-minimal \
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+    python2-minimal \
     python-pip \
+    python-setuptools \
+    ca-certificates \
     nodejs \
     npm \
     git \
     libc-bin \
     bash \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
     && npm install -g npm \
     && pip install platformio
 
@@ -27,15 +32,14 @@ RUN mkdir -p /espurna && \
 
 USER worker
 
-RUN git clone -b dev https://github.com/xoseperez/espurna.git /espurna && \
-    cd /espurna/code && \
-    npm install --only=dev && \
+RUN git clone -b dev https://github.com/xoseperez/espurna.git /espurna
+
+WORKDIR /espurna/code
+
+RUN npm install --only=dev && \
     platformio run --target clean
 
 VOLUME ["/espurna", "/firmware"]
 
-WORKDIR /espurna/code
-
-COPY docker-entrypoint.sh /
 ENTRYPOINT ["./build.sh", "-d", "/firmware"]
 CMD []
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,9 +15,12 @@ RUN apt-get update \
     && npm install -g npm \
     && pip install "platformio>=3.6,<3.7"
 
-RUN groupadd --gid 1000 worker && \
+ARG UID=1000
+ARG GID=1000
+
+RUN groupadd --gid $GID worker && \
     useradd \
-    --uid 1000 \
+    --uid $UID \
     --gid worker \	
     --shell /bin/bash \
     --create-home worker
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,17 +4,16 @@ ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-    python2-minimal=2.7.16-1 \
-    python-pip=18.1-5 \
-    python-setuptools=40.8.0-1 \
-    ca-certificates=20190110 \
-    nodejs=10.15.2~dfsg-2 \
-    npm=5.8.0+ds6-4 \
-    git=1:2.20.1-2 \
-    && apt-get clean \
+    python2-minimal \
+    python-pip \
+    python-setuptools \
+    ca-certificates \
+    nodejs \
+    npm \
+    git \
     && rm -rf /var/lib/apt/lists/* \
     && npm install -g npm \
-    && pip install platformio==3.6.7
+    && pip install "platformio>=3.6,<3.7"
 
 RUN groupadd --gid 1000 worker && \
     useradd \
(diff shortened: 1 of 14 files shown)


docker/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+```bash
+sudo docker run --rm -it -v /tmp/firmware/:/firmware espurna-build
+```
+
+### Building A Single Environment
+This example will only build firmware for environment `intermittech-quinled` from local files in `/home/user/espurna`
+
+```bash
+udo docker run --rm -it -v /tmp/firmware/:/firmware espurna-build intermittech-quinled
```
mcspr:
```suggestion
sudo docker run --rm -it -v /tmp/firmware/:/firmware espurna-build intermittech-quinled
```

docker/README.md
```diff
@@ -0,0 +1,61 @@
...
+Building for version 1.15.0-dev (83a49b19)
+--------------------------------------------------------------
+Building web interface...
+[18:38:22] Using gulpfile /espurna/code/gulpfile.js
+[18:38:22] Starting 'default'...
+[18:38:22] Starting 'webui'...
+...
+...
+Cloning into '/root/.platformio/.cache/tmp/pkg-installing-rwukc302'...
+Cloning into '/root/.platformio/.cache/tmp/pkg-installing-20osjkr2'...
```
mcspr, line 56: I think previous version meant to use userid map here?

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    nodejs \
+    npm \
+    git 
+
+# NOTE: there is currently a compability warning with the target versions of Node vs. NPM - using latest until something is identified as the target version
+RUN rm -rf /var/lib/apt/lists/* \
+    && npm install npm@latest -g
+
+# NOTE: Python module for pip requires upgrade adhead of PIO install to work around defects
+RUN python3 -m pip install -U pip
```
mcspr, line 21: *Nit / rant / OT (?)*

Perhaps I was living in a bubble so far, since I used very little of debian-like stuff outside of desktop / real servers. I guess this works for now, but I just tried this one:
```dockerfile
FROM fedora:latest
RUN dnf -y install nodejs python python-pip git && dnf clean all
RUN pip install platformio
```
Since we want bleeding-edge sw anyways... Feels like a waste doing all of this with stable debian :)
(...and gyp also pulls python2, we need to wait until march-april for the next release including it, or pull from testing repos... I wonder if things like [pyenv](https://github.com/pyenv/pyenv) + [nodeenv](https://github.com/ekalinin/nodeenv) would do a better job instead)

mcspr (approved):

docker/README.md
wyattearp: resolved

docker/README.md
wyattearp: previously, but there was questions on if everyone would have a valid userid that started at 1000 or similar, figured it was just easier to drop it than deal with it as most times people just sudo the docker anyway (for good or bad).

docker/Dockerfile
wyattearp: doesn't really matter to me - just happened to be what was there and it looked like the repo needed more than just the python or node images.

i will say that i'd still recommend explicitness to avoid any random breakage (python3 vs. python :) )

docker/README.md
mcspr: If this is just an example. Build sends things as root already, so the file needs some tweaking too, i.e. docker run --id is not enough.

docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    mkdir -p /firmware 
+
+# Copy in source code 
+COPY . /espurna
+
+# update our working director so that ./build.sh will execute correctly
+WORKDIR /espurna/code
+
+RUN npm install --only=dev && \
+    platformio run --target clean
```
mcspr, line 37: I missed this part. We also want the libs installed here? Check out the scripts/pio_pre.py, ESPURNA_PIO_SHARED_LIBRARIES env var and the platformio.ini config entry. Since PIO4 it does install libraries per-environment, so you end up with .pio/libdeps/sonoff-basic, .pio/libdeps/sonoff-pow and etc. However, since docker makes a snapshot of the espurna version and lib_deps is shared between envs, it would make sense to place libs there.

This can be a separate target for the platformio, check out https://docs.platformio.org/en/latest/projectconf/advanced_scripting.html#custom-targets
(both platform tools and libraries in the same step)

It's not on by default though (and I'll need to remember why not...) and needs that env flag to install libs there.

wyattearp: Took me a bit of digging on this to try and understand what you're looking for and I'm still not certain I understand what change your asking for here (probably because I'm not as familiar with the build environment setup). I think it's as simple as calling a `platformio lib install` ... but that does significantly increase the build time of the docker image, especially if you really only wanted it to build one target so you didn't have to setup a ton of environments (a couple of minutes to 20+ minutes). My original use case for this was far more for the person that had 1 target to build as opposed to **all** of the targets to build, given that if your run with a single target, it still installs the associated libraries that you need to generate a firmware image.

docker/README.md
wyattearp: I don't know that I understand - are you wanting to wire back in the previous ID stuff that wasn't working / might not be compatible or you are looking for something else?

docker/Dockerfile
mcspr: Not quite, the idea is that libs would be fetched just once into this dir:
https://github.com/xoseperez/espurna/blob/ac6f0b71e63206562acfdb8cec8e8e0331cb1f79/code/platformio.ini#L169-L170
Historically, PIO used a common library directory as .pio/libdeps/ until PIO4, which is what this essentially does try to simulate. No need to call each $env, just *something* that will fetch the libraries + build tools once.

docker/README.md
mcspr: Previous ID was fixed, true, I'm just saying that in case docker run --id=$(id -u) is used, it will fetch build tools all over again, since PIO installs stuff in the $HOME/.platformio

Maybe we want a comment in the readme with --build-arg UID=$(id -u) --build-arg GID=$(id -g)?
https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg

mcspr:
I think something broke in the merge :/ It should not list changes from dev

wyattearp:
abandoning pr

Issue #2397, opened 2020-11-13
ci: travis-ci.org stops working on December 31st


mcspr:
*subj*

https://docs.travis-ci.com/user/migrate/open-source-repository-migration

> **Q. When will the migration from travis-ci.org to travis-ci.com be completed? #**
A. In an effort to ensure that all of our users - whether you build open-source, public or private repositories - receive regular feature updates, security patches and UX/UI enhancements, we are announcing that travis-ci.org will be officially closed down completely no later than December 31st, 2020, allowing us to focus all our efforts on bringing new features and fixes to travis-ci.com and all of our awesome users like yourself on the travis-ci.com domain.

> **Q. What will happen to travis-ci.org after December 31st, 2020? #**
A. Travis-ci.org will be switched to a read-only platform, allowing you to see your jobs build history from all repositories previously connected to travis-ci.org.

https://github.com/mcspr/espurna-nightly-builder was using the .com version since the announcement, but...
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
(and whatever comes next)

> For those of you who have been building on public repositories (on travis-ci.com, with no paid subscription), we will upgrade you to our trial (free) plan with a 10K credit allotment (which allows around 1000 minutes in a Linux environment).
> ...
> Usage will be tracked under your account information so that you can better understand how many credits/minutes are being used

We don't depend on either macOS builds or massively-concurrent jobs. But, the current interface does not (yet) tell any stats related to the minutes, used or available.

Another alternative (and pretty viable one) is to just use the Github Actions, which seems to be pretty stable these days.

@xoseperez (will also send a dup to the email)

mcspr:
> ...does not (yet) tell any stats related to the minutes, used or available.

Not really true. Insights page -> https://travis-ci.com/github/mcspr?tab=insights

> OCTOBER 17, 2020 - NOVEMBER 15, 2020
TOTAL JOB MINUTES
252 mins

That's 3 nightly builds, 2 ~93minutes and 1 ~34minutes. Plus, 1 minute per daily invocation of the cron. IDK if that's viable anymore (but, I do have some ideas about how to speed up builds even more)

And I did not even do anything, really :>

mcspr:
Running via https://github.com/xoseperez/espurna/commit/f11468abba4ac9c0ef987524593b4d8bb85558e8
build.sh still references TRAVIS_... env though, but those seem no-op. Will need to fix as well

No release config yet

mcspr:
[...]

PR #2405, opened 2020-12-05
api: rework plain and JSON implementations

mcspr:
- match paths through a custom AsyncWebHandler instead of using generic not-found fallback handler
- allow MQTT-like patterns when registering paths (`simple/path`, `path/+/something`, `path/#`)
Replaces `relay/0`, `relay/1` etc. with `relay/+`. Magnitudes are plain paths, but using `/+` in case there's more than 1 magnitude of the same type.
- restore `std::function` as callback container (no more single-byte arg nonsense). Still, limit to 1 type per handler type
- adds JSON handlers which will receive JsonObject root as both input and output. Same logic as plain - GET returns resource data, PUT updates it.
- breaking change to `apiAuthenticate(request)`, it no longer will do `request->send(403)` and expect this to be handled externally.
- allow `Api-Key` header containing the key, works for both GET & PUT plain requests. The only way to set apikey for JSON.
- add `ApiRequest::param` to retrieve both GET and PUT params (aka args), remove ApiBuffer
- remove `API_BUFFER_SIZE`. Allow custom form-data key=value pairs for requests, allow to send basic `String`.
- add `API_JSON_BUFFER_SIZE` for the JSON buffer (both input and output)
- `/apis` replaced with `/api/list`, no longer uses custom handler and is an `apiRegister` callback
- `/api/rpc` custom handler replaced with an `apiRegister` callback

WIP further down:
- no more `webLog` for API requests, unless `webAccessLog` / `WEB_ACCESS_LOG` is set to `1`. This also needs to happen to the other handlers. 
- migrate to ArduinoJson v6, since it become apparent it is actually a good upgrade :)
- actually make use of JSON endpoints more, right now it's just existing GET for sensors and relays
- fork ESPAsyncWebServer to cleanup path parsing and temporary objects attached to the request (also, fix things a lot of things based on PRs there...)


mcspr:
This took longer than expected and became a much larger patch as a result :>
travis-ci.org checks are very slow... (ref. #2397)

RAM usage is almost the same, and it also avoids the std::vector reserved space.
Magnitudes... in theory, those could be `/api/sensor/+` (since `/api/+` would match every other handler), but I'd really like `/api/list` to actually work.

PR #2408, opened 2020-12-12
Introduce Garland module 

ElderJoy:
Port of https://github.com/Vasil-Pahomov/ArWs2812 from Arduino to ESP8266
Implementing garland of WS2812
Features for now:
- 8 animation modes
- Web control of ON/OFF and brightness
- env:garland setup

In plans:
- add more animations
- add more controls to web

It is quite raw but works for me on wemos-d1mini. Hope with community help it can be useful. New Year is coming... :)


ElderJoy:
Check failed on env:esp8266-4m-latest-base. Checked locally, build for env:esp8266-4m-latest-base failed with error:
**espurna\config/hardware.h:5078:6: error: #error "UNSUPPORTED HARDWARE!!"**
Does anybody know if it is my fault? What's wrong and what can I do to fix this issue?

code/espurna/garland/scene.h
```diff
@@ -0,0 +1,148 @@
+#ifndef anim_h
+#define anim_h
+#include <Adafruit_NeoPixel.h>
+#include "palette.h"
+#include "..\config\general.h"
```
mcspr: This one?
https://travis-ci.org/github/xoseperez/espurna/jobs/749181452#L315

You need to use `/`-slashes for paths.

mcspr:
Also note it merges every .cpp file into one when building in CI. `static` / `namespace { ... }` might still have name collisions. I also noticed end-of-file line-endings are missing for some of the .cpp files.

Do you think it is possible to move Anim code into a library? i.e. you can just use a separate repo ElderJoy/GarlandESP8266 or something like that

mcspr:
> Checked locally, build for env:esp8266-4m-latest-base failed with error:
espurna\config/hardware.h:5078:6: error: #error "UNSUPPORTED HARDWARE!!"

...I missed the question about local builds. You can check out the travis log above the error, since it's just a shell script and it shows every step it does. The -base environments are mission DEVICE and MANUFACTURER pair of build flags *somehow*, either custom.h, src_build_flags of the custom env, ESPURNA_FLAGS=... environ, PLATFORMIO_SRC_BUILD_FLAGS=... environ, etc.

You can try adding test/build/garland.h with the GARLAND_SUPPORT=1 (+ any additional flags), so the CI can try to build it

ElderJoy:
Hi @mcspr,
Thank you for pointing out my mistakes. I fixed slashes, add end-of-file where it was missed and wrap code that belongs to garland into macro to exclude it from build in other configurations.

As for separate garland-dedicated code into separate library. I supposed this functionality as extendable. Anyone can easily add and modify animations. Hope with community help we will have much more animations soon.
It would be not so easy with external library. But it is possible of cause.
So, should I separate this code in library?

ElderJoy:
>  The -base environments are mission DEVICE and MANUFACTURER pair of build flags somehow, either custom.h, src_build_flags of the custom env, ESPURNA_FLAGS=... environ, PLATFORMIO_SRC_BUILD_FLAGS=... environ, etc.

Yes, I noticed that base envs missed that flags, but not clearly understand how supposed to build them in this way. 

> You can try adding test/build/garland.h with the GARLAND_SUPPORT=1 (+ any additional flags), so the CI can try to build it

I added env:garland with correspondent flags. Ok, will add configuration for CI also.

mcspr:
See https://github.com/xoseperez/espurna/wiki/PlatformIO#building
The idea is to preserve the build environment when changing build flags. Locally or in CI, effectively it skips the framework + libs build, only rebuilding the relevant project source

> As for separate garland-dedicated code into separate library. I supposed this functionality as extendable. Anyone can easily add and modify animations. Hope with community help we will have much more animations soon.
It would be not so easy with external library. But it is possible of cause.
So, should I separate this code in library?

Let me get back on that...

mcspr:
So, how about:
- reduce globals like pallets, colors etc. expecting 'extern', sometimes cross-including. Meaning, we explicitly create lists in our code, and give to garland sort-of-lib to process.
- every animation class itself can keep the data-member, no need for the globals (specifically, AnimComet)
- try to give values to the lib part, instead of having things like BRI, LEDS, ets.
- make sure to add `#pragma once`es so headers don't get included more than once

Would the same structure as sensors make sense here? Re. the 1st point, animations could be a single-header definitions, without the need to have any names for them (i.e. `extern AnimStart anim_start` -> just `AnimStart{}` in the animations list)

code/espurna/garland/anim_spread.cpp
```diff
@@ -0,0 +1,46 @@
...
+AnimSpread::AnimSpread() : Scene::Anim("Spread") {
+}
+
+void AnimSpread::SetupImpl() {
+    inc = random(2,8);
+    memset(seq, 0, LEDS);
+}
+
+void AnimSpread::Run() {
+    memset(_leds, 0, 3*LEDS);
```
mcspr: I'd try to avoid memset when possible for objects. It is a valid C-pattern, but not really in C++ where `struct` aka `class` can have a constructor attached (and note there's no real difference between the two, only in the requirement for explicit `public:` or `private:` at the start of the body).

code/espurna/garland/color.h
```diff
@@ -0,0 +1,83 @@
...
+        struct
+        {
+            byte r;
+            byte g;
+            byte b;
+        };
+        byte raw[3];
+    };
+
+    inline Color() __attribute__((always_inline)) : r(0), g(0), b(0) {}
```
mcspr: It is, technically, already inline, even without the keyword. Below one as well, anything declared in the class body is inline implicitly

code/espurna/garland/color.h
```diff
@@ -0,0 +1,83 @@
...
+{
+    union
+    {
+        struct
+        {
+            byte r;
+            byte g;
+            byte b;
+        };
+        byte raw[3];
```
mcspr: Is there a point to the union? raw is not really used at all

code/espurna/garland/palette.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+// palettes below are taken from http://www.color-hex.com/color-palettes/ (and modified)
+
+Color PalIceBlue_[] = 
+{
+    0xffffff, 0x0000ff, 0x00ffff
+};
+Palette PalIceBlue = { 3, PalIceBlue_};
+
+Palette * pals[PALS] = {&PalRgb, &PalRainbow, &PalRainbowStripe, &PalParty, &PalHeat, &PalFire, &PalIceBlue};
```
mcspr: Not really sure what to do about underscore-temporaries :/ There's always std::vector though, which would help by allowing inline definition. Also avoids explicit size definition

code/espurna/garland.cpp
```diff
@@ -0,0 +1,172 @@
...
+        .onConnected(_garlandWebSocketOnConnected)
+        .onKeyCheck(_garlandWebSocketOnKeyCheck)
+        .onAction(_garlandWebSocketOnAction);
+#endif
+
+    espurnaRegisterLoop(garlandLoop);
+    espurnaRegisterReload(_garlandReload);
+
+    pixels.begin();
+    randomSeed(analogRead(0) * analogRead(1));
```
mcspr: Not valid on 8266, there's only one analog pin. There is `secureRandom` though, which ignores the seed

*edit:* however, it sometimes works, since analogRead is helpful enough to output the result as either digital 0 or 1023:
https://github.com/esp8266/Arduino/blob/dd8c2e6b8bfd15b02995fc4f7e0005ad2cf02942/cores/esp8266/core_esp8266_wiring_analog.cpp#L37

code/espurna/garland.cpp
```diff
@@ -0,0 +1,172 @@
...
+    if (strcmp(action, NAME_GARLAND_SWITCH) == 0) {
+        if (data.containsKey("status") && data.is<int>("status")) {
+            const char* payload = data["status"];
+            if (strlen(payload) == 1) {
+                if (payload[0] == '0') {
+                    _garland_enabled = false;
+                    pixels.clear();
+                    pixels.show();
+                } else if (payload[0] == '1')
+                    _garland_enabled = true;
```
mcspr: `_garland_enabled = (1 == data.as<int>("status"));`?

code/espurna/garland.cpp
```diff
@@ -0,0 +1,172 @@
...
+#include "web.h"
+#include "ws.h"
+
+const char* NAME_GARLAND_ENABLED        = "garlandEnabled";
+const char* NAME_GARLAND_BRIGHTNESS     = "garlandBrightness";
+
+const char* NAME_GARLAND_SWITCH         = "garland_switch";
+const char* NAME_GARLAND_SET_BRIGHTNESS = "garland_set_brightness";
+
+#define ANIMS                             8 //number of animations
```
mcspr: There is always `sizeof(anims) / sizeof(std::uintptr_t)`, in case we are using non-c++ containers and just plain arrays

ElderJoy:
Thank you for your advises. I will try to tune code correspondently to your comments.
As I mentioned in description, it is just raw port from existing project, that is definitely C-style one. I slightly refactored it to make more C++ style, but it is far from ideal indeed. I decided to put it now to allow someone to use it now. But you're right, it should be refined before commit.
Will notify you when done. Also, if you find more places to refine, please notify.

code/espurna/garland/color.h
ElderJoy: It used when Color assigned value like 0xFF0000

mcspr:
```cpp
inline Color( uint32_t colorcode)  __attribute__((always_inline))
    : r((colorcode >> 16) & 0xFF), g((colorcode >> 8) & 0xFF), b((colorcode >> 0) & 0xFF)
    {
    }
```
<- it simply shifts and accesses named bytes, not indexes. Byte order in mind, I'd keep only the named ones

(or, [rotate them via `#if defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__` like here](https://github.com/microsoft/terminal/blob/main/src/inc/til/color.h) / at least leave a comment if that is ever needed)

ElderJoy:
Here we go.
@mcspr please take a look at code again. I refactored it taking in account your notes.

code/espurna/garland.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+bool _garland_enabled                   = true;
+unsigned long _last_update              = 0;
+unsigned long _interval_effect_update;
+
+int paletteInd;
+int animInd                             = 0; 
+
+// Palette should
+std::vector<Palette> pals = {
```
mcspr: Top level vector is not really required, as compiler is able to work out the size even with a simple `pals[]`

code/espurna/garland/palette.h
```diff
@@ -0,0 +1,88 @@
...
+
+#if GARLAND_SUPPORT
+
+#include <vector>
+
+#include "color.h"
+
+class Palette {
+   public:
+    Palette(const String name, std::vector<Color> colors) : _name(name), _numColors(colors.size()), _colors(colors), _cache(256) {
```
mcspr: Given the declaration is static and only at build, just use `const char* name` as member?
(also saves a bit of ram, since it changes 20byte String for 4 byte pointer)

code/espurna/garland/scene.cpp
```diff
@@ -0,0 +1,119 @@
...
+unsigned int rng() {
+    static unsigned int y = 0;
+    y += micros();  // seeded with changing number
+    y ^= y << 2;
+    y ^= y >> 7;
+    y ^= y << 7;
+    return (y);
+}
+
+byte rngb() { return (byte)rng(); }
```
mcspr: Why not `secureRandom()` (aka `RANDOM_REG32`, it has the same 32bit num range)?

code/espurna/garland/scene.h
```diff
@@ -0,0 +1,81 @@
...
+
+#if GARLAND_SUPPORT
+
+#include <array>
+#include <vector>
+
+#include "anim.h"
+
+#define DEFAULT_BRIGHTNESS    12    // brightness adjustment, up to 255
+#define TRANSITION_MS       1000    // transition time between animations, ms
```
mcspr: Move to general.h with `GARLAND_...` prefix? Or just place in the body, without the definition (we don't want generic names added by just including the scene.h)

code/espurna/garland/anim.h
```diff
@@ -0,0 +1,56 @@
...
+    //glow animation - must be called for each LED after it's BASIC color is set
+    //note this overwrites the LED color, so the glow assumes that color will be stored elsewhere (not in leds[])
+    //or computed each time regardless previous leds[] value
+    void glowForEachLed(int i);
+
+    //glow animation - must be called at the end of each animaton run
+    void glowRun();
+
+private:
+    const String _name;
```
mcspr: *c++ quirk*
`const Object` members are generally a bad idea when dealing with things living in contiguous containers like vector, since we are not allowing it to be ['moved' and save time on duplicating the object, simply replacing new members with the old ones](https://en.cppreference.com/w/cpp/utility/move) (e.g., it will always call the `String(const String&)`, cloning the internal string buffer). Any resize operation or the case when we insert something in the middle will cause the relocation of the objects inside

It is harmless here though, since the String allows us to store up to 11bytes inside of the `String` object itself instead of always calling `malloc(strlen(...) + 1)`

code/espurna/garland/anim.h
```diff
@@ -0,0 +1,56 @@
+#pragma once
+
+#include "color.h"
+
+#define BRA_AMP_SHIFT          1    // brigthness animation amplitude shift. true BrA amplitude is calculated 
+                                    // as (0..127) value shifted right by this amount
+#define BRA_OFFSET           127    //(222-64) // brigthness animation amplitude offset
```
mcspr: Same as scene.h, in case of 'unity' build it will leak define after anim.h is included
`GARLAND_...` / `namespace garland` / as constexpr in .cpp ?

code/espurna/garland/anim.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+
+void Anim::glowRun() { braPhase += braPhaseSpd; }
+
+bool operator== (const Color &c1, const Color &c2)
+{
+    return (c1.r == c2.r && c1.g == c2.g && c1.b == c2.b);
+}
+
+#endif
```
mcspr: eol
*(<s>and I thought git should warn about this on commit b/c of the .gitattributes...</s> git does warn, but only for .ini and not .cpp)*

code/espurna/garland.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+//------------------------------------------------------------------------------
+void _garlandWebSocketOnAction(uint32_t client_id, const char* action, JsonObject& data) {
+    if (strcmp(action, NAME_GARLAND_SWITCH) == 0) {
+        if (data.containsKey("status") && data.is<int>("status")) {
+            _garland_enabled = (1 == data["status"].as<int>());
+            if (!_garland_enabled) {
+                pixels.clear();
+                pixels.show();
```
mcspr: note that this runs in a different 'thread' than loop, which does not really like long operations. clear seems ok, show idk how long it could take in case leds are turning off
One option is to `schedule_function([]() { stuff });`, or just set some `clear pixels` bool flag for the loop()?

code/espurna/garland/anim.h
```diff
@@ -0,0 +1,56 @@
...
+
+#define BRA_AMP_SHIFT          1    // brigthness animation amplitude shift. true BrA amplitude is calculated 
+                                    // as (0..127) value shifted right by this amount
+#define BRA_OFFSET           127    //(222-64) // brigthness animation amplitude offset
+
+class Palette;
+
+class Anim {
+public:
+    Anim(String name);
```
mcspr: -//- as palette, simply use `const char*`?

code/espurna/garland.cpp
```diff
@@ -0,0 +1,223 @@
...
+    // Pastel: Pastel Fruity Mixture
+    Palette("Pastel", {0x75aa68, 0x5960ae, 0xe4be6c, 0xca5959, 0x8366ac}),
+
+    // Green: Vibrant greens
+    Palette("Green", {0x89ff01, 0x42c501, 0x349404, 0x0f6902, 0x004208})};
+
+Adafruit_NeoPixel pixels = Adafruit_NeoPixel(LEDS, PIN, NEO_GRB + NEO_KHZ800);
+Scene scene(&pixels);
+
+std::vector<Anim*> anims = {new AnimStart(), new AnimPixieDust(), new AnimSparkr(), new AnimRun(), new AnimStars(),
```
mcspr: Same as above regarding top level vector

<s>new is redundant btw</s>
nvm, new is not redundant because of virtual requirement, my mistake

code/espurna/garland.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+//------------------------------------------------------------------------------
+bool garlandEnabled() {
+    return _garland_enabled;
+}
+
+//------------------------------------------------------------------------------
+// Setup
+//------------------------------------------------------------------------------
+void commonSetup() {
```
mcspr:
```suggestion
void _garlandCommonSetup() {
```
or just use configure () directly

mcspr (approved):
lgtm besides the comments

code/espurna/garland.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+#define EFFECT_UPDATE_INTERVAL_MIN      5000  // 5 sec
+#define EFFECT_UPDATE_INTERVAL_MAX      10000 // 10 sec
+
+bool _garland_enabled                   = true;
+unsigned long _last_update              = 0;
+unsigned long _interval_effect_update;
+
+int paletteInd;
+int animInd                             = 0;
```
mcspr: 'static' in loop?

code/espurna/garland.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+    // Green: Vibrant greens
+    Palette("Green", {0x89ff01, 0x42c501, 0x349404, 0x0f6902, 0x004208})};
+
+Adafruit_NeoPixel pixels = Adafruit_NeoPixel(LEDS, PIN, NEO_GRB + NEO_KHZ800);
+Scene scene(&pixels);
+
+std::vector<Anim*> anims = {new AnimStart(), new AnimPixieDust(), new AnimSparkr(), new AnimRun(), new AnimStars(),
+                            new AnimSpread(), new AnimRandCyc(), new AnimFly(), new AnimComets(), new AnimAssemble()};
+constexpr bool disableAutoChangeEffects = false;
```
mcspr: unused

code/espurna/garland/anim.cpp
```diff
@@ -0,0 +1,38 @@
+#if GARLAND_SUPPORT
```
mcspr: Hmmm... Note that this won't build unless it's either single-source build or -DGARLAND_SUPPORT=1 in the flags.
(and why I wanted this as a lib :)

PIO also accidentally picks up Adafruit lib via code/espurna/garland/scene.cpp when resolving build deps, but it's not really working properly and is something that I wanted to fix anyway to be able to build without all the libs all the time

code/espurna/garland/anim_assemble.cpp
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+}
+
+void AnimAssemble::Run() {
+}
+
+AnimAssemble anim_assemble;
+
+#endif  // GARLAND_SUPPORT
```
mcspr: eol

code/espurna/garland/anim_comets.cpp
```diff
@@ -0,0 +1,40 @@
...
+            int p = c.head + l * -c.dir;
+            if (p >= 0 && p < numLeds) {
+                leds[p] = c.points[l];
+            }
+        }
+        c.head = c.head + c.speed * c.dir;
+    }
+}
+
+#endif  // GARLAND_SUPPORT
```
mcspr: eol
(could you try adding *.cpp to /.gitattributes of the repo? copying existing .ini line)

code/espurna/garland.cpp
ElderJoy: Do you mean define pals like this: `Palette pals[] = { ..`

I obtained next errors:

espurna\garland.cpp: In function 'void garlandLoop()':
espurna\garland.cpp:185:73: error: request for member 'size' in 'pals', which is of non-class type 'Palette [11]'
         while (prevPalInd == paletteInd) paletteInd = secureRandom(pals.size());
                                                                         ^
espurna\garland.cpp: In function 'void garlandSetup()':
espurna\garland.cpp:215:36: error: request for member 'size' in 'pals', which is of non-class type 'Palette [11]'
     paletteInd = secureRandom(pals.size());
                                    ^
*** [.pio\build\garland-wemos\src\garland.cpp.o] Error 1

In this case I need to define it's size using sizeof and can't using modern iteration. Do we really need it? Or I did something wrong?

I also can use std::array - it eat less size and works faster than vector and good for static usage but require size in definition. For me it is the best solution.

mcspr: I may be wrong, but you can also do `std::distance(std::begin(pals), std::end(pals))` in place of `.size()` for basic arrays (which is why we could do `for (auto& x : pals) { ... }` stuff)
`std::array` also works, in case you don't mind typing out the size.

code/espurna/garland/scene.cpp
ElderJoy: It is much faster than secureRandom that is valuable for intensive usage. For example average  numbers:
with **rngb**: [080205] [GARLAND] Anim: RandCyc    Pal: Rainbow  timings: **calc:  100** pixl:  166 show: 1872
with **secureRandom**: [255650] [GARLAND] Anim: RandCyc    Pal: Blue     timings: **calc:  208** pixl:  147 show: 1871
I think use rngb for optimization PixieDust animation as it is most time-consuming one.
I'll add comment to it.

code/espurna/garland.cpp
ElderJoy: Well, compiler don't let me use std::distance in constexpr function. But at least 
`constexpr size_t palsSize() { return sizeof(pals)/sizeof(pals[0]); }`
works well for me. Use pals[] with it.

code/espurna/garland.cpp
ElderJoy: There is no class for garland (think it is redundant) I store "state": current used animation and palette in "static". Would it be better to define them static int inside the loop?

code/espurna/garland/anim.cpp
ElderJoy: I agree, that adding Adafruit_Neopixel to the whole libs set just for garland feature is redundant. And also can cause unwanted side effects as similar library already included. I can move most files from garland folder to the library, that can hosted on github.com/ElderJoy/Garland, now it way more ready to movement.
But I prefer leave anims in garland folder inside espurna project as examples and to encourage community write their own animations.
And I don't really know how to include Adafruit_Neopixel just for garland except placing it as lib_deps for specially added [env:garland-wemos].
So, please tell me what to do better.

mcspr: > I agree, that adding Adafruit_Neopixel to the whole libs set just for garland feature is redundant. And also can cause unwanted side effects as similar library already included. I can move most files from garland folder to the library, that can hosted on github.com/ElderJoy/Garland, now it way more ready to movement.

The gist of the PIO feature is https://docs.platformio.org/en/latest/librarymanager/ldf.html btw
But, I have not had much luck making it ignore any libs specified in lib_deps based on #define or -D... flags. It seems to only work correctly when:
- lib_deps is empty
- lib is somehow installed in either ~/.platformio/lib/$libname or the location of lib_extra_dirs=...
.pio/libdeps/$env/... surprisingly does not work
- project's .h #include library header (for some reason even with the deep+, main.cpp including the lib is ignored :/)

Otherwise, it simply builds everything in lib_deps list regardless of the things it finds via the scanner. I do agree it is better to keep animations in the repo though. Then, not sure it is really worth splitting only a small part of the code. atm I'd focus on adafruit class used multiple times. So, either:
- scene.cpp contents can be moved into garland.cpp
- the interface and implementation of the scene should not refer to the pixel class (e.g. it can link with garland.cpp.o class implementing wrappers calling pixel.show(), pixel.set...(), etc.)

> But I prefer leave anims in garland folder inside espurna project as examples and to encourage community write their own animations.

The 'sensor' approach I mentioned also avoids another .cpp.o to link, by making animation classes header-only - moving anim.h declaration and AnimSomething.h into one file, then either garland.cpp or anim.h includes everything available. It also allows top-level `#if GARLAND_SUPPORT` to work correctly. Dummy .o file is still built though (~57kb in size), not sure what to do about that just yet.

code/espurna/garland.cpp
mcspr: Yes, the reason being nothing else uses these but the loop function

code/espurna/garland.cpp
mcspr: oops, yes. distance is compile-time with c++17, which is only available when building -git-base environments atm
(as well as simple `std::array something { 1,2,3,4,5 }` syntax...)

code/espurna/garland/anim.cpp
ElderJoy: Ok, then I making next:
1. moving scene.cpp into garland.cpp
2. other sources from garland folder combining into one header garland.h and place next to garland.cpp

I totally agree with 1 but a bit doubt about 2. Combining all sources into one huge header decrease its readability, maintainability and extensibility. In case when two community members add their animations they probably would forced to merge this header. It's not a big deal but I prefer avoid that if possible. 
But if it is much better from whole project perspective and leads to smaller binary size and proper working of #if GARLAND_SUPPORT I agree to do that. Should I?

mcspr: Not all though, only animation part. What I meant about the animation headers was:
- `garland/animation.h` with basic interface
- `garland/animation/comet.h` (or `garland/AnimationComet.h`)
- `garland/animation/assemble.h`
- and etc.

ElderJoy: Ok, got it. Looks reasonable. Will do in that way and then commit new bunch of changes. However, I'm adding one more feature now - animation speed and possibility to control it from UI. So, commit after finishing it. Soon :)

ElderJoy:
Ok, next attempt. Fixed all pointed places.
New features:
- possibility to control common animation speed from UI
- reset brightness and speed to the default
- each animation can tune self speed factor
- calculate average frame rate for animation and output with debug info
Tune some animations
Please take a look on it now.

code/espurna/garland/palette.h
```diff
@@ -0,0 +1,88 @@
...
+        // decimal part is used to interpolate between the two colors
+        float t0 = i * _numColors - trunc(i * _numColors);
+        return _colors[i0].interpolate(_colors[i1], t0);
+    }
+
+    Color getCachedPalColor(byte i) {
+        if (!_cache[i].empty())
+            return _cache[i];
+        
+        Color col = getPalColor((float)i / 256);
```
mcspr, line 41: Was 256 meant to be cache size?

code/espurna/garland/scene.h
```diff
@@ -0,0 +1,104 @@
...
+    bool setUpOnPalChange = true;
+
+    unsigned long sum_calc_time = 0;
+    unsigned long sum_pixl_time = 0;
+    unsigned long sum_show_time = 0;
+    unsigned int calc_num = 0;
+    unsigned int show_num = 0;
+    unsigned int pixl_num = 0;
+
+    std::array<byte, 256> bri_lvl = {{0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3,
```
mcspr, line 97: just to verify, it is not in PROGMEM to improve transition speed?
(esp8266 has a lot RAM to spare and there's only one scene, so not a big deal :)

mcspr (commented):
Something have been bothering me... sorry that I missed this when initially reading the description

Original project does not specify any license. And, based on the Github's explanation from here:
https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/licensing-a-repository#choosing-the-right-license
> However, without a license, the default copyright laws apply, meaning that you retain all rights to your source code and no one may reproduce, distribute, or create derivative works from your work. If you're creating an open source project, we strongly encourage you to include an open source license.
...
 If you publish your source code in a public repository on GitHub, according to the Terms of Service, other users of GitHub have the right to view and fork your repository.

It is not really clear if with the changes here it is 'derivative work' or something else. We are not commercially distributed software though, so there's that. But, I wonder if we could at least make the author aware of the inclusion to receive an ok that we can use it?
(there is also a continued work, as the project deemed deprecated, but it still lacks any license there as well)

ElderJoy:
> It is not really clear if with the changes here it is 'derivative work' or something else. We are not commercially distributed software though, so there's that. But, I wonder if we could at least make the author aware of the inclusion to receive an ok that we can use it?
> (there is also a continued work, as the project deemed deprecated, but it still lacks any license there as well)

Yes, you're right. It is better to contact to Author of library that I used as base one. Unfortunately, I didn't found any contact information neither in GitHub account nor in the code. So, I found a bit weird way - I created issue in the currently maintained project. Hope Author will contact to me or write here.

code/espurna/garland/scene.h
ElderJoy: It was in PROGMEM originally. I placed it as dynamic array as I understand one of your previous comments in this way. I can put it back to PROGMEM, there is no problem for me. Am I right understand, that RAM works faster? Than probably it is better to live it as is (taking in account that there is only one scene indeed :) )?

code/espurna/garland/palette.h
ElderJoy: Yes. And it takes 256*3 bytes per palette. But dramatically reduce processing time. And some animations use this actively.
Parameter (byte i) has 256 possible variants. So, there are 256 variants of interpolated colors for the palette doesn't matter how many different static colors it has. Any animation that produce smooth color sequence for palette use this for each pixel each Run cycle.

code/espurna/garland/scene.h
mcspr: Yep, I'd just leave it as-is.
It is indeed faster to deal with RAM, although ROM reads are backed up by a cache.

ElderJoy:
@mcspr Ok, what we can do more regarding License? Version that I used is deprecated now (I used Arduino version) and I rewrite it  mostly. But of cause base version was that I pointed out and most animations from there. So, can we use it based on GitHub licensing politics?

mcspr:
It may be a stretch, but assuming we follow the chain to the https://github.com/bportaluri/ALA, it is GPLv3 still even without an explicit statement that it is.
I'd add attribution as 'inspired / based on the code from ... and ...' to the files though.

*edit:* Github docs are only meant to point out that the author holds the rights, but does not tell us what our rights are and what we could do with the code (besides just looking at it)

ElderJoy:
Added header with reference to the project garland based on

mcspr:
Just in time :) sry for delays

I also wanted to point out that Adafruit's neopixel lib might not be the most efficient one on the platform, you might want to check out either:
- https://github.com/FastLED/FastLED
- https://github.com/Makuna/NeoPixelBus

and esp8266 code there was not really updated since august of 2015
https://github.com/adafruit/Adafruit_NeoPixel/commits/master/esp8266.c

mcspr:
Could you update the wiki describing the module?

ElderJoy:
Well, it works not so bad with Adafruit's Neopixel currently. However, it eat most time for showing pixels inside library and if it can be speed up, than it worth to do. So, sure, I will check both mentioned libs and adapt library to better one. But my proposition to make it with next pull request. I going to add couple animations also.

ElderJoy:
> Could you update the wiki describing the module?

Yep, will do.

PR #2425, opened 2021-02-08
Garland: Added control over mqtt, fix wdt reset for long led stripes. Some updates for animations.

ElderJoy:


mcspr:
btw, do you mind I still squash everything here? garland branch from git(hub) pov is still a separate tree, starting from the very first PR you did back in 25 Jan 2020. merge-commit will make commit history a nightmare to read, or I can go over the latest commits and cherry-pick them manually on top of the dev

ElderJoy:
I'm Ok with any solution you'll find the best. You're right, I'll start new branch for the next improvements.

code/espurna/garland.cpp
```diff
@@ -87,11 +138,15 @@ constexpr size_t palsSize() { return sizeof(pals)/sizeof(pals[0]); }
...
-                 new AnimRandCyc(), new AnimFly(), new AnimComets(), new AnimAssemble(), new AnimDolphins(), new AnimSalut()};
+Anim* anims[] = {new AnimGlow(), new AnimStart(), new AnimPixieDust(), new AnimSparkr(), new AnimRun(), new AnimStars(), new AnimSpread(), 
+                 new AnimRandCyc(), new AnimFly(), new AnimComets(), new AnimAssemble(), new AnimDolphins(), new AnimSalut(), new AnimFountain()};
 
 constexpr size_t animsSize() { return sizeof(anims)/sizeof(anims[0]); }
 
+#define START_ANIMATION  1
+Anim* _currentAnim       = anims[1];
+Palette* _currentPalette = &pals[0];
+auto one_color_palette = std::unique_ptr<Palette>(new Palette("White", {0xffffff}));
```
mcspr, line 149: Check out https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
I did a backport, since I constantly ran into this issue:
https://github.com/xoseperez/espurna/blob/7434ea3878f9a81752694fa87127cec676d58138/code/espurna/compat.h#L107-L110

code/espurna/garland.cpp
```diff
@@ -178,59 +252,209 @@ void _garlandWebSocketOnAction(uint32_t client_id, const char* action, JsonObjec
...
+                uint32_t color = (uint32_t)strtoul(palette, NULL, 0);
+                if (color != 0) {
+                    one_color_palette.reset(new Palette("Color", {color}));
+                    newPalette = one_color_palette.get();
+                }
+            }
+        }
+    }
+
+    unsigned long newAnimDuration = LONG_MAX;
```
mcspr, line 353: ULONG_MAX?
also, see https://en.cppreference.com/w/cpp/types/numeric_limits

e.g. `std::numeric_limits<unsigned long>::max()` (or, when needed, `decltype(newAnimDuration)` to avoid issues)
`unsigned long var { ... };` init will also warn about the mistypes like that, just wary of the `auto` that will be deduced into a list instead of the var, b/c we use such an old gcc version

code/espurna/garland.cpp
```diff
@@ -37,12 +67,33 @@ const char* NAME_GARLAND_SET_BRIGHTNESS = "garland_set_brightness";
...
+const char* MQTT_PAYLOAD_BRIGHTNESS     = "brightness";
+const char* MQTT_PAYLOAD_ANIM_SPEED     = "speed";
+const char* MQTT_PAYLOAD_ANIMATION      = "animation";
+const char* MQTT_PAYLOAD_PALETTE        = "palette";
+const char* MQTT_PAYLOAD_DURATION       = "duration";
+
+const char* MQTT_COMMAND_IMMEDIATE      = "immediate";
+const char* MQTT_COMMAND_RESET          = "reset"; // reset queue
+const char* MQTT_COMMAND_QUEUE          = "queue"; // enqueue command payload
+const char* MQTT_COMMAND_SEQUENCE       = "sequence"; // place command to sequence
```
mcspr, line 83: *pedantic note*

`const char* const` is a more correct type here, since we are still able to `MQTT_TOPIC_GARLAND = "blah";` 
(or `const char var[] = ...;`, to const-qualify both at the same time)

mcspr (commented):
I did a rebuild [here](https://github.com/xoseperez/espurna/actions/runs/600174992)
Will merge in a bit and close

ElderJoy:
> I did a rebuild [here](https://github.com/xoseperez/espurna/actions/runs/600174992)
> Will merge in a bit and close

So, should I fix placed, you've noted on the garland-rebase branch?

code/espurna/garland.cpp
ElderJoy: Not quite understand this tricky issue with creating unique_ptr. I unable to create unique_ptr with make_unique just placing arguments into it. Finally I found working set as this: 
`auto one_color_palette = std::make_unique<Palette>("White", std::vector<Color>({0xffffff}));`
Would it be better?

mcspr: s/this issue/make_unique not being available in the c++11/

initializer_list template deduction is very dumb though, i totally forgot. it either wants `std::vector<Color>` or `std::initializer_list<Color>`. make_unique should always be preferred... for multiple reasons, here being we only specify the type once
e.g.
```cpp
using Colors = std::vector<Color>; // also note other declarations, maybe also inside of namespace { ... }
auto one_color_palette = std::make_unique<Palette>("White", Colors{{0xffffff}});
```

mcspr:
> So, should I fix placed, you've noted on the garland-rebase branch?

Already in dev.

I removed the platformio.ini changes though, idk if those were intended. As a reminder, there is platformio_override.ini:
https://github.com/xoseperez/espurna/blob/dev/code/platformio_override.ini.example

ElderJoy:
@mcspr Thank you!

PR #2429, opened 2021-03-21
Remove the need to patch BME680's static library

ruimarinho:
Bosch Sensortec has listened to the community and released a version of their proprietary library which does not require any modification to the linker script.


mcspr:
Idk about the CI, something with the recent PIO changes. I think it either wants an `pio update` to fetch latest native platform version, or library dependencies break it :/ locally, it works ok
https://github.com/platformio/platform-native/issues/12

Also idk about the new version. I watched the repo (bosch response time is stellar :) and noticed these replacing old blob sizes
https://github.com/BoschSensortec/BSEC-Arduino-library/commit/7a9357566c75048331c15b55a4eb20f1de14f36e#diff-61d5bff668b4530537b21e88250797087f2bad8ccb4aa261551665b293aafb62R284
https://github.com/BoschSensortec/BSEC-Arduino-library/commit/7a9357566c75048331c15b55a4eb20f1de14f36e#diff-61d5bff668b4530537b21e88250797087f2bad8ccb4aa261551665b293aafb62R303
https://github.com/BoschSensortec/BSEC-Arduino-library/blob/7a9357566c75048331c15b55a4eb20f1de14f36e/src/inc/bsec_datatypes.h#L87
esp loop() runs in a fixed-size 4kb stack, and new version functions eat half of it (and I hope it is only specific to the public part). It should not be a big problem, but I wonder what are the values with `ESP.getFreeContStack()` + `ESP.resetFreeContStack()` (reset paints the memory, get searches for the paint to find the border after stack is filled with data)

mcspr:
re. above. if something breaks - I'd look there :) at least that part of the lib is possible to patch
thanks!

PR #2430, opened 2021-03-25
 Add anim_waves

ElderJoy:
Added animation waves.
Some minor improvements and fixes.

### DIFF
--- a/code/espurna/garland/animations/anim_glow.h
+++ b/code/espurna/garland/animations/anim_glow.h
@@ -22,7 +22,7 @@ class AnimGlow : public Anim {
                 glowForEachLed(i);
             }
         } else {
-            for (int i = 0; i < numLeds; ++i) {
+            for (int i = numLeds - 1 ; i >= 0; --i) {
                 leds[i] = curColor;
                 glowForEachLed(i);
             }
--- a/code/espurna/garland/animations/anim_waves.h
+++ b/code/espurna/garland/animations/anim_waves.h
@@ -3,8 +3,6 @@
 #include "../anim.h"
 #include "../palette.h"
 
-#define _BRA_OFFSET 65
-
 //------------------------------------------------------------------------------
 class AnimWaves : public Anim {
     private:
--- a/code/espurna/garland/animations/anim_waves.h
+++ b/code/espurna/garland/animations/anim_waves.h
@@ -1,0 +1,57 @@
+#if GARLAND_SUPPORT
+
+#include "../anim.h"
+#include "../palette.h"
+
+#define _BRA_OFFSET 65
+
+//------------------------------------------------------------------------------
+class AnimWaves : public Anim {
+    private:
+     int bra_phase;
+     int bra_phase_speed;
+     byte bra_speed;
+     byte bra_min;
+     int sign;
+
+   public:
+    AnimWaves() : Anim("Waves") {
+    }
+
+    void SetupImpl() override {
+        curColor = palette->getRndInterpColor().max_bright();
+        glowSetUp();
+        sign = braPhaseSpd > 128 ? -1 : 1;
+        bra_phase = secureRandom(255);
+        bra_phase_speed = secureRandom(2, 5);
+        bra_speed = secureRandom(4, 12);
+        bra_min = secureRandom(20, 30);
+    }
+
+    void Run() override {
+        int bra = bra_phase;
+        int8_t bra_inc = -sign * bra_speed;
+
+        for (int i = 0; i < numLeds; ++i) {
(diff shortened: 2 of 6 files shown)


code/espurna/garland/animations/anim_waves.h
```diff
@@ -0,0 +1,57 @@
+#if GARLAND_SUPPORT
+
+#include "../anim.h"
+#include "../palette.h"
+
+#define _BRA_OFFSET 65
```
mcspr: ?

code/espurna/garland/color.h
```diff
@@ -81,6 +98,12 @@ struct Color
         Serial.print(("b="));Serial.println(b);
     }
 
+    std::string to_str() {
+        char buf[20];
+        sprintf(buf, "r=%3d, g=%3d, b=%3d", r, g, b);
```
mcspr: `String`?
also note there's `%hhu` for the bytes
(but I am never sure of the effect of differently-sized integers in format and va_args)

code/espurna/garland/animations/anim_glow.h
```diff
@@ -22,7 +22,7 @@ class AnimGlow : public Anim {
                 glowForEachLed(i);
             }
         } else {
-            for (int i = 0; i < numLeds; ++i) {
+            for (int i = numLeds - 1 ; i >= 0; --i) {
```
mcspr, line 25: I wonder if begin() & rbegin() alternatives would clearer? i.e.
```cpp
auto begin = leds.begin();
auto end = leds.end();
while (begin != end) {
    doStuff(*it); // something that accepts Color&
    ++it;
}
```

There's also a way to bind method at runtime a-la virtual function:
https://godbolt.org/z/n3jaKrWYT

mcspr (commented):
CI technically works, it just needs `pio platform install native` in the test phase:
https://github.com/platformio/platformio-core/issues/3901

code/espurna/garland/animations/anim_waves.h
ElderJoy: Yes, this is redundant and should be removed

code/espurna/garland/color.h
ElderJoy: Yes, will use String. As for %hhu I will try. But %3d also works fine. Why not?

code/espurna/garland/animations/anim_glow.h
ElderJoy: glowForEachLed(i) requires using counter, so using iterator can be more complex in this case.

ElderJoy:
> CI technically works, it just needs `pio platform install native` in the test phase:
> [platformio/platformio-core#3901](https://github.com/platformio/platformio-core/issues/3901)

Not sure where to insert this step. Should it be included somewhere in the project or just installed once on build machine?

mcspr:
> > CI technically works, it just needs `pio platform install native` in the test phase:
> > [platformio/platformio-core#3901](https://github.com/platformio/platformio-core/issues/3901)
> 
> Not sure where to insert this step. Should it be included somewhere in the project or just installed once on build machine?

Something like this -> https://github.com/xoseperez/espurna/commit/76f187b13b122d1dbb1f09ca71d6da8fb5dabfa3
Pushed into dev, looks like it builds

ElderJoy:
> > > CI technically works, it just needs `pio platform install native` in the test phase:
> > > [platformio/platformio-core#3901](https://github.com/platformio/platformio-core/issues/3901)
> > 
> > 
> > Not sure where to insert this step. Should it be included somewhere in the project or just installed once on build machine?
> 
> Something like this -> [76f187b](https://github.com/xoseperez/espurna/commit/76f187b13b122d1dbb1f09ca71d6da8fb5dabfa3)
> Pushed into dev, looks like it builds

Thank you!

code/espurna/garland/animations/anim_glow.h
mcspr: True... but given it is a vector, it is a simple pointer arithmetic via `it - leds.begin()` or `leds.end() - it`, since size of the object is known beforehand
Or `std::distance`. Either works though

code/espurna/garland/color.h
mcspr: Main issue is overflow of the target buffer, since %3d is technically allowed to output more that 3 digits and up to what `std::numeric_limits<int>::max()` digit size is (including the minus sign). Another is implicit cast to `int` of the input, since printf is a variadic function and `va_list` is a dynamically typed container that will extract based on the format hints (`va_arg`)

GCC might warn about this, but not always :/

PR #2431, opened 2021-04-01
Feature/schedule api

### LINKED ISSUE #2288: Feature Request: Make schedules available via REST API
When using the schedule option, every change has to be done in the web interface.

Is it possible to make this function available via the rest API as well?

For example, something like
GET /API/schedule/list             get list of schedules
GET /API/schedule/{number}  get info about this schedule
PUT /API/schedule/{number}  put 0, 1 or 2 like for relays to turn schedule on or off
DELETE /API/schedule/{number}  remove schedule
POST /API/schedule/{number}  add new schedule

data could look like

time: string in format hh:mm like n web interface
days: string of numbers like in web interface
utc: 0 or 1
state: 0 or 1
relay: number of relay to affect
switch: 0, 1 or 2 (switch relay to ...)

profawk:
added REST API support for the scheduler


code/espurna/scheduler.cpp
```diff
@@ -60,6 +62,96 @@ constexpr bool restoreLast() {
...
+#if API_SUPPORT
+
+// UTILS
+
+template <typename V = const char*>
+bool _setJsonVariant(const SettingsKey& k, const JsonVariant& v) {
+    if (!v.is<V>()) {
+        return false;
+    }
+    return setSetting(k, std::move(v.as<V>()));
```
mcspr: move is redundant, v.as<V>() already becomes either T&& or const T& depending on the function

code/espurna/scheduler.cpp
```diff
@@ -60,6 +62,96 @@ constexpr bool restoreLast() {
...
+}
+
+template <>
+bool _setJsonVariant<const char*>(const SettingsKey& k, const JsonVariant& v) {
+    const auto& tmp = v.as<const char*>();
+    if (tmp == nullptr) {
+        return false;
+    }
+    return setSetting(k, std::move(tmp));
```
mcspr: no need to move, since the const char is handled by the setsetting

code/espurna/scheduler.cpp
```diff
@@ -60,6 +62,96 @@ constexpr bool restoreLast() {
...
+
+template <typename T = const char*, typename ObjKey>
+bool _setJsonKey(const SettingsKey& k, const JsonObject& o, const ObjKey& key) {
+    if (!o.containsKey(key)) {
+        return false;
+    }
+    return _setJsonVariant<T>(k, o[key]);
+}
+
+// taken from relay.cpp
```
mcspr: note of the `tryParseId` from the utils, and the `scheduler::build::max()` (and the rest of the ::build namespace)

code/espurna/scheduler.cpp
```diff
@@ -60,6 +62,96 @@ constexpr bool restoreLast() {
...
+
+
+bool _schApiSetSchedule(const unsigned char id, JsonObject& sched) {
+    if (!_setJsonKey<int>({"schSwitch", id}, sched, "switch")) {
+        return false;
+    }
+    if (sched.containsKey("enabled")) {
+        const auto& enabled = sched["enabled"];
+        if (!_setJsonVariant<bool>({"schEnabled", id}, enabled)) {
+            setSetting({"schEnabled", id}, std::move(enabled.as<String>()));
```
mcspr: both moves are redundant, -//- above

code/espurna/scheduler.cpp
```diff
@@ -347,6 +439,47 @@ void schSetup() {
...
+                for (unsigned char i = 0; i < SCHEDULER_MAX_SCHEDULES; ++i) {
+                    if (!hasSetting({"schSwitch", i})) continue;
+                    auto& sched = scheds.createNestedObject();
+                    _schApiPrintSchedule(i, sched);
+                }
+                return true;
+            },
+            [](ApiRequest&, JsonObject& root) {
+                unsigned char id = 0;
+                while (hasSetting({"schSwitch", id})) id++;
```
mcspr, line 444: check for max and fail when id is out of bounds?

code/espurna/scheduler.cpp
profawk: is `scheduler::build::max()` new?

mcspr: yes. this was part of another update to push constant definitions into const/constexpr functions and variables that came with the... sizable update 2d ago

profawk:
fixed the issues :)

mcspr:
thanks!

sychron:
Thanks!

PR #2471, opened 2021-09-11
Fix initialization of _adc_counts for BaseAnalogEmonSensor

msprunck:
`_adc_counts` is not initialized correctly in [BaseAnalogEmonSensor.h](https://github.com/xoseperez/espurna/blob/f748b6487ca172824266c873916e76786a0080df/code/espurna/sensors/BaseAnalogEmonSensor.h#L239) 

https://github.com/xoseperez/espurna/blob/f748b6487ca172824266c873916e76786a0080df/code/espurna/sensors/BaseAnalogEmonSensor.h#L239-L240

It produces 20 instead of 1024 when the resolution is 10 bits.

This PR fixes `_adc_counts`, restores `EMON_ANALOG_RESOLUTION` and sets the resolution of `EmonAnalogSensor` when instantiated.

### DIFF
--- a/code/espurna/config/sensors.h
+++ b/code/espurna/config/sensors.h
@@ -508,6 +508,10 @@
 #define EMON_ANALOG_SUPPORT             0       // Do not build support by default
 #endif
 
+#ifndef EMON_ANALOG_RESOLUTION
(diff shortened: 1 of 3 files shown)


mcspr:
@msprunck thanks!

Issue #2477, opened 2021-10-02
Serial Sensor


pbek:
I'm using the wonderful Espurna firmware on my cheap 5$ Sonoff devices (230V, 1 LED, 1 relays, 1 button) for some years to control all kind of stuff. The limitation of those devices is that they only have one extra GPIO pin routed on the board.

So I'm plugging very cheap Arduino Nano (nockoff) boards to the Sonoff boards via serial interface to get more GPIO pins and AD converters from the Arduino (still cheaper than to build everything with a "normal" ESP8266 or ESP32 board).

For that usecase I wrote a small `Serial Sensor` addition to Espurna that just uses some commands like `SENSOR0 1234` (sent by the Arduino via serial) to publish that value 1234 via MQTT.

Is there any interest that I share this addition upstream (just one cpp and h file)?

mcspr:
> For that usecase I wrote a small Serial Sensor addition to Espurna that just uses some commands like SENSOR0 1234 (sent by the Arduino via serial) to publish that value 1234 via MQTT.
> Is there any interest that I share this addition upstream (just one cpp and h file)?

Sure. There are even certain custom boards in the tree, so I don't see any harm here.

> So I'm plugging very cheap Arduino Nano (nockoff) boards to the Sonoff boards via serial interface to get more GPIO pins and AD converters from the Arduino (still cheaper than to build everything with a "normal" ESP8266 or ESP32 board).

Notice that recent versions have tried to tackle GPIO HAL, to handle things besides the hardware itself. Only one current example is mcp23xxx digital pins that can be used as relays and buttons (ref. mcp.cpp, gpio.{h,cpp}, relay.cpp, BasePin header). It is not completely done though, since not everything is done by using them. And certain analog pin functions are handled on case-by-case basis (like buttons and i2c extenders)

pbek:
The lack of "reachable" GPIO/AD pins on the https://github.com/xoseperez/espurna/wiki/Hardware-Itead-Sonoff-Basic board drove me to the modification. 😅  There only is GPIO14.

![image](https://user-images.githubusercontent.com/1798101/135741510-d7c25f4f-465b-41b5-8034-653c8d140659.png)

pbek:
Code looks like this: https://github.com/xoseperez/espurna/pull/2478

mcspr:
[...]

PR #2478, opened 2021-10-03
Add SERIAL_SENSOR support (#2477)

pbek:
Implements https://github.com/xoseperez/espurna/issues/2477


code/espurna/serialsensor.cpp
```diff
@@ -0,0 +1,141 @@
...
+    terminalRegisterCommand(F("SENSOR0"), [](const terminal::CommandContext& ctx) {
+        if (ctx.argc == 2) {
+            String value = String(ctx.argv[1]);
+            mqttSend("SENSOR0", value.c_str());
+            ctx.output.printf_P(PSTR("SENSOR0 has value %s\n"),
+                value.c_str());
+        }
+
+        terminalOK(ctx);
+    });
```
mcspr: b/c of the code repetition, maybe what we want is `mqtt.send` command with 2 args - 'topic' and 'payload'? could also just add it to the mqtt.cpp directly
also note of the 'retained' option, should these be ephemeral messages?

code/espurna/serialsensor.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+// -----------------------------------------------------------------------------
+// COMMUNICATIONS
+// -----------------------------------------------------------------------------
+
+/**
+ * Setup function
+ *
+ * I haven't managed to get the terminalRegisterCommand commands in a method
+ * because of the lambda function
```
mcspr: btw there's at least one way to 'proxy' certain dynamic arguments inside of a lambda like this:
https://github.com/xoseperez/espurna/blob/62d83c6194630076fa198ddbf0a899ae7fb79672/code/espurna/sensors/PZEM004TSensor.h#L414-L415
(and true, assuming that this only gets called exactly once)

code/espurna/serialsensor.cpp
pbek: > b/c of the code repetition, maybe what we want is mqtt.send command with 2 args - 'topic' and 'payload'? could also just add it to the mqtt.cpp directly

But how would I get the variables into the callback of `terminalRegisterCommand`? I couldn't sneak them in in the `[]`.

> also note of the 'retained' option, should these be ephemeral messages?

Yes, I guess they could be retained.

code/espurna/serialsensor.cpp
pbek: Sneaky, I'll try that. 😁  Thank you.

code/espurna/serialsensor.cpp
mcspr: > But how would I get the variables into the callback of terminalRegisterCommand? I couldn't sneak them in in the [].

Which ones though? I meant something like this

```cpp
terminalRegisterCommand(F("MQTT.SEND"), [](const terminal::CommandContext& ctx) {
    if (ctx.argc == 3) {
        if (mqttSend(ctx.argv[1].c_str(), ctx.argv[2].c_str(), false, false)) {
            terminalOK(ctx);
        } else {
            terminalError(ctx, F("Cannot queue the message")); //
        }
        return;
    }

    terminalError(ctx, F("MQTT.SEND <topic> <payload>"));
});
```

Unless we somehow want the command names to end up in the `commands` / `help`

pbek: Yes, sounds like a great idea too. I'll try that later this week!

pbek: In now, works nice.

mcspr: > In now, works nice.

Thanks! Will pull things tomorrow

mcspr:
Sorry for the delay :)
I have tried to use [Codespaces](https://github.dev/pbek/espurna/blob/feature/serial-sensor/code/espurna/main.cpp) to fixup based on my last comment, but it failed to push to the writable PR branch... So that involved some manual git manipulations with a proper PC.

I moved the func to the mqtt.cpp, so no need for the extra `-DSERIAL_SENSOR_SUPPORT=1` and it just works when MQTT is enabled.

pbek:
Great too, thank you very much! 😁

PR #2484, opened 2021-11-12
SHT3X: add 0x44 address

drc38:
amend code to work with both i2c addresses as per datasheet https://dfimg.dfrobot.com/nobody/wiki/88b31350da4f54d00989c74c6fa392f7.pdf


mcspr:
Thanks!

PR #2488, opened 2021-11-24
Add support for Yagusmart 1, 2 ad 3 gang switches

MelanieT:
https://www.amazon.co.uk/gp/product/B086MV5MC8

These don't come with an esp8266 anymore, but can be trivially converted
as the new chip is pin compatible. Note, GPIO15 needs to be connected to
GND on 1 and 2 gang switches in order to enable the ESP to boot.

Older versions and some US models may still ship with ESP8266.

Caution, do NOT solder a serial port while the board is connected to the mains baseboard, it will blow your fuse and your USB! If you need to flash with a programmer, use a bench power supply on the logic board only!

### DIFF
--- a/code/espurna/config/hardware.h
+++ b/code/espurna/config/hardware.h
@@ -825,116 +825,6 @@
     // Only used when RFB_PROVIDER is RCSWITCH
     #define RFB_RX_PIN          4
     #define RFB_TX_PIN          5
-#elif defined(YAGUSMART_TOUCH_HWMOD_1G)
-    // Info
-    #define MANUFACTURER        "YAGUSMART"
-    #define DEVICE              "TOUCH_HWMOD_1G"
-
-    #define BUTTON1_CONFIG      BUTTON_PUSHBUTTON | BUTTON_DEFAULT_HIGH
-    #define BUTTON1_PIN         3
-    #define BUTTON1_RELAY       1
-    #define BUTTON1_PRESS       BUTTON_ACTION_TOGGLE
-    #define BUTTON1_CLICK       BUTTON_ACTION_NONE
-
-    #define RELAY1_TYPE         RELAY_TYPE_NORMAL
-    #define RELAY1_PIN          13
-
-    #define LED2_PIN                    14
-    #define LED2_PIN_INVERSE            1
-    #define LED2_MODE                   LED_MODE_FOLLOW
-    #define LED2_RELAY                  1
-
-    #define LED1_PIN                    0
-    #define LED1_PIN_INVERSE            1
-    #define LED1_MODE                   LED_MODE_WIFI
-
-    #define DEBUG_SERIAL_SUPPORT        0
-
-#elif defined(YAGUSMART_TOUCH_HWMOD_2G)
-    // Info
-    #define MANUFACTURER        "YAGUSMART"
-    #define DEVICE              "TOUCH_HWMOD_2G"
-
-    #define BUTTON1_CONFIG      BUTTON_PUSHBUTTON | BUTTON_DEFAULT_HIGH
-    #define BUTTON1_PIN         3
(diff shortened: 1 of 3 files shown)


code/espurna/config/hardware.h
```diff
@@ -825,6 +825,116 @@
     // Only used when RFB_PROVIDER is RCSWITCH
     #define RFB_RX_PIN          4
     #define RFB_TX_PIN          5
+#elif defined(YAGUSMART_TOUCH_HWMOD_1G)
+    // Info
+    #define MANUFACTURER        "YAGUSMART"
+    #define DEVICE              "1_GANG_SWITCH"
```
mcspr: `TOUCH_HWMOD_1G`? or move DEVICE text into the build flag token
-//- the rest of the .h, and also the .ini

The idea here is to have `MANUFACTURER "_" DEVICE` be the same as the build flag

code/platformio.ini
```diff
@@ -1,5 +1,5 @@
 [platformio]
-default_envs = nodemcu-lolin
+default_envs = yagusmart-touch-hwmod-1g
```
mcspr:
```suggestion
default_envs = nodemcu-lolin
```

MelanieT: Of course. Sorry I didn't look closely enough

code/espurna/config/hardware.h
MelanieT: I didn't know.

mcspr:
Thanks!
I included the comment above in the header, plus modified the changelog

MelanieT:
I have another thing that has come up in hacking these switches: They have relays in them that click even if all you wnt is to use it as a controller for, say. a smart bulb. I would suggest that, under switches, there be a checkbox that would tell espurna to not actually toggle a relay gpio, but only pretend, so that on a button action, a relay/x/set message is sent without an actual relay being energised. That would allow an easy choice between having a smart switch as a pure control device, or an actual switch. I could envision what can be done in the switch module to facilitate that, but I have no idea how to surface something like that in the web UI. However, it seems to be a vaild use case, at least for openhab. Do you have any pointers on how to do that?

mcspr:
> I would suggest that, under switches, there be a checkbox that would tell espurna to not actually toggle a relay gpio, but only pretend, so that on a button action, a relay/x/set message is sent without an actual relay being energised.

The way Xose originally implemented this is by introducing DUMMY switches and a global PROVIDER build flag so we could have RFBRIDGE toggle virtual relays on RF code arrival, and also send some RF codes when relay toggles. `dev` right now replaced global 'providers' with a per-switch settings that may change it's configuration.
https://github.com/xoseperez/espurna/blob/03442608da44edaafab464a1e224eb07bffc2688/code/espurna/relay.cpp#L463
https://github.com/xoseperez/espurna/blob/03442608da44edaafab464a1e224eb07bffc2688/code/espurna/relay.cpp#L408-L421
https://github.com/xoseperez/espurna/blob/03442608da44edaafab464a1e224eb07bffc2688/code/espurna/relay.cpp#L582-L596

current setting is a simple RelayProvider enum <-> integer cast e.g.
```
set relayProv0 1
reset
```
will no longer do anything with the associated GPIO

*edit*
or, there are always build flags
https://github.com/xoseperez/espurna/blob/03442608da44edaafab464a1e224eb07bffc2688/code/espurna/relay.cpp#L194-L205

MelanieT:
But how to surface that in the UI? Sure, a telnet command is a good start, it means the groundwork is in place, but all the nontechnical users woudln't know about it.

mcspr:
For the setting, see existing `<select>` code in the .js & .html? List of available options is missing atm, but the relay's provider key-value is already sent to the UI.
DEBUG panel is both log and terminal interface btw

PR #2516, opened 2022-05-20
Add methods to set shunt resistor & cal values

hamed-ta:


code/espurna/sensors/INA219Sensor.h
```diff
@@ -377,6 +381,15 @@ class INA219Sensor : public I2CSensor<BaseEmonSensor> {
...
     }
 
+    void setShuntRes(float shunt) {
+        _shunt_Factor = shunt / 0.1;
+        _dirty = true;
+    }
+
+    void setCorrectionFactor(float corr){
+        corr_calibration = _params.calibration * corr;
+    }
```
mcspr, line 391: See datasheet, plus my excerpt from it in the comment.
$cal = \frac{0.04096}{current LSB * R shunt}$
$current LSB = \frac{maximum expected current}{2^{15}}$
$power LSB = 20 * current LSB$

Expected current is based on PGA gain, if I understood the values correctly.
Why not apply it like that so we have every value correctly scaled, keeping the structure?
(and yeah, existing lsb values are also off, looking at the table in the comment I mentioned... mA vs. A, and I think the calc is in milli scale)

mcspr, line 391: See datasheet, plus my excerpt from it in the comment.
$cal = \frac{0.04096}{current LSB * R shunt}$
$current LSB = \frac{maximum expected current}{2^{15}}$
$power LSB = 20 * current LSB$

Expected current is based on PGA gain, if I understood the values correctly.
Why not apply it like that so we have every value correctly scaled, keeping the structure?
(and yeah, existing lsb values are also off, looking at the table in the comment I mentioned... mA vs. A, and I think the calc is in milli scale)

code/espurna/sensors/INA219Sensor.h
```diff
@@ -431,7 +444,10 @@ class INA219Sensor : public I2CSensor<BaseEmonSensor> {
             .bus_range = _bus_range,
         });
 
-        _port.calibration(_params.calibration);
+         corr_calibration = corr_calibration ? corr_calibration * _params.calibration : _params.calibration;
```
mcspr, line 447: ^ and would mean, we apply some sort of func that creates Parameters on the fly instead of choosing one of our hard-coded ones

code/espurna/sensors/INA219Sensor.h
hamed-ta: Yes, good idea, we can calculate the cal value based on the shunt value that has been set but we still need to keep the setCorrectionFactor function as one may need to change the calibration value based on the desired current resolution. 

do you want me to change the code?

code/espurna/sensors/INA219Sensor.h
hamed-ta: what do you mean?

code/espurna/sensors/INA219Sensor.h
hamed-ta: Also, we can add another function for the expected current.

mcspr: Both expected current and resistance, as it would remove the need for `_fromGain()` based on pga config. I would really prefer to follow datasheet here.

code/espurna/sensors/INA219Sensor.h
mcspr: see above. right now, we hard-code calibration based on pga config in `_fromGain()`.

PR #2518, opened 2022-05-21
sht3x: add crc and status debug

drc38:


code/espurna/sensors/SHT3XI2CSensor.h
```diff
@@ -78,11 +78,43 @@ class SHT3XI2CSensor : public I2CSensor<> {
...
+        void status_register() {
+            const auto address = lockedAddress();
+            
+            unsigned char buffer[3];
+            bool crc, cmd, htr;
+            // Read status register
+            i2c_write_uint8(address, 0xF3, 0x2D);
+            espurna::time::blockingDelay(
+                espurna::duration::Milliseconds(500));
+            i2c_read_buffer(address, buffer, std::size(buffer));
```
mcspr: What's the point of 500ms waits? Maximum wait time mentioned in the datasheet is 15ms for measurements to happen, while pre() happens every second at a minimum

code/espurna/sensors/SHT3XI2CSensor.h
```diff
@@ -78,11 +78,43 @@ class SHT3XI2CSensor : public I2CSensor<> {
...
+                htr = buffer[0] & 0b00100000;
+                DEBUG_MSG_P(PSTR("[SHT3X] Status crc:%u cmd:%u htr:%u\n"), crc, cmd, htr);
+            }
+            else {
+                DEBUG_MSG_P(PSTR("[SHT3X] Checksum error\n"));
+            }
+            // Clear status register
+            i2c_write_uint8(address, 0x30, 0x41);
+            espurna::time::blockingDelay(
+                espurna::duration::Milliseconds(500));
```
mcspr: -//-
why wait?

code/espurna/sensors/SHT3XI2CSensor.h
```diff
@@ -78,11 +78,43 @@ class SHT3XI2CSensor : public I2CSensor<> {
...
+            i2c_write_uint8(address, 0xF3, 0x2D);
+            espurna::time::blockingDelay(
+                espurna::duration::Milliseconds(500));
+            i2c_read_buffer(address, buffer, std::size(buffer));
+            if (CRC8(buffer[0],buffer[1],buffer[2])) {
+                // see https://sensirion.com/resource/datasheet/sht3x-d
+                crc = buffer[1] & 0b00000001;
+                cmd = buffer[1] & 0b00000010;
+                htr = buffer[0] & 0b00100000;
+                DEBUG_MSG_P(PSTR("[SHT3X] Status crc:%u cmd:%u htr:%u\n"), crc, cmd, htr);
```
mcspr: is the log here intentional w/o SENSOR DEBUG flag? this happens every read.
plus, `%02X%02X` with buffer msb and lsb so we see the whole number?

code/espurna/sensors/SHT3XI2CSensor.h
drc38: I hadn't thought about it particularly and just used the delay from the `begin()` function, reduced delay to 20ms in latest commit

code/espurna/sensors/SHT3XI2CSensor.h
drc38: Agree, removed

code/espurna/sensors/SHT3XI2CSensor.h
drc38: Used SENSOR_DEBUG flag and added whole status output

code/espurna/sensors/SHT3XI2CSensor.h
mcspr: btw soft-reset per datasheet is 0.5ms
I wonder if it was interpreted as 0.5seconds

PR #2521, opened 2022-05-26
sns: use LastFilter for real time api

drc38:
@mcspr not sure if this is the best way in order to match the description for MQTT when enabling the Real Time Api option  "_Enable this to get unfiltered values in the API (HTTP, MQTT, Prometheus, etc.)_"

### DIFF
--- a/code/espurna/sensor.cpp
+++ b/code/espurna/sensor.cpp
@@ -1193,7 +1193,9 @@ BaseFilterPtr _magnitudeCreateFilter(unsigned char type) {
     if (!filter) {
         filter = std::make_unique<MedianFilter>();
     }
-
+    if (sensor::settings::realTimeValues()) {
+        filter = std::make_unique<LastFilter>();
+    }
     return filter;
 }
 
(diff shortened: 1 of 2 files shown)


drc38:
Another approach would be to include an if statement to change the reporting to `value.processed` rather than `value.filtered`
`_sensorReport(magnitude, magnitude_index, value.filtered);`

mcspr:
idk. We could call HTTP whenever on our own schedule, so that's the reason we would need to be able to grab read data directly and create our own report / filter / processing with some external program using the data. Also note that there are `{type}MinDelta{number}` and `{type}MaxDelta{number}` magnitude options that may report before normal report count overflow when the value changes happen.

3rd option would be to introduce per-magnitude filter setting?
(something like `{type}Filter{number} => "last"`)

And filter is only created on boot, still, so it should also be able to re-create it when changed?
(with both options targeting filter change)

drc38:
> 3rd option would be to introduce per-magnitude filter setting

Assuming you can set this from web interface that would provide the most flexibility. Ie you could select median, mean, last etc filter per sensor?

mcspr:
> > 3rd option would be to introduce per-magnitude filter setting
> 
> Assuming you can set this from web interface that would provide the most flexibility. Ie you could select median, mean, last etc filter per sensor?

Per 'magnitude' object, something like units.

mcspr:
09c0750c693b5f0546e7fb3ba8329b8c8dfc19f7 allows changing magnitude filter with `${name}Filter${index}`
e.g. `set tmpFilter0 last` and `reset` will create a `LastFilter` object instead of the `MedianFilter`

not in the WebUI, though, idk how frequent that use-case would be. if you still want that, you might want to follow the unit functions with enumerable collections and selects, plus mark it as a 'reboot' action.

PR #2534, opened 2022-09-04
add link to shop for openenergymonitor

gotjoshua:
https://shop.openenergymonitor.com/wifi-mqtt-relay-thermostat/


mcspr:
looking the list and the paragraph before... what about
https://github.com/xoseperez/espurna/wiki/Hardware-OpenEnergyMonitor-Wifi-MQTT-Relay ?
(yet, it would also be a sole link to anywhere in that part of the readme :)

hardware .h should also have this url

gotjoshua:
> would also be a sole link to anywhere

I thought to check first with a one off if the maintainers like the idea to add some links in that table... Could be the first of many... Or could be that you all feel it doesn't fit

mcspr:
Links are useful addition. But, I'd prefer internal ones to wiki over the external shop / description / product pages

mcspr:
> Links are useful addition. But, I'd prefer internal ones to wiki over the external shop / description / product pages

README updated btw, openenergymonitor included

Looking theengs/decoder and its docs, I wonder if wiki could also use an update so that hardware.h entries receive a direct link (or, some script just takes an output of `gcc -E` at puts it there for our device definitions)
https://github.com/theengs/decoder/blob/development/docs/devices/ABTemp.md

PR #2552, opened 2022-11-07
sns: fix slash removal in modules

ktkd:
fix invalid prometheus syntax `strconv.ParseFloat: parsing "/0": invalid syntax`


mcspr:
Tnx!
I also took the liberty to update another `remove('/')` call in rules that was also broken in 5fcac5d27f713ca4a84da3dc8f0b4d10c25c3050

PR #2584, opened 2023-03-17
Show message only for real pin-lock failures

iprak:
This PR addresses the message appearing at the top of web UI about not being able to lock pins.

The data received did not have any failed-locks information.
image


iprak:
Thanks for adding those files. I was not sure if they got automatically built as part of commit hook.

PR #2612, opened 2024-07-09
webui: Improve Scheduler type interface

opticron:
The scheduler type selection is not intuitive as there is no prompt to change the empty drop-down box to another value. This makes it obvious that something needs to be selected.

### DIFF
--- a/code/html/src/template-sch.html
+++ b/code/html/src/template-sch.html
@@ -8,9 +8,9 @@
         <div class="pure-g">
             <div class="pure-control-group pure-u-1 pure-u-lg-1-4">
                 <select class="pure-input-2-3" name="schType" required>
-                    <option selected disabled value=""></option>
+                    <option selected disabled value="">Select Option</option>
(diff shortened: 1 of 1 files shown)


mcspr:
What about reportValidity() instead?
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/reportValidity

Non-interactive part
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/checkValidity

having more text in an already pretty narrow input might be too much?
ed26dbd attempts to force everything through validity checks before being allowed to apply any settings, I'd expect this would be much more apparent than select label. for schedule inputs, this generates a warning immediately after template is injected. plus, there is also an annoying warning after editing the field for basic text inputs

opticron:
Thanks for adding that. Does that commit obsolete this change, then?

mcspr:
"makes it obvious" that elem is required when editing should be solved, yes

opticron:
Ack

PR #2622, opened 2024-10-11
Add support for DYP A02YYU ultrasonic sensor

toniSg:
This brings the support of DYP A02YYU waterproof ultrasonic sensor to espurna.

It will need a UART port configured as 9600/8N1 (RX pin only is enough). To enable this sensor, the follwing build flags shall be configured:

**A02YYU_SUPPORT** set to 1
**A02YYU_PORT** set to the UART port ID

This has been tested on an esp8266 D1 Mini reporting steady distance measurements over more than 2 weeks.


code/espurna/sensors/A02YYUSensor.h
```diff
@@ -0,0 +1,148 @@
...
+                } else {
+                    _rdIndex = (_rdIndex + 1) & _RX_MASK;
+                    _rxCount--;
+                }
+            }
+        }
+
+        // ---------------------------------------------------------------------
+
+        char _rx[_RX_LEN];
```
mcspr: `uint8_t`
*(edit: see read() / readBytes() below)*

code/espurna/config/sensors.h
```diff
@@ -90,6 +90,19 @@
...
+// Enable support by passing A02YYU_SUPPORT=1 build flag
+//------------------------------------------------------------------------------
+
+#ifndef A02YYU_SUPPORT
+#define A02YYU_SUPPORT                  0
+#endif
+
+#ifndef A02YYU_PORT
+#define A02YYU_PORT                     1   // By default, use the first port
+                                            // (needs `UART[1-3]_BAUDRATE 9600`)
```
mcspr, line 103: per A02YYU https://www.dypcn.com/uploads/A02-Output-Interfaces.pdf

> (2) Communication instruction
When the pin(RX)is suspended or the input high level, the module outputs processed value, and the data is more stable. The response time of is 100~500ms. When the pin(RX)is input low level, the module outputs real-time value. Response time of is 100ms.

(assuming 'suspended' when referencing pin translates to 'floating' / 'not connected')

So TX from ESP side can be NONE, and probably does not need any connection at all.
Not sure whether it is worth adding extra cfg opt here for that PIN state, though... I suppose it can be set to either low or high through other means.

code/espurna/config/sensors.h
```diff
@@ -1519,6 +1532,7 @@
 // Provide generic way to detect sensor presence
 #ifndef SENSOR_SUPPORT
 #define SENSOR_SUPPORT ( \
+    A02YYU_SUPPORT || \
```
mcspr, line 1535: code/espurna/config/arduino.h
should have `//#define A02YYU_SUPPORT 1` for IDE user reference

code/test/build/sensor.h
should have `#define A02YYU_SUPPORT 1` to trigger CI build

code/espurna/sensors/A02YYUSensor.h
```diff
@@ -0,0 +1,148 @@
...
+            int n;
+
+            _error = SENSOR_ERROR_OK;
+
+            // read all available bytes into rx circular buffer
+            while ((n = _serial->available()))
+            {    
+                for (; n > 0; n--) 
+                {
+                    char c = _serial->read();
```
mcspr: readBytes(..., available) (will block when port->available() < available) ?
read(..., available) (does not block) ?

Both write to u8 buffer output immediately, and it might as well scan the whole available chunk of data at once and not be limited by 16bytes per tick. I am not sure I fully get why do we need per-char reads here with 5 separate counters.

code/espurna/config/sensors.h
toniSg: Thanks for pointing this out. I've added the missing defines.

code/espurna/config/sensors.h
toniSg: Indeed, I've assumed that the default configuration (leaving the sensor's RX pin floating) was the most common usage for this device. For advanced users who need unprocessed realtime measurements from the sensor, I've considered that they would drive the sensor's RX pin by their own leaving them the choice of doing it through a configurable gpio or by pulling down directly.

code/espurna/sensors/A02YYUSensor.h
toniSg: According my understanding of stream.readBytes(), it would block if there is not enough data and timeout (defined by setTimeout()).

The point here is to implement an asynchronous read/consume process through a circular buffer which is fed at a given input pace, while being parsed anytime there is enough data. The  per byte read facilitates the wrapping of the circular buffer. Would you suggest that it would have been better to implement it using for instance 2x readBytes() (one to fill until the end of the buffer and the second to fill the beginning on wrapping situation)?

Also the motivation to use such circular buffer is to offer some error resisience against erroneous or missing data. In case the parse() fails, it just drops the oldest received byte and try again starting with the next received byte ensuring that if any valid data is received, it would be properly parsed. This demonstrated pretty good result with a faulty sensor I had, this code was able to recover a few valid measurements here and there among randomly broken data.

code/espurna/sensors/A02YYUSensor.h
toniSg: I missed that one, I've changed the type to uint8_t.

code/espurna/sensors/A02YYUSensor.h
mcspr: > Would you suggest that it would have been better to implement it using for instance 2x readBytes() (one to fill until the end of the buffer and the second to fill the beginning on wrapping situation)?

Something like that. We already have 'available' recorded, reading with its value would not block.

> Also the motivation to use such circular buffer is to offer some error resisience against erroneous or missing data. In case the parse() fails, it just drops the oldest received byte and try again starting with the next received byte ensuring that if any valid data is received, it would be properly parsed. This demonstrated pretty good result with a faulty sensor I had, this code was able to recover a few valid measurements here and there among randomly broken data.

I am just wary of having the reader tracking a lot of state, when essentially the algo is 'find the 0xff and read the next 3 bytes'

BTW are you sure `_error = OVERFLOW` below is valid? Shouldn't it scan the whole 'available' instead of dropping it midway?

toniSg: Following your comments, I revisited the approach and would like to suggest a simpler implementation of the processing loop. Rather than working with a circular buffer and asynchonous read/consume, the draft code below actually uses a linear frame buffer and parses it immediately as soon as there are enough data.

Let me know if this looks easier to review. If so, I'll finalize it, do some tests and validation before resubmitting it.

Thanks.

```cpp
#define _FRAME_LEN 4            // frame buffer length
uint8_t _frame[_FRAME_LEN];     // distance frame data buffer
int _free;                      // index of the first free byte in the frame buffer

void _read()
{
    int available = _serial->available();

    while (available) 
    {
        // read bytes from the device into the frame buffer
        int n = readBytes(_frame + _free, min(_FRAME_LEN - _free, available));

        available -= n;
        _free += n;
        
        if (_free == _FRAME_LEN) // the frame buffer is filled
        {
            int d = _parse_frame();
            
            if (d >= 0)
            {
                // distance successfully parsed
                _distance = d;      // update the distance
                _free = 0;          // empty the frame buffer
            } else {
                // out of sync or corrupted, shift the frame buffer by one byte
                for (int i = 0; i < _FRAME_LEN - 1; i++)
                    _frame[i] = _frame[i + 1];

                _free = _FRAME_LEN - 1;
            }
        }
    }
}
```

mcspr: It does look more streamlined, thanks.

Considering 'faulty sensor', should you need some validation checks to persist in the repo, I have also tried to fix local sensor builds w/ cmake
https://github.com/xoseperez/espurna/commit/2611b6294f5840d9ba7492d8588074c27d4dc44d#diff-b726c4b4ac60a04fa60bfd29302c3fb9aa412748579309e39ff1974c336a7080R1
code/test/unit/src/sensor/sensor.cpp *should* be able to pull the sensor header and call ->tick() then ->value(0) / ->error() reading from a provided Stream

code/espurna/sensors/A02YYUSensor.h
```diff
@@ -0,0 +1,148 @@
...
+
+#if SENSOR_SUPPORT && A02YYU_SUPPORT
+
+#pragma once
+
+#include "Arduino.h"
+#include "BaseSensor.h"
+
+#define _RX_LEN     16
+#define _RX_MASK    (_RX_LEN - 1)
```
mcspr: btw these should either be local to the class as 'constexpr static' or at least prefixed w/ A02YYU\_...

toniSg: Thanks for the comment, I've changed to static constexpr.

code/espurna/sensors/A02YYUSensor.h
toniSg: I've committed the simplified implementation and a few cleanups. It has run overnight successfully. So far, I believe the open items have been covered. Let me know if I missed anything.

mcspr:
Thanks!

Still, I do think it might be useful to keep payload / overall data string examples around. Small example w/ test sensor cpp update
https://github.com/xoseperez/espurna/commit/3cdccfad8dedeba53c48acb817fc1525b7fa2450
And I think I'll try to cover the rest of serial sensors (and clean-up existing pre() / tick() / value() inconsistent reads and timeout uses / no timeouts while at it)

toniSg:
> Thanks!
> 
> Still, I do think it might be useful to keep payload / overall data string examples around. Small example w/ test sensor cpp update [3cdccfa](https://github.com/xoseperez/espurna/commit/3cdccfad8dedeba53c48acb817fc1525b7fa2450) And I think I'll try to cover the rest of serial sensors (and clean-up existing pre() / tick() / value() inconsistent reads and timeout uses / no timeouts while at it)

Sure, I'll try to do some data capture from the sensor this week end and I'll submit them.

toniSg:
Hello Max, 

As promised, please find attached some data capture from the device. Unfortunately, I couldn't capture data from my faulty device, it doesn't power up anymore :( I think it's definitely dead.

I'll think about how to exploit those data in the test and also how to simulate faulty/corrupted data. Basically, with the faulty device, some values were randomly dropped or zeroed. It should be easy to replicate that pattern for the tests.

I've also looked at your test and it's a very good thing, I'll try to contibute to this effort. I'd have one question about this though, in the standard behavior, several frames that are parsed in once within a tick() call. The test will only test the latest parsed value. would it be possible to check at all parsed values?

[a02yyu-data-capture.txt](https://github.com/user-attachments/files/17535466/a02yyu-data-capture.txt)

mcspr:
> I'd have one question about this though, in the standard behavior, several frames that are parsed in once within a tick() call. The test will only test the latest parsed value. would it be possible to check at all parsed values?

Most naïve approach would be to just throttle the input to one frame per tick()? (like it does currently)

Otherwise, manually pick results with an extra external method to handle one frame at a time. Bulk of the data passed at once continues to only have one result at the end.

PR #2645, opened 2025-06-02
Fix typos

hamed-ta:
- fix typos in documentation and comments
- correct various misspellings across source files

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Since November 2018, Max Prokhorov (**@mcspr**) is also actively working as a co
 
 ## Contributors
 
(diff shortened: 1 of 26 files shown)


hamed-ta:
What was the issue with it?

mcspr:
> What was the issue with it?

I merged dictionary changes with! 
https://github.com/xoseperez/espurna/commit/7260ec833a178f2d7e51a1857e288389b214b12f
(I also learned I cannot spell immediatly) 

For one... long sentences, weird words, etc. I changed or removed instead of only correcting typos.
Second, comments that no longer apply / are useful to the reader. Like in webui, there are still some things from c/p of the original custom.js talking about jquery

PR #2649, opened 2025-06-07
Updated the reverse loop in AnimFly::Run to cast numLeds to a signed.

hamed-ta:
Updated the reverse loop in AnimFly::Run to cast numLeds to a signed type, preventing an endless loop when shifting LEDs backward

### DIFF
--- a/code/espurna/garland/animations/anim_fly.h
+++ b/code/espurna/garland/animations/anim_fly.h
@@ -27,7 +27,7 @@ class AnimFly : public Anim {
             }
         } else {
             launchpos = 0;
-            for (auto i = numLeds - 2; i >= 0; i--) {
+            for (int i = static_cast<int>(numLeds) - 2; i >= 0; --i) {
                 leds[i + 1] = leds[i];
             }
         }


hamed-ta:
There are two issues with this code:

1: since numLeds is unsigned i will be unsigned too, if number of leds are less than 2 it will underflow, resulting in a large unsigned value.

2: as i is unsigned the condition is i >= 0 is always true and the loop will never terminate.

mcspr:
> 1: since numLeds is unsigned i will be unsigned too, if number of leds are less than 2 it will underflow, resulting in a large unsigned value.
> 2: as i is unsigned the condition is i >= 0 is always true and the loop will never terminate.

note the `auto`, it is already signed type
c & as the result c++ too do a lot of hidden promotions under the hood. `>>` is one example, here's another with `-`

using lsp client / ide w/ language hints already shows it, but might as well be helpful to the reader
(since I also generally forget about it)

mcspr:
> > 1: since numLeds is unsigned i will be unsigned too, if number of leds are less than 2 it will underflow, resulting in a large unsigned value.
> > 2: as i is unsigned the condition is i >= 0 is always true and the loop will never terminate.
> 
> note the `auto`, it is already signed type c & as the result c++ too do a lot of hidden promotions under the hood. `>>` is one example, here's another with `-`
> 
> using lsp client / ide w/ language hints already shows it, but might as well be helpful to the reader (since I also generally forget about it)

https://godbolt.org/z/ecz7KrTz3
(but, only because sizeof(u16) < sizeof(int). compare w/ similarly sized types)

hamed-ta:
You're absolutely right — I overlooked the promotion behavior of small unsigned types. That said, I still think it would improve readability to explicitly use a signed type here.

mcspr:
> You're absolutely right — I overlooked the promotion behavior of small unsigned types. That said, I still think it would improve readability to explicitly use a signed type here.

...since we agree on readability part, I added it in the commit above combining 2 PRs
> [mcspr](https://github.com/mcspr) closed this in https://github.com/xoseperez/espurna/commit/d940abc0aba9512e8ac9a24ca105a054a849280c [2 days ago](https://github.com/xoseperez/espurna/pull/2649#event-18046090196)

commit message is auto-closing issues
> resolve #2649
> resolve #2650

github is not really into this whole raw '.patch' thing, cannot mark PR as 'resolved' same as issues

*edit:* see https://github.com/orgs/community/discussions/6414
afaik, only thing I could do from my side when this kind of merge happens is maybe press 'Approve' button in a review

hamed-ta:
I’ve noticed that fixes I propose are often declined but later implemented in a similar form in your changes. While I appreciate the effort, it’s discouraging when contributions go unacknowledged.

mcspr:
> I’ve noticed that fixes I propose are often declined but later implemented in a similar form in your changes. While I appreciate the effort, it’s discouraging when contributions go unacknowledged.

So there is a misunderstanding in the approach to patching then? I have merged PRs, just not through Github UI! I don't say thanks enough, that's true. But, I also though it is obvious by the comment marker and 'resolved / closed by' commit message(s)

#2645 was the instance there I actually read the comments. As acknowledged, that was a copilot change, so I went over and fixed the prose instead of most direct stuff. I don't suppose the better approach was to bore you with back and forth?

#2649 & #2650 are merged manually as they are both affecting garland and are related semantically.
Once again, notice that d940abc0a 'From:' field is set to you. I won't forget CHANGELOG.md, but I guess that does not count.

PR #2650, opened 2025-06-07
Revised the pixel cache update in ColorWave::move to iterate with size_t and decrement safely

hamed-ta:
Revised the pixel cache update in ColorWave::move to iterate with size_t and decrement safely


hamed-ta:
The same issue here.

PR #2658, opened 2025-12-29
feat(button): Add AC frequency detection switch support (Moes MS-104B)

igorufox:
## Description

This PR introduces support for **AC Frequency Detection Switches**.

Some devices, such as the **Moes MS-104B** (2-Gang Switch Module), do not use standard logic levels (High/Low) to represent the state of the physical wall switch. Instead, they use an optocoupler connected to the AC mains.
*   **Switch Closed:** The GPIO pin receives a 50Hz/60Hz pulsed signal (zero-crossing detection).
*   **Switch Open:** The GPIO pin is floating or pulled high/low (no pulses).

Standard debounce logic interprets these pulses as rapid button presses. This PR adds a new `AcSwitch` module that counts pulses within a time window to determine the stable state (ON/OFF).

## Key Changes

1.  **New `AcSwitch` Class**:
    *   Uses interrupts to count pulses on the GPIO pin.
    *   Implements a **latching mechanism** (software filter) to ignore noise and contact bounce, ensuring the relay doesn't toggle multiple times during a single switch transition.
    *   Configurable timeout and minimum pulse threshold.

2.  **Button Module Integration**:
    *   Added `BUTTON_FREQUENCY` flag to `types.h`.
    *   Updated `button.cpp` to initialize `AcSwitch` when `BUTTON_FREQUENCY` is set in the config.
    *   Added a runtime setting `btnAcDbg` to enable specific debug logs for AC detection tuning without recompiling.

## Configuration

To use this feature, define the button config as `BUTTON_FREQUENCY`.

**Example `platformio.ini` / `hardware.h`:**

```ini
// Example for Moes MS-104B
#define BUTTON1_PIN 13
#define BUTTON1_CONFIG BUTTON_FREQUENCY
#define BUTTON1_RELAY 1
#define BUTTON1_PRESS BUTTON_ACTION_TOGGLE
#define BUTTON1_RELEASE BUTTON_ACTION_TOGGLE
#define BUTTON1_DEBOUNCE_DELAY 100 // 100ms window is recommended for 50Hz

### CODE REVIEW

File: code/espurna/ac_switch.cpp [hunk 1]
```diff
@@ -0,0 +1,134 @@
...
+    } else if (frequency == 60) {
+        // ~8333us period for 60Hz. Allow 3.5% tolerance.
+        _min_filter_period = 8041;
+        _max_filter_period = 8625;
+    }
+
+    // INPUT_PULLUP is usually safer if the optocoupler pulls to ground.
+    // If the schematic is different, INPUT might be required.
+    pinMode(_pin, INPUT_PULLUP);
+    attachInterrupt(_pin, [this]() { this->_isr(); }, CHANGE);
```
mcspr, line 30: iirc this does not work w/ current SDK, lambda function itself is not isr-safe and *might* be purged from IRAM dynamic part
passing this as arg does work though, simply unwrap void back into class type

code/espurna/button.cpp
```diff
@@ -933,6 +993,12 @@ ButtonEvent Button::loop() {
         }
     }
 
+    if (ac_switch) {
+        if (ac_switch->loop()) {
+            return ac_switch->state() ? ButtonEvent::Pressed : ButtonEvent::Released;
+        }
+    }
+
```
mcspr, line 1001: any reason why `event_emitter->loop()` can't be used? see AnalogPin below, but assume that `ac_switch->loop()` is replaced w/ `T::digitalRead()` polling
programmatic LOW or HIGH signal is then used in the DebounceEvent loop to deduce button state

code/platformio_override.ini
```diff
@@ -0,0 +1,40 @@
+; User defined configuration for Moes MS-104B 2-Gang Switch Module
+
+[env:moes-ms-104b]
```
mcspr, line 3: cfg below usually goes into `config/hardware.h`, .ini flags are then replaced with just `-DMOES_MS_104B`

